PR #16, opened 2020-08-06
solution

oggyrock:


src/formatDate.js
```diff
@@ -19,7 +25,28 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const arr = date.split(fromFormat[3]);
+  const format = {};
+  const ans = [];
```
rmnkk, line 30: I think that this variable should be renamed.

PR #17, opened 2020-08-06
solution

varseniuk:


src/formatDate.js
```diff
@@ -19,7 +9,36 @@
...
+      } else if (fromFormat.includes('YYYY')) {
+        result.push(arrDate[fromFormat.indexOf('YYYY')].slice(-2));
+      }
+    }
+
+    if (toFormat.indexOf('YYYY') === i) {
+      if (fromFormat.includes('YYYY')) {
+        result.push(arrDate[fromFormat.indexOf('YYYY')]);
+      } else if (fromFormat.includes('YY')) {
+        result.push(`20${arrDate[fromFormat.indexOf('YY')]}`);
```
krajnova, line 36: maybe you should add case if there is not 21st century but 20th also?

varseniuk: There's no mention in this task about whether it should be 21'st or 20'th century, and there is no specific condition mentioned. So I decided not to fantasize  and since we live in the 21'st century, I assumed that there should always be added '20'.

varseniuk: Asked Volodia on q&a about this. He told that this check should be applied. So I've added it.
Thank for your comment. It was useful 👍

PR #21, opened 2020-08-07
Solution

rmnkk:


src/formatDate.js
```diff
@@ -20,6 +11,29 @@
...
+
+  if (fromFormat[0] !== toFormat[0]) {
+    [formattedDate[0], formattedDate[1]] = [formattedDate[1], formattedDate[0]];
+  }
+
+  if (fromFormat[0].includes('YY')) {
+    return formattedDate.reverse().join(`${separator}`);
+  }
+
+  return formattedDate.join(`${separator}`);
```
Andrii256, line 36: I cant surely say that it is bad code. May be it happends beacuse of I am too silly. But I cant understand your code, I cant understand what is happening here.

PR #22, opened 2020-08-07
Solution

nznmv:


src/formatDate.js
```diff
@@ -19,7 +25,34 @@
...
+    MM: '',
+    DD: '',
+    YYYY: redate.getFullYear() + '',
+
+    get YY() {
+      return this.YYYY.slice((2));
+    },
+  };
+
+  for (let i = 0; i <= fromFormat.length - 1; i++) {
```
artsashko, line 42: `i < fromFormat.length;` можно просто вот так))

src/formatDate.js
```diff
@@ -19,7 +25,34 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const result = [];
+  const dateArr = date.split(fromFormat[3]);
+  const redate = new Date(date);
+
+  const dateObj = {
+    MM: '',
+    DD: '',
+    YYYY: redate.getFullYear() + '',
```
irenhh, line 35: how about template string?

src/formatDate.js
```diff
@@ -19,7 +25,34 @@
...
+
+    get YY() {
+      return this.YYYY.slice((2));
+    },
+  };
+
+  for (let i = 0; i <= fromFormat.length - 1; i++) {
+    switch (true) {
```
irenhh, line 43: why true?
use the value for case here

src/formatDate.js
```diff
@@ -19,7 +25,34 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const result = [];
+  const dateArr = date.split(fromFormat[3]);
+  const redate = new Date(date);
```
irenhh, line 30: redate is a strange date, how about dateObj?

src/formatDate.js
nznmv: I do not fully understand how to do it

src/formatDate.js
```diff
@@ -19,7 +25,34 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const result = [];
+  const dateArr = date.split(fromFormat[3]);
+  const newDateObj = new Date(date);
+
+  const dateObj = {
+    MM: '',
+    DD: '',
+    YYYY: newDateObj.getFullYear() + '',
```
irenhh, line 35:
```suggestion
    YYYY: `${newDateObj.getFullYear()}`,
```

PR #37, opened 2020-08-08
solved

DmJSCourse:
коменты пришлось удалить, линт их не пропускал.

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -1,25 +1,41 @@
 'use strict';
 
-/**
- * Time flies, standards change. Let's get rid of the routine of changing the date format,
- * and create a function for formatting dates.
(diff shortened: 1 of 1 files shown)


src/formatDate.js
```diff
@@ -1,25 +1,41 @@
...
+        break;
+      case 'YY':
+        if (fromFormat.includes(value)) {
+          result[index] = values[fromFormat.indexOf(value)];
+        } else if (fromFormat.includes('YYYY')) {
+          result[index] = values[fromFormat.indexOf('YYYY')].substring(2);
+        }
+        break;
+      default:
+        break;
```
furude-rika, line 34: maybe it's better to write 'Invalid values' before the break to show that values didn't match those you expected

PR #45, opened 2020-08-09
Solution

martakupka:


src/formatDate.js
```diff
@@ -19,7 +25,59 @@
...
+      if (toFormat[i].length === year.length) {
+        newDate[i] = year.join('');
+      } else {
+        if (toFormat[i].length < year.length) {
+          const shortYear = year.slice(2);
+
+          newDate[i] = shortYear.join('');
+        } else {
+          if (toFormat[i].length > year.length) {
+            if (+year.join('') <= 99 && +year.join('') > 20) {
```
krajnova, line 59:
```suggestion
            if (+year.join('') <= 99 
              && +year.join('') > 20) {
```

krajnova: It would be more readable and you should also change 20 to 30

PR #62, opened 2020-09-21
Solution

DmytroArkhypenko:


src/formatDate.js
```diff
@@ -51,6 +51,44 @@
...
+    }
+
+    if (key === 'YY') {
+      if (dateArr[yearIndexShort]) {
+        resultArr.push(dateArr[yearIndexShort]);
+      } else {
+        resultArr.push(dateArr[yearIndex].slice(2));
+      }
+    }
+  }
```
vkryvytskyy, line 89: use switch..case instead of if..else. It's more readable

vkryvytskyy (approved):

PR #113, opened 2020-12-16
Solution

sasha39612:


5Mountains (approved):
Nice solution with implementing switch/case!

Denchepornyuk (approved):

PR #122, opened 2020-12-16
Solution

khrulenko:


dima-bond (approved):

src/formatDate.js
```diff
@@ -50,7 +50,42 @@
...
+
+  const obj = {
+    get YY() {
+      if (fromFormat.includes('YYYY')) {
+        return arr[fromFormat.indexOf('YYYY')].substr(2, 2);
+      }
+
+      return arr[fromFormat.indexOf('YY')];
+    },
+    get YYYY() {
```
aionova, line 63: There is no need to use getters here. You can use methods.

aionova (approved):

PR #134, opened 2020-12-17
If-solution format date

WhiteRhin0:


MarchD (approved):
good job

src/formatDate.js
```diff
@@ -51,6 +51,43 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const dateArr = date.split(fromFormat[3]);
+  const dateObj = {};
+  const resultArr = [];
+
+  for (let i = 0; i < 3; i++) {
+    dateObj[fromFormat[i]] = dateArr[i];
+  }
```
varseniuk, line 60: For this task, when you know the exact amount of array's elements, it's ok to use '3', but better always use 'array.length' when iterating through the array in a loop.
```suggestion
  for (let i = 0; i < dateArr.length; i++) {
    dateObj[fromFormat[i]] = dateArr[i];
  }
```

varseniuk (approved):
Clever solution

PR #136, opened 2020-12-17
Solution

Roman-Matsuk:


YuriiFitsyk (approved):
Interesting way to format.

src/formatDate.js
```diff
@@ -50,7 +50,38 @@
...
-  // write code here
+  let dateArray = date.split(fromFormat[3]);
+  let fromFormatCopy = [...fromFormat];
+  const result = [];
+  let add = 0;
+
+  if (fromFormatCopy[0][0] !== toFormat[0][0]) {
+    fromFormatCopy = fromFormatCopy.reverse();
+    dateArray = dateArray.reverse();
+  };
```
vsm2k17, line 61: that does not work for cases when position 2 switched with position 3
1, 2, 3 => 1, 3, 2

vsm2k17: image

vsm2k17 (changes requested):
image
does not work for cases when YY => YYYY

vsm2k17 (changes requested):
image
image
something went wrong

vsm2k17 (approved):

PR #138, opened 2020-12-19
Solution

NikitaTudakov:


vitalya2626 (approved):
kunetuj5lki

src/formatDate.js
```diff
@@ -50,7 +50,43 @@
...
+        break;
+
+      case 'MM':
+        convert.push(month);
+        break;
+
+      case 'YY':
+        convert.push((yearFull).slice(2));
+
+        break;
```
varseniuk, line 73:
```suggestion
      case 'YY':
        convert.push((yearFull).slice(2));
        break;
```

keep the same formatting for entire code

src/formatDate.js
```diff
@@ -50,7 +50,43 @@
...
+        if (fromFormat.indexOf('YYYY') !== -1) {
+          convert.push(yearFull);
+        } else {
+          if (newArr[fromFormat.indexOf('YY')] < 30) {
+            convert.push(`20${yearShort}`);
+          } else {
+            convert.push(`19${yearShort}`);
+          }
+        }
+        break;
```
varseniuk, line 85: The code is correct but it's better not to use it in 'default' case. 
The 'default' case's purpose is to trigger when no one of the previous 'cases' hasn't matched.
In your code, all of the cases will match, and 'default' case with the check for 'YYYY' is not semantically correct.
Try to put this check outside of 'default' and add another 'case' for ir

varseniuk (changes requested):

varseniuk (approved):
Good.
But don't forget about 'default' case anyway

PR #139, opened 2020-12-20
Develop

taniazt:


MarchD (approved):

src/formatDate.js
```diff
@@ -51,6 +51,46 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
```
varseniuk, line 53:
```suggestion

```

don't leave unnecessary comments

src/formatDate.js
```diff
@@ -51,6 +51,46 @@
...
+  const separator = fromFormat[3];
+  const arrDate = date.split(separator);
+  const separNew = toFormat[3];
```
varseniuk, line 56: 'separNew' is a very strange name. Try to use more meaningful names in future

varseniuk (approved):
Clever solution. Nice

PR #140, opened 2020-12-20
Solution

ShyianO:


alexmogwaiii (approved):
Nice

src/formatDate.js
```diff
@@ -50,7 +50,45 @@
...
+      result.push(yyyy !== undefined ? yyyy : 20 + yy);
+    } else if (toFormat[j] === 'YYYY' && yy >= 30) {
+      result.push(yyyy !== undefined ? yyyy : 19 + yy);
+    } else if (toFormat[j] === 'YYYY') {
+      result.push(yyyy);
+    } else if (toFormat[j] === 'YY') {
+      result.push(
+        yy !== undefined
+          ? yy
+          : ((Math.floor(yyyy / 100) - yyyy / 100) * -100).toFixed()
```
varseniuk, line 80: Why would you use such a hard and weird calculation instead of just using the string's method 'slice' or smth else?

varseniuk (changes requested):

varseniuk (approved):

PR #144, opened 2020-12-23
solution

Vasilkoff:


src/formatDate.js
```diff
@@ -50,7 +50,61 @@
...
+  for (let i = 0; i < 3; i++) {
+    if (fromFormat[i] === 'MM') {
+      startDate.push(dateArr[i]);
+    }
+  }
+
+  for (let i = 0; i < 3; i++) {
+    if (fromFormat[i] === 'YY' || fromFormat[i] === 'YYYY') {
+      startDate.push(dateArr[i]);
+    }
```
vkryvytskyy, line 75: combine into one loop, why do you need three separate loops?

vkryvytskyy (approved):

Vasilkoff:
i did it to push the date in the needed format dd mm yyyy\yy. Firstly i
fined dd, then mm  and yy

PR #159, opened 2021-02-13
solution

chri911:


Ruslanwp (approved):
I have no doubt that solution is great. impressive way to solve it. Nice job!

Vladimir-Beliy (approved):

PR #195, opened 2021-03-14
solution

Andriykooo:


src/formatDate.js
```diff
@@ -51,6 +51,37 @@
...
+  };
+
+  if (typeof format.YYYY === 'undefined') {
+    if (format.YY < 30) {
+      format['YYYY'] = '20' + format.YY;
+    } else if (format.YY >= 30) {
+      format['YYYY'] = '19' + format.YY;
+    } else {
+      format['YYYY'] = format.YY + '\'';
+    };
```
Andrey203, line 73:
```suggestion
    if (format.YY < 30) {
      format['YYYY'] = '20' + format.YY;
    } else if (format.YY >= 30) {
      format['YYYY'] = '19' + format.YY;
    } else {
      format['YYYY'] = format.YY + '\'';
    };
```
I don't understand, why do you need so much checked. There is will be enough one "if else" to check more 30 or less.

ArtemTopolskyi (approved):

PR #199, opened 2021-03-17
Solution

PeterShp:


src/formatDate.js
```diff
@@ -50,7 +50,23 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const obj = {};
+  const result = [];
+  const str = date.split(fromFormat[3]);
```
vkryvytskyy, line 55: you have an array after split, why do you call it `str`?

vkryvytskyy (approved):

src/formatDate.js
PeterShp: changed

PR #210, opened 2021-04-06
implemented formatDate function

RomanDubovskyi:
wrote function for dynamic converting given date depending on input and output format

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -50,7 +50,57 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateMap = initDateMap(date, fromFormat);
+  let formatedDate = String();
+  const toSeparator = toFormat[3];
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    formatedDate += getDatePart(toFormat[i], dateMap) + toSeparator;
+  }
+
+  return formatedDate.substr(0, formatedDate.length - 1);
+}
+
+function initDateMap(date, fromFormat) {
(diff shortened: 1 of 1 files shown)


src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
+function getDatePart(dateMapping, map) {
+  const key = dateMapping.charAt(0);
+
+  if (key === 'Y') {
+    return convertYear(map.get(key), dateMapping);
+  }
+
+  return map.get(key);
+}
+
```
vkryvytskyy, line 90: Cложна
- сначала можно циклом пройтись по fromFormat и создать объект в котором будет день, месяц и год

```
date = {
  day: '10',
  month: '2',
  year: '2020',
}
```
После этого циклом пройтись по toFormat и искать сответствующее поле в объекте date
```
for(let i = 0; i < toFormat.length; i++) {
  switch(toFormat[i]) {
    case 'YY':
    case 'YYYY':
      // добавляем в результат нужную часть из объекта date
      break;
    case 'DD':
      // и так все три части
```

vkryvytskyy (changes requested):

src/formatDate.js
RomanDubovskyi: changed according to suggestion, with left / right indexes it was overcomplicated from my side, split looks much better. But as far as I know switch case is considered to be a bad practice because it usually violates OCP from SOLID. In this particular case, if I want to widen date format scenarios - I need to refactor switch case construction manually each time so my code is opened for modification. That's why map still looks more flexible to me.

vkryvytskyy (approved):

PR #213, opened 2021-04-24
added function formatDate

Ihor-Karpyn:


SerhiiKolesnyk (approved):
good job

src/formatDate.js
```diff
@@ -50,7 +50,37 @@
...
+  const separatorTo = toFormat[3];
+  const dateNew = date.split(separatorFrom);
+  const needReverse = fromFormat.indexOf('DD') !== toFormat.indexOf('DD');
+  const yearStart = fromFormat.includes('YY')
+    ? 'YY'
+    : 'YYYY';
+  const yearEnd = toFormat.includes('YY')
+    ? 'YY'
+    : 'YYYY';
+  const yearPosition = fromFormat.indexOf(yearStart);
```
mykola-yarchuk, line 63: there is one more approach, you can use an object to save your date parts (DD, YY, YYYY, MM), and use them instead of this count of variables.
Your solution is fine, just can be a bit simplified, if you want you can try to solve this in another way

mykola-yarchuk (approved):
good job

PR #214, opened 2021-04-24
Solution

igor-stasiv:


src/formatDate.js
```diff
@@ -50,7 +50,50 @@
...
+    return newDateFormat.join(separator);
+  }
+
+  if (fromFormat[0] !== toFormat[0] && fromFormat[2] === toFormat[2]) {
+    const newDateFormat = changeYearFormat(fromFormat, dateSplited, 0);
+
+    return newDateFormat.join(separator);
+  }
+
+  if (fromFormat[0] !== toFormat[2] && fromFormat[2] === toFormat[0]) {
```
varseniuk, line 92: these checks are not reliable at all. Now you just dealing with the first and third elements and all tests pass because all tests have 'MM' as a second element. But what if 'MM' will not be in second place? What if there will be `YY/DD/MM`

varseniuk (changes requested):

varseniuk (commented):
Let's try to find some different, more reliable solution

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+  }
+
+  const dateSplited = date.split(fromFormat[3]);
+  const newDate = {};
+
+  for (let i = 0; i < dateSplited.length; i++) {
+    switch (fromFormat[i]) {
+      case 'YY':
+      case 'YYYY':
+        newDate.year = dateSplited[i];
```
varseniuk, line 72: As a variant, you may have not created fields 'year', 'month', and 'day' but just use 'DD', 'YY' or 'MM' as keys in your `newDate` object:
```
  for (let i = 0; i < fromFormat.length - 1; i++) {
    newDate[fromFormat[i]] = dateSplited[i];
  }
```
But your variant is ok too 👍

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  function changeYearFormat(year, requiredLength) {
```
varseniuk, line 53: 🔥

varseniuk (approved):
Good job! 🚀 
Left one comment on how you can simplify your code

PR #216, opened 2021-04-26
solution

SerhiiKolesnyk:


Ihor-Karpyn (approved):
Good job.

src/formatDate.js
```diff
@@ -50,7 +50,30 @@
...
-  // write code here
+  const separ = fromFormat[3];
+  const dateArr = date.split(separ);
+  const fromDate = fromFormat.slice(0, -1);
+  const futureSepar = toFormat[3];
+  const toDate = toFormat.slice(0, -1);
+  const result = [];
+
+  for (let i = 0; i < fromDate.length; i++) {
+    for (let j = 0; j < toDate.length; j++) {
```
mykola-yarchuk, line 61: try to avoid using loop inside of loop, that's bad for performance

src/formatDate.js
```diff
@@ -50,7 +50,30 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const separ = fromFormat[3];
+  const dateArr = date.split(separ);
+  const fromDate = fromFormat.slice(0, -1);
+  const futureSepar = toFormat[3];
+  const toDate = toFormat.slice(0, -1);
```
mykola-yarchuk, line 57: there is one more approach, you can use an object to save your date parts (DD, YY, YYYY, MM), and use them.
Your solution is fine, just can be a bit simplified, if you want you can try to solve this in another way

mykola-yarchuk (approved):
it works, but it could be simplified, you can try if you want

PR #218, opened 2021-04-26
Solution

IvanVasylenko:


src/formatDate.js
```diff
@@ -50,7 +50,60 @@
...
+    if (toFormat[i] === 'MM') {
+      newDateArr.push(month);
+    }
+
+    if (toFormat[i] === 'YY') {
+      if (year.length === 2) {
+        newDateArr.push(year);
+      }
+
+      if (year.length === 4) {
```
Namart, line 88: there`re only 2 conditions. so if you are checking first one, no need to check second

Namart (approved):
good job!

src/formatDate.js
```diff
@@ -50,7 +50,60 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let day = 0;
+  let month = 0;
+  let year = 0;
```
varseniuk, line 55:
```suggestion
  let day;
  let month;
  let year;
```
These initial values have no use

src/formatDate.js
varseniuk, line 88:
```suggestion
      if (year.length === 2) {
        newDateArr.push(year);
      } else {
```
If `year.length !== 2` then obvious that `year.length === 4`

varseniuk: Anastasiia is right

src/formatDate.js
```diff
@@ -50,7 +50,60 @@
...
+        newDateArr.push(year[2] + year[3]);
+      }
+    }
+
+    if (toFormat[i] === 'YYYY') {
+      if (year.length === 4) {
+        newDateArr.push(year);
+      }
+
+      if (year.length === 2 && +year < 30) {
```
varseniuk, line 98:
```suggestion
      if (year.length === 2 && (+year < 30)) {
```
It's much more readable with parentheses

varseniuk (approved):
Nice work! 👍 
Left some comments

PR #222, opened 2021-04-27
Solution

iryna-vys:


src/formatDate.js
```diff
@@ -50,7 +50,37 @@
...
+    obj[fromFormat[i]] = arrDate[i];
+  }
+
+  const newArrDate = [];
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    if (!obj.hasOwnProperty(toFormat[i])) {
+      for (const key in obj) {
+        if (key.includes(toFormat[i])) {
+          newArrDate.push(obj[key][2] + obj[key][3]);
```
varseniuk, line 66:
```suggestion
          newArrDate.push(obj[key].slice(2));
```

varseniuk (approved):
Though your solution is clever and a bit tricky, it's really hard to read your code. Would be great to make it more simple. E.g. to get rid of nested loops (it also impacts the performance of your code), or to try to avoid `break` and `continue`.
But the logic with creating the object is good. Nice job 👍

PR #224, opened 2021-04-27
solution

RuslanaPodshyvalova:


src/formatDate.js
```diff
@@ -50,7 +50,69 @@
...
+  let yearOldFormat;
+
+  for (const item of dateArr) {
+    if (item.startsWith('Y')) {
+      yearOldFormat = item;
+    }
+  }
+
+  return yearOldFormat;
+}
```
varseniuk, line 104:
```suggestion
function defineYearFormat(dateArr) {
  let yearFormat;

  for (const item of dateArr) {
    if (item.startsWith('Y')) {
      yearFormat = item;
    }
  }

  return yearFormat;
}
```
Look, you have two similar functions here. You can make only one and name it `defineYearFormat` and this function will receive as an argument either `toFormat` or `fromFormat` and return the needed value:
```
const yearNewFormat = defineYearFormat(toFormat);
const yearOldFormat = defineYearFormat(fromFormat);
```

No need to create two functions that are doing the same job

varseniuk (changes requested):
Nice solution! 👍 
But let's remove the redundant copy of the function and it's done

varseniuk (approved):
Great job! 🔥

PR #225, opened 2021-04-27
added test for unpredictable parts order

ArtemTopolskyi:


src/formatDate.test.js
```diff
@@ -53,6 +53,17 @@ describe('formatDate', () => {
         .toBe('21-12-2012');
     });
 
+    it(`for unpredictable parts order`, () => {
```
mgrinko, line 56:
```suggestion
    it(`for non standard parts order '12-2012-21'
         from ['MM', 'YYYY', 'DD', '-']
         to ['DD', 'MM', 'YYYY', '-']`, () => {
```

mgrinko (approved):

PR #227, opened 2021-04-27
Solution

avgusta92:


src/formatDate.js
```diff
@@ -50,7 +50,42 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
-}
+  const dateObj = {};
```
Ihor-Karpyn, line 53: Dont use type in name.

Ihor-Karpyn (approved):

src/formatDate.js
```diff
@@ -50,7 +50,42 @@
...
+  let result = '';
+
+  date.split(fromFormat[3]).forEach((el, index) => {
```
varseniuk, line 56: 🔥

src/formatDate.js
varseniuk: Good advice from Ihor

varseniuk (approved):
Great job! 🚀

PR #229, opened 2021-04-27
solution

ivan-shpynda:


src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldSeparator = fromFormat[fromFormat.length - 1];
+  const newSeparator = toFormat[toFormat.length - 1];
+  let oldFormatOfYear = 0;
+  let newFormatOfYear = 0;
+
+  const oldMonth = fromFormat.indexOf('MM');
+  const oldDay = fromFormat.indexOf('DD');
```
Ihor-Karpyn, line 59:
```suggestion
  const oldMonthPosition = fromFormat.indexOf('MM');
  const oldDayPosition = fromFormat.indexOf('DD');
```
will better

Ihor-Karpyn (approved):

varseniuk (approved):
Good 👍 
There's also an approach when you can create an object that will store the old data as values and keys `DD`, `MM`, `YY` etc (`fromFormat` elements). It can reduce the number of iterations and checks and make your code faster

PR #230, opened 2021-04-27
Solution

nazarmatsevych:


SkyPKWolf (approved):

src/formatDate.js
```diff
@@ -50,7 +50,56 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let day, month, year;
+
+  const dates = date.split(fromFormat[3]);
```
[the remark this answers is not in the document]
mykola-yarchuk: there is a convention that says we should name our variable in the plural form if variable keep array

src/formatDate.js
```diff
@@ -50,7 +50,56 @@
...
+  const newDates = [];
```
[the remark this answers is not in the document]
mykola-yarchuk: same

src/formatDate.js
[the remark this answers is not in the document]
mykola-yarchuk: maybe you can call it  `dateParts`

src/formatDate.js
```diff
@@ -50,7 +50,56 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let day, month, year;
```
mykola-yarchuk, line 53: try to use object for that

mykola-yarchuk: the object is helpful if, for example, you need to use the property 'YYYY' and your object doesn't have any, you already knew that you need to transform value from 'YY' property.

mykola-yarchuk (changes requested):

src/formatDate.js
```diff
@@ -51,6 +51,38 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const fromFormatObj = {};
```
mykola-yarchuk, line 54: don't use types in names

src/formatDate.js
```diff
@@ -51,6 +51,38 @@
...
+  const fromF = fromFormat[3];
```
mykola-yarchuk, line 55: `formF`, not the best name try to use something meaningful.

src/formatDate.js
```diff
@@ -51,6 +51,38 @@
...
+  const toF = toFormat[3];
```
mykola-yarchuk, line 56: same

src/formatDate.js
```diff
@@ -51,6 +51,38 @@
...
+  const dateArr = date.split(fromF);
```
mykola-yarchuk, line 57: arr = tepe, don't use types as names if you need to call variable with array use plural form

src/formatDate.js
```diff
@@ -51,6 +51,38 @@
...
+  const res = [];
```
mykola-yarchuk, line 58:
```suggestion
  const result = [];
```

src/formatDate.js
```diff
@@ -51,6 +51,38 @@
...
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    fromFormatObj[fromFormat[i][0]] = dateArr[i];
+  }
+
+  if (fromFormat.find(el => el.startsWith('Y')).length === 2) {
+    if (fromFormatObj['Y'] > 20) {
+      fromFormatObj['Y'] = (+fromFormatObj['Y'] + 1900).toString();
+    } else {
+      fromFormatObj['Y'] = (+fromFormatObj['Y'] + 2000).toString();
+    }
```
mykola-yarchuk, line 69: use ternary operator for that
```suggestion
     fromFormatObj['Y'] = fromFormatObj['Y'] > 20
        ? (+fromFormatObj['Y'] + 1900).toString()
        :  fromFormatObj['Y'] = (+fromFormatObj['Y'] + 2000).toString();
```

src/formatDate.js
```diff
@@ -51,6 +51,38 @@
...
+  const fromF = fromFormat[3];
+  const toF = toFormat[3];
+  const dateArr = date.split(fromF);
+  const res = [];
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    fromFormatObj[fromFormat[i][0]] = dateArr[i];
+  }
+
+  if (fromFormat.find(el => el.startsWith('Y')).length === 2) {
```
mykola-yarchuk, line 64: save the result of method `find` in a variable, in that way your code gonna look cleaner

mykola-yarchuk (changes requested):

VinnieJ-2k20 (approved):

PR #232, opened 2021-04-27
Solution v0.1

SkyPKWolf:


src/formatDate.js
```diff
@@ -50,7 +50,33 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldSeparator = fromFormat.pop();
+  const newSeparator = toFormat.pop();
```
Ihor-Karpyn, line 54: You can't change function's arguments 
```suggestion
  const oldSeparator = fromFormat[3];
  const newSeparator = toFormat[3];
  const startFormat = fromFormat.slice(0, 3)
  const EndFormat = toFormat.slice(0, 3)
```

src/formatDate.js
```diff
@@ -50,7 +50,33 @@
...
+  const SplitArray = date.split(oldSeparator);
```
Ihor-Karpyn, line 55: maybe newDate or oldDate.
And dont start var's name with upper case

Ihor-Karpyn (approved):

src/formatDate.js
mykola-yarchuk: yes Ihor is right, you really can`t change function's arguments

src/formatDate.js
mykola-yarchuk: yes, agree with Ihor, don't start from the upper case and use a better name, use the word in plural form for arrays

src/formatDate.js
```diff
@@ -50,7 +50,33 @@
...
+  const dataObject = {};
+  const newDate = [];
```
mykola-yarchuk, line 57:
```suggestion
  const dateParts = [];
```

src/formatDate.js
```diff
@@ -50,7 +50,33 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldSeparator = fromFormat.pop();
+  const newSeparator = toFormat.pop();
+  const SplitArray = date.split(oldSeparator);
+  const dataObject = {};
```
mykola-yarchuk, line 56: don't use types in names, and maybe you wanted date instead of data.

mykola-yarchuk (changes requested):
almost done

src/formatDate.js
```diff
@@ -50,7 +50,35 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldSeparator = fromFormat[3];
+  const newSeparator = toFormat[3];
+  const startFormat = fromFormat.slice(0, 3);
+  const endFormat = toFormat.slice(0, 3);
```
mykola-yarchuk, line 56: that's not critical but try to figure out how to avoid these two arrays

mykola-yarchuk (approved):
fire, very good

PR #233, opened 2021-04-27
solution

dmitry-puhliakov:


src/formatDate.js
```diff
@@ -50,7 +50,41 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  // main declaration block
+  const oldDate = date.split(fromFormat[3]);
+  const newDate = [];
+
+  const y2 = 'YY';
+  const y4 = 'YYYY';
```
illisogori, line 58:
```suggestion
  const yearFormatShort = 'YY';
  const yearFormat = 'YYYY';
```

illisogori (approved):
Good job.

src/formatDate.js
varseniuk: Good advise from Igor

varseniuk (approved):
Good solution! 👍

varseniuk (commented):
There's also an approach when you can create an object that will store the old data as values and keys `DD`, `MM`, `YY` etc (`fromFormat` elements). It can reduce the number of iterations and checks and make your code faster

PR #234, opened 2021-04-28
solution

Nazar-Kudatsky:


den-bulaev:
Good, but there are too much if operators. You can use object and put all date parts there as a values, where keys will be 'DD'/'MM' etc. fromFormat items indexes === dateArr items indexes

mykola-yarchuk (approved):
great

PR #235, opened 2021-04-28
formatDate implemented

illisogori:


den-bulaev:
Cool)

mykola-yarchuk (approved):
good job, you solve this, there is one more way to solve this, you can use an object to save parts of the date, in this case, you don't need to save indexes. but as I sad your solution also good

PR #243, opened 2021-05-06
'Solution'

ViPiven:


ched88xtc (approved):

aionova (changes requested):
If you name your variables not 'obj', 'arr', 'newArr', it will get much easier to understand what's inside of them ;)
Also could you try to refactor the code to get rid of loop inside of another loop?

varseniuk (approved):
Good job! 👍

PR #298, opened 2021-08-04
formatDate was done

illiarerun:


Anna-Tkachenko (approved):
Nice Going!⭐

src/formatDate.js
```diff
@@ -50,7 +50,31 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldSeparator = fromFormat.pop();
+  const newSeparator = toFormat.pop();
```
VinnieJ-2k20, line 54: In this task it doesn't really matter, but remember that `pop()` is a mutating method.
And in this case it mutates the function arguments.
You need to avoid this if it's not mentioned directly in the task description.

src/formatDate.js
```diff
@@ -50,7 +50,31 @@
...
+    if (fromFormat[i][0] === 'Y' && toFormat.indexOf(fromFormat[i]) < 0) {
+      const oldYearFormat = fromFormat[i];
+      let yearWithoutCentury = oldDate[i];
+
+      if (oldYearFormat === 'YYYY') {
+        yearWithoutCentury = oldDate[i].slice(-2);
+
+        newDate[toFormat.indexOf('YY')] = yearWithoutCentury;
+      } else if (oldYearFormat === 'YY') {
+        const century = +oldDate[i] < 30 ? 20 : 19;
```
VinnieJ-2k20, line 68: Nice 👍

VinnieJ-2k20 (changes requested):
Good solution with only one flaw:
You need to avoid mutating function arguments.

src/formatDate.js
illiarerun: Thanks for noticing and explaining 👍  It's quite important thing, dunno how i missed that 😅

VinnieJ-2k20 (approved):
Good job 👍

PR #299, opened 2021-08-04
Solution

Anna-Tkachenko:


src/formatDate.js
```diff
@@ -50,7 +50,53 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldDateSeparator = fromFormat.pop();
+  const newDateSeparator = toFormat.pop();
```
illiarerun, line 54: Круто сделано, я не додумался pop использовать((

ihor-jpeg: ого, я аж присів коли побачив. slick move

VinnieJ-2k20, line 54: In this task it doesn't really matter, but remember that `pop()` is a mutating method.
And in this case it mutates the function arguments.
You need to avoid this if it's not mentioned directly in the task description.

VinnieJ-2k20 (changes requested):
Good solution with only one flaw:
You need to avoid mutating function arguments.

VinnieJ-2k20 (approved):
Good job 👍

PR #300, opened 2021-08-04
Solution

alex-glotov67:


Moroz-Dmytro (approved):
Looks good, but a little bit difficult because of many 'if' operators

Denchepornyuk (approved):

PR #309, opened 2021-08-07
solved

lera-yaskevych:


src/formatDate.js
```diff
@@ -50,7 +50,56 @@
...
+          newFormatDate.push(actualDate.YY);
+        }
+        break;
+      default:
+        break;
+    }
+  }
+
+  return newFormatDate.join(toFormat[3]);
 }
```
Illia-Boiko, line 104: I think this solution is clever)

Illia-Boiko (approved):
All is good

src/formatDate.js
```diff
@@ -50,7 +50,56 @@
...
+  if (fromFormat.indexOf('YYYY') === -1) {
+    indexOfYear = fromFormat.indexOf('YY');
+  } else {
+    indexOfYear = fromFormat.indexOf('YYYY');
+  }
+
+  const actualDate = {
+    DD: initialFormatDate[indexOfDay],
+    MM: initialFormatDate[indexOfMonth],
+    YY: initialFormatDate[indexOfYear],
```
VinnieJ-2k20, line 68: Not the best choice of prop names - try to follow naming convention for variables.

src/formatDate.js
```diff
@@ -50,7 +50,56 @@
...
+  };
+
+  const { YY } = actualDate;
```
VinnieJ-2k20, line 71: You can destructure all props here. This will simplify your code even further.

src/formatDate.js
```diff
@@ -50,7 +50,56 @@
...
+    switch (key) {
+      case 'DD':
+        newFormatDate.push(actualDate.DD);
+        break;
+      case 'MM':
+        newFormatDate.push(actualDate.MM);
+        break;
+      case 'YY':
+        if (YY.length === 4) {
+          newFormatDate.push(actualDate.YY[2] + actualDate.YY[1]);
```
VinnieJ-2k20, line 83: Maybe, use `slice` method here rather then add characters one by one?
Also, last two digits in year are on 3rd and 4th positions in this case.
```suggestion
          newFormatDate.push(actualDate.YY[2] + actualDate.YY[3]);
```

VinnieJ-2k20 (changes requested):
Nice solution, but it needs some code style edits.

VinnieJ-2k20 (approved):
Good job 👍

PR #310, opened 2021-08-07
Solution without loop

Illia-Boiko:


lera-yaskevych (approved):
Looks good 🔥 I like object version best, because it gives the code more structure, however both commits are understandable and I believe that is the point

Denchepornyuk (approved):

PR #313, opened 2021-08-08
solution

Vova-Stoliar:


src/formatDate.js
```diff
@@ -51,6 +51,42 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
-}
+  const dateArray = date.match(/\d+/gi);
```
alex-glotov67, line 54: RegExp 🔥

alex-glotov67 (approved):
Very interesting solution

evgen1sider (approved):
looks good

src/formatDate.js
```diff
@@ -51,6 +51,42 @@
...
+    if (dateFormatStorage[ell] === undefined) {
+      return newData;
+    }
+
+    return dateFormatStorage[ell];
+  });
+
+  const newFormat = cloneToFormat.join(`${toFormatSeparator}`);
+
+  return newFormat;
```
VinnieJ-2k20, line 90: If you create a variable and return it right away, it can be simplified: 
```suggestion
  return cloneToFormat.join(`${toFormatSeparator}`);
```

VinnieJ-2k20 (approved):
Good job 👍

PR #317, opened 2021-08-09
Solution format-date-advanced

AndriyNikorych:


Andrii-Pinchuk (commented):
Good work 💪

Andrii-Pinchuk (approved):

src/formatDate.js
```diff
@@ -50,7 +50,49 @@
...
+        break;
+
+      case 'MM':
+        newFormatOfDate[i] = month;
+        break;
+
+      case 'DD':
+        newFormatOfDate[i] = day;
+        break;
+    }
```
herkoss, line 92: All `case` blocks must have a `break` statement. Also, the `switch/case` statement always must have a `default` block. In the `default` block, we usually raise exceptions if something went wrong. formatDate function can return a string "Get wrong date format" in the `default` block

herkoss (changes requested):
Well done, but one little mistake should be fixed)
Details in comment)
Good luck!🤞🍀

herkoss (approved):
Goog job!💪👌

PR #318, opened 2021-08-09
it work

evgen1sider:


alex-glotov67 (approved):
Nice solution, but you can make it without a cycle for

src/formatDate.js
```diff
@@ -50,7 +50,42 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateFrom = date.split(fromFormat[3]);
+  const newDate = [];
+
+  const dayFormat = 'DD';
+  const monthFormat = 'MM';
+  const yearFormat = ['YY', 'YYYY'];
+
+  const day = fromFormat.indexOf(dayFormat);
```
VinnieJ-2k20, line 60: These variables store index, so correct name in this case would be smth like `dayIndex`

src/formatDate.js
```diff
@@ -50,7 +50,42 @@
...
+      if (toFormat[i] === yearFormat[1]) {
+        if (Number(dateFrom[year]) < 30) {
+          newDate[i] = (`20${dateFrom[year]}`);
+        } else {
+          newDate[i] = (`19${dateFrom[year]}`);
+        }
+      } else {
+        newDate[i] = dateFrom[year].slice(2);
+      }
+    }
```
VinnieJ-2k20, line 85: If you go this way, it's better to use `switch/case` structure here. (switch `toFormat[i]`)

VinnieJ-2k20 (changes requested):
Almost done, just need to restructure your code a little bit.

src/formatDate.js
```diff
@@ -50,7 +50,47 @@
...
+        if (Number(dateFrom[yearIndex]) < 30) {
+          newDate[i] = (`20${dateFrom[yearIndex]}`);
+        } else {
+          newDate[i] = (`19${dateFrom[yearIndex]}`);
+        }
+        break;
+      case yearFormat[0]:
+        newDate[i] = dateFrom[yearIndex].slice(2);
+        break;
+    }
```
VinnieJ-2k20, line 90: Always add default case to `switch/case`, event if it does nothing.
Usually, it logs or throws errors.

VinnieJ-2k20 (changes requested):
Only one fix left - add default case.

VinnieJ-2k20 (approved):
Good job 👍

PR #326, opened 2021-08-10
Develop

Clonvhs:
Solution

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -50,7 +50,48 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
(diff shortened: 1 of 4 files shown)


src/formatDate.js
```diff
@@ -50,7 +50,46 @@
...
+        index + 1 === toFormat.length - 1 ? timeData[item] : timeData[item]
+          + toFormat[toFormat.length - 1]
+      );
+    }
+  });
+
+  newFormat = newFormat.join('');
+
+  return newFormat;
+}
```
DmytroMykoliv, line 94: you can just return newFormat.join('');

DmytroMykoliv (approved):
Interesting solution ) good work )

Denchepornyuk (approved):

PR #352, opened 2021-09-20
solution

Saharij:
Solution

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -50,7 +50,24 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateArray = date.split(fromFormat[3]);
+  const year = dateArray[fromFormat.indexOf('YYYY')]
+        || (parseInt(dateArray[fromFormat.indexOf('YY')]) >= 30 ? '19' : '20')
+        + dateArray[fromFormat.indexOf('YY')];
+
+  const passedDate = {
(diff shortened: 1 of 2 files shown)


Bohdan-Kochetov (approved):

src/formatDate.js
```diff
@@ -50,7 +50,24 @@
...
+    'DD': dateArray[fromFormat.indexOf('DD')],
+    'MM': dateArray[fromFormat.indexOf('MM')],
+    'YYYY': year,
+    'YY': year.slice(2),
+  };
+
+  const result = [passedDate[toFormat[0]],
+    passedDate[toFormat[1]],
+    passedDate[toFormat[2]],
+  ];
```
aliceblunt8, line 68:
```suggestion
  const result = [
    passedDate[toFormat[0]],
    passedDate[toFormat[1]],
    passedDate[toFormat[2]],
  ];
```

aliceblunt8: or you can use loop for this

src/formatDate.js
```diff
@@ -50,7 +50,24 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateArray = date.split(fromFormat[3]);
```
aliceblunt8, line 53: it's better to create a variable for fromFormat[3] and toFormat[3], cause it's not obvious that they are separators

aliceblunt8 (approved):

PR #357, opened 2021-09-21
solution

Tarasulko:


src/formatDate.js
```diff
@@ -50,7 +50,72 @@
...
+
+      case 'MM':
+        month = dateItems[i];
+        break;
+
+      case 'DD':
+        day = dateItems[i];
+        break;
+    }
+  }
```
aliceblunt8, line 80: I think this for-loop is unnecessary, just use dateItems[i] in the next loop

aliceblunt8 (approved):

PR #369, opened 2021-10-10
Date was created

misharosa:


src/formatDate.js
```diff
@@ -38,19 +38,88 @@
...
  * @param {string[]} toFormat
  *
- * @returns {string}
+ * @returns {string[]}
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  if (fromFormat[0] === toFormat[0] && fromFormat[2] === toFormat[2]) {
+    return date.split(`${fromFormat[3]}`).join(`${toFormat[3]}`);
```
cyb1ove, line 54: Interpolation is not neccessary in `split` and `join` methods. Repeating in other codelines. Fix this.

src/formatDate.js
```diff
@@ -38,19 +38,88 @@
...
+  if (fromFormat[0] === toFormat[2] && fromFormat[2] === toFormat[0]) {
+    return date
+      .split(`${fromFormat[3]}`)
+      .reverse()
+      .join(`${toFormat[3]}`);
+  }
+
+  const splitDate = date.split(`${fromFormat[3]}`);
+
+  if (fromFormat[2] !== toFormat[2] && fromFormat[2] === 'YYYY') {
```
cyb1ove, line 66: You make same check in first `if` block after that proramm ending with `return` statement. You don't need a checks like `fromFormat === toFormat`. Delete this.

src/formatDate.js
```diff
@@ -38,19 +38,88 @@
...
+
+      return splitDate.reverse().join(`${toFormat[3]}`);
+    }
+
+    if (splitDate[0] < 30) {
+      splitDate[0] = '20' + splitDate[0];
+
+      return splitDate.reverse().join(`${toFormat[3]}`);
+    }
+  }
```
cyb1ove, line 110: This block is absolutely identical to previous which doesn't work. Delete this.

cyb1ove (changes requested):
Need some fixes.

PR #388, opened 2021-11-10
Solution

dariia-romanova:


src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+  if (+years < 30) {
+    return '20' + years;
+  }
+
+  return '19' + years;
+}
+
+function shortYearsFormat(years) {
+  return years.slice(2);
 }
```
dmtrhrytsak, line 86: You could make one function that would handle all edge cases with years. It's okay this way too, though👍

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+
+  for (let i = 0; i < oldDate.length; i++) {
+    if (fromFormat[i] === 'YY') {
+      dateInfo['YYYY'] = extendYearsFormat(oldDate[i]);
+    } else if (fromFormat[i] === 'YYYY') {
+      dateInfo['YY'] = shortYearsFormat(oldDate[i]);
+    }
+
+    dateInfo[fromFormat[i]] = oldDate[i];
+  }
```
dmtrhrytsak, line 67: It seems like you could iterate using `for of` or `for each` loops)

A rule of thumb: if you don't need any manipulation with indices (terminating earlier, getting the previous/next element [i + 1]/[i - 1] etc.), use either `for of` or `for each` loops or something like that)

dmtrhrytsak (approved):

src/formatDate.js
dmtrhrytsak: Oops, I overlooked the last statement😅

AlinaMatuschak: Agree. For of will be better 👌

AlinaMatuschak (approved):
Awesome 👏

src/formatDate.js
dmtrhrytsak: If ypu need to keep track of the current index, you could try `forEach` it accepts index as the third argument)

PR #392, opened 2021-11-11
Solution

vlad-tuzenko:


src/formatDate.js
```diff
@@ -50,7 +50,41 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldSeparator = fromFormat[3];
+  const newSeparator = toFormat[3];
```
alex-kostomarov, line 54: you can use `array.length - 1` instead of using `3`

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+    } else if (year.length === 2) {
+      if (year < 30) {
+        return '20' + year;
+      } else {
+        return '19' + year;
+      }
+    }
+  }
+
+  for (let i = 0; i < 3; i++) {
```
alex-kostomarov, line 83: here the same - it's better to use `array.length - 1` instead of using just a number `3`

alex-kostomarov (approved):
Cool solution!

aionova (approved):
LGTM!

PR #393, opened 2021-11-11
solution

Val-space:
all suggestions are welcome😜

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -50,7 +50,32 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldSeparator = fromFormat[3];
+  const newSeparator = toFormat[3];
+  const newDate = date.split(oldSeparator);
+  const oldDateFormat = fromFormat.slice(0, -1);
+  const newDateFormat = toFormat.slice(0, -1);
+
+  if (fromFormat[0][0] !== toFormat[0][0]) {
+    newDate.reverse();
+    oldDateFormat.reverse();
+  }
+
+  for (let i = 0; i < newDateFormat.length; i++) {
+    if (oldDateFormat[i].length < newDateFormat[i].length) {
+      if (newDate[i] < 30) {
+        newDate[i] = '20' + newDate[i];
+      } else {
+        newDate[i] = '19' + newDate[i];
+      }
+    }
+
+    if (oldDateFormat[i].length > newDateFormat[i].length) {
+      newDate[i] = newDate[i].slice(-2);
+    }
+  }
+
+  return newDate.join(newSeparator);
 }
 
 module.exports = formatDate;
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.6.7",
+    "@mate-academy/scripts": "^0.9.7",
     "eslint": "^5.16.0",
     "eslint-plugin-jest": "^22.4.1",
     "eslint-plugin-node": "^8.0.1",


src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
+  const newSeparator = toFormat[3];
+  const newDate = date.split(oldSeparator);
+  const oldDateFormat = fromFormat.slice(0, -1);
+  const newDateFormat = toFormat.slice(0, -1);
+
+  if (fromFormat[0][0] !== toFormat[0][0]) {
+    newDate.reverse();
+    oldDateFormat.reverse();
+  }
+
```
dmtrhrytsak, line 63: I have no idea what it does and how it works, but all I can say is that you really hacked the system👀
`  if (fromFormat[0][0] !== toFormat[0][0]) {
    newDate.reverse();
    oldDateFormat.reverse();
  }`

src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
+  for (let i = 0; i < newDateFormat.length; i++) {
+    if (oldDateFormat[i].length < newDateFormat[i].length) {
+      if (newDate[i] < 30) {
+        newDate[i] = '20' + newDate[i];
+      } else {
+        newDate[i] = '19' + newDate[i];
+      }
+    }
+
+    if (oldDateFormat[i].length > newDateFormat[i].length) {
```
dmtrhrytsak, line 73: How can someone guess which part of the day you're converting?👀 Dog years to human's?

dmtrhrytsak: Also your program will hit this line of code even if the first `if` was true, not a good idea😁

dmtrhrytsak (approved):
Taras Shevchenko's life would've been so much simpler if he had your date formatter with him😢  Simply awesome (prosto porashayshe)!

src/formatDate.js
dmtrhrytsak: Here I mean make it somehow obvious that years might be converted, and only years, not days, months etc.

Val-space: Agree. I didn't notice at first that we must convert only years...I was preparing for more

src/formatDate.js
Val-space: > I have no idea what it does and how it works, but all I can say is that you really hacked the system👀 ` if (fromFormat[0][0] !== toFormat[0][0]) { newDate.reverse(); oldDateFormat.reverse(); }`

I compare the first characters of old format and new format and reverse old format( if it's should be reversed) , so they both fit in the 'for loop'. That is just how I have to live with my straightforward logic;

dmtrhrytsak: As far as I can tell, you're doing the right thing, when life doesn't go your way, you make it to😂

illiarerun: So, if fromFormat will be like yyyy/mm/dd, and toFormat will be like yyyy/dd/mm, your code won't be working properly then(( Reversing a date isn't the best solution here

src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
+    oldDateFormat.reverse();
+  }
+
+  for (let i = 0; i < newDateFormat.length; i++) {
+    if (oldDateFormat[i].length < newDateFormat[i].length) {
+      if (newDate[i] < 30) {
+        newDate[i] = '20' + newDate[i];
+      } else {
+        newDate[i] = '19' + newDate[i];
+      }
```
illiarerun, line 70: Ternary operator would be a great fit here

illiarerun (changes requested):
Good job 🔥  Let's make your solution more error-proof:

src/formatDate.js
Val-space: could you add this to the tests?

aionova (approved):
GJ!

PR #396, opened 2021-11-12
Solution

olya-shyian:


dmtrhrytsak (approved):
Awesome coding style ✨ The only thing that I'd probably change is reducing the number of ifs)) I mean it's not necessary to create `month, day` variables, you could just push them directly to `result[]` array and only handling years, since they're the only one that are tricky and dynamic here 

But other than that, it's perfect! If Lesya Ukrainka was alive now, your date formatter would be the one hanging on her wall, no need for календарь anymore!

dmtrhrytsak:
Well, hold on! Maybe without ifs for all date pars it wouldn't work, but think of creating an object for every day part form old format

If I confused you with all of my suggestions, then leave it as it, it's good as it is)

aionova (approved):
LGTM!

PR #397, opened 2021-11-12
Solution

alex-kostomarov:


vlad-tuzenko (approved):
Your solution is a little tricky, but it works 👌

aionova (approved):
Looks complicated and can be refactored to get rid of loop inside of loop, but it works) So if you have time and desire, it would be nice if you refactored it

PR #398, opened 2021-11-12
Solution of format_date_advanced

Dima-Ziakhor:


JamesSunderlanddd (approved):
Well done!

ihor-jpeg (commented):
why do we need the second `for`?

ihor-jpeg (changes requested):
Please, explain why you used two `for` loops

Dima-Ziakhor:
> why do we need the second `for`?


Ok, I agree) It can be done without a second loop

aionova (changes requested):

ihor-jpeg (approved):
Good solution! You can make it even better if you use objects. When you think about it, you only need to go over `fromDate` array and to collect the key/value pairs (because `fromDate` and `date` have corresponding indices) and then just map it to the `toFormat` array.

PR #399, opened 2021-11-12
solution added

JamesSunderlanddd:
I could unite two loops that I had but then it is harder to read the code and understand what it does. So I decided to stick with both of them.

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -50,7 +50,78 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldFormatDelimeter = fromFormat[3];
+  const newFormatDelimeter = toFormat[3];
+  const newDate = date.split(oldFormatDelimeter);
+  const newDateArray = [];
+  
+  let oldYearLength;
+  let oldYearPosition;
+  let oldMonthPostion;
+  let oldDayPosition;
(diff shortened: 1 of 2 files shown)


src/formatDate.js
```diff
@@ -50,7 +50,78 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldFormatDelimeter = fromFormat[3];
```
Vitaliy-0, line 53: separator will be better

src/formatDate.js
```diff
@@ -50,7 +50,78 @@
...
+  const newFormatDelimeter = toFormat[3];
+  const newDate = date.split(oldFormatDelimeter);
+  const newDateArray = [];
+  
+  let oldYearLength;
+  let oldYearPosition;
+  let oldMonthPostion;
```
Vitaliy-0, line 60: you have no need to write this variables, in fact, you need only oldYear and newYear

Vitaliy-0 (approved):
Try to simplify code

aionova (approved):
Your solution works, but it's overcomplicated indeed and can be simplified. If you have time and desire, it would be great if you tried to refactor it. Refactoring - is a vital skill for developer

PR #403, opened 2021-11-13
Solution

SerhiiKirik:


kurtkobein (approved):
Good job

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+    }
+
+    if (fromFormat[index] === 'YYYY') {
+      dateInfo.YY = oldDateIndex.slice(-2);
+    }
+  }
+
+  for (const elem of toFormat) {
+    for (const key in dateInfo) {
+      if (elem === key) {
```
aionova, line 79: can you do it without loop in a loop?

aionova (approved):

src/formatDate.js
SerhiiKirik: I think so
I tried, but I did not succeed :(

PR #405, opened 2021-11-14
Solution

msaikun:


src/formatDate.js
```diff
@@ -50,7 +50,45 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldDate = date.split(fromFormat[3]);
+  const newFormat = [];
+  let yearShort, yearLong, month, day;
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    if (fromFormat[i] === 'YYYY') {
```
volodymyr-soltys97, line 58: Maybe should try to use a switch case in this task, but I don't know is it a good idea ;)

volodymyr-soltys97 (approved):
Very well!

ihor-jpeg (changes requested):
Why do you need two for loops here? if your format is length 3 without the separator, you will go over the format 9 times. This might be unnecessary

ihor-jpeg (approved):
your code works but it is hard to read. you can use a switch statement as your colleague suggested

PR #418, opened 2021-12-10
Solution

ABfeMate:


aliceblunt8 (changes requested):
try to simplify your solution or at least create variables inside cycles
now it's too hard to grasp

PR #425, opened 2021-12-16
Solution

Basil-Panasiuk:


src/formatDate.js
```diff
@@ -48,9 +48,44 @@
  *
  * @returns {string}
  */
-
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const arr = date.split(fromFormat[fromFormat.length - 1]);
+  const [ a, b, c ] = fromFormat;
```
Olexandr6, line 53: Rename `a, b, c` variables

src/formatDate.js
```diff
@@ -48,9 +48,44 @@
...
+  const fromArr = [a, b, c].sort();
+  const [ d, e, f ] = toFormat;
```
Olexandr6, line 55: The same case ...

src/formatDate.js
```diff
@@ -48,9 +48,44 @@
  *
  * @returns {string}
  */
-
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const arr = date.split(fromFormat[fromFormat.length - 1]);
```
Olexandr6, line 52:
```suggestion
  const dateParts = date.split(fromFormat[fromFormat.length - 1]);
```

src/formatDate.js
```diff
@@ -48,9 +48,44 @@
...
+  const [ a, b, c ] = fromFormat;
+  const fromArr = [a, b, c].sort();
+  const [ d, e, f ] = toFormat;
+  const toArr = [d, e, f].sort();
```
Olexandr6, line 56: Avoid the names of variable words `array`,` object` `string`

src/formatDate.js
```diff
@@ -48,9 +48,44 @@
...
+    }
+  }
+
+  if (toArr[2].length < fromArr[2].length) {
+    dateObj[fromArr[2]] = dateObj[fromArr[2]].slice(-2);
+  }
+
+  dateObj[toArr[2]] = dateObj[fromArr[2]];
+
+  const res = [];
```
Olexandr6, line 82: Rename ...

src/formatDate.js
```diff
@@ -48,9 +48,44 @@
...
+    } else {
+      const val = '19' + dateObj[fromArr[2]];
+
+      dateObj[fromArr[2]] = val;
+    }
+  }
+
+  if (toArr[2].length < fromArr[2].length) {
+    dateObj[fromArr[2]] = dateObj[fromArr[2]].slice(-2);
+  }
```
Olexandr6, line 78: It would be better to create a separate function for this logic.
For example: create function `formatYears`

Olexandr6 (changes requested):
Pay attention to variable naming

Olexandr6 (approved):
Good job 👍

PR #427, opened 2021-12-17
Solution

Toros-Zz:


YevheniiKorzun (approved):

src/formatDate.js
```diff
@@ -50,7 +50,36 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newDate = date.split(fromFormat[3]);
+  let result = '';
+  let index = 0;
```
Elena-brv, line 55: why do you need to create this variable before the cycle?

src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+
+    if (index === -1 && toFormat[i] === 'YY') {
+      index = fromFormat.indexOf('YYYY');
+      newDate[index] = newDate[i].slice(2);
+    }
+
+    result += newDate[index];
+
+    if (i !== toFormat.length - 2) {
+      result += toFormat[3];
```
Elena-brv, line 78: you already know method join, use it

Elena-brv (changes requested):

Elena-brv (approved):

PR #432, opened 2021-12-17
Solution

pasha-melnyk:


Bohdan-Maksymiuk (approved):

src/formatDate.js
```diff
@@ -50,7 +50,55 @@
...
+
+  const dateYearShortened = +(dateYear.slice(0, 2));
+
+  const result = [];
+
+  if (fromYearFormat.length > toYearFormat.length) {
+    dateYear = dateYear.slice(2);
+  }
+
+  if (fromYearFormat.length < toYearFormat.length
```
Elena-brv, line 77: else if?
you no need to check the second condition if the first was true (and the same further)

src/formatDate.js
```diff
@@ -50,7 +50,55 @@
...
+      result.push(dateDay);
+    }
+
+    if (item === 'MM') {
+      result.push(dateMonth);
+    }
+
+    if (item === 'YY' || item === 'YYYY') {
+      result.push(dateYear);
+    }
```
Elena-brv, line 98: `switch case` better suites here

Elena-brv (changes requested):

src/formatDate.js
pasha-melnyk: I think I need, because there is the third condition: fromYearFormat.length === toYearFormat.length, and I should do nothing if it is true. So i can't just check only one condition. The only way to avoid repeating the same code further is nesting one if inside another

src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+    dateYear = dateYear.slice(2);
+  }
+
+  if (fromYearFormat.length < toYearFormat.length) {
+    if (dateYearShortened < 30) {
+      dateYear = '20' + dateYear;
+    } else {
+      dateYear = '19' + dateYear;
+    }
+  }
```
Elena-brv, line 83: you still can use `else if` here

```suggestion
  if (fromYearFormat.length > toYearFormat.length) {
    dateYear = dateYear.slice(2);
  } else if (fromYearFormat.length < toYearFormat.length) {
      dateYear =  dateYearShortened < 30
        ? '20' + dateYear
        : '19' + dateYear;
  }
```

Elena-brv (approved):

PR #436, opened 2021-12-19
solution_v1.0

Rozdorozhnii:


src/formatDate.js
```diff
@@ -50,7 +50,35 @@
...
+  const resultArr = [];
+  const tempDate = {};
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    Object.assign(tempDate, { [fromFormat[i]]: dateArr[i] });
+  }
+
+  const fullYear = tempDate.YY < 30
+    ? 2000 + Number(tempDate.YY)
+    : 1900 + Number(tempDate.YY);
```
andrii-koptiev, line 63: Денис говорил что лучше не юзать тернарные операторы. Они плохо читаемы. Лучше if else. 
Но это пусть лучше менторы скажут

andrii-koptiev (approved):

src/formatDate.js
Rozdorozhnii: достатньо простий і очевидний прийом, як на мене

src/formatDate.js
```diff
@@ -50,7 +50,35 @@
...
+
+  if (tempDate.hasOwnProperty('YY')) {
+    Object.assign(tempDate, { 'YYYY': fullYear });
+  }
+
+  const shortYear = tempDate.YYYY > 1999
+    ? Number(tempDate.YYYY) - 2000
+    : Number(tempDate.YYYY) - 1900;
```
Elena-brv, line 71: you can just take only 2 last digits (slice), no need to check year

Elena-brv (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,31 @@
...
+  const dateArr = date.split(fromFormat[fromFormat.length - 1]);
+  const resultArr = [];
+  const tempDate = {};
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    tempDate[fromFormat[i]] = dateArr[i];
+  }
+
+  if (tempDate.hasOwnProperty('YYYY')) {
+    tempDate['YY'] = tempDate['YYYY'].toString().split('').slice(2).join('');
```
Elena-brv, line 62:
```suggestion
    tempDate['YY'] = tempDate['YYYY'].slice(2);
```

src/formatDate.js
```diff
@@ -50,7 +50,31 @@
...
+  }
+
+  const fullYear = tempDate.YY < 30
+    ? 2000 + Number(tempDate.YY)
+    : 1900 + Number(tempDate.YY);
```
Elena-brv, line 67: no need to do it, if there is no YYYY on the new format

src/formatDate.js
```diff
@@ -50,7 +50,31 @@
...
+  const resultArr = [];
+  const tempDate = {};
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    tempDate[fromFormat[i]] = dateArr[i];
+  }
+
+  if (tempDate.hasOwnProperty('YYYY')) {
+    tempDate['YY'] = tempDate['YYYY'].toString().split('').slice(2).join('');
+  }
```
Elena-brv, line 63: no need to do it, if there is no YY on the new format

Elena-brv (changes requested):

src/formatDate.js
Rozdorozhnii: I wrote universal function which collect all possible format and should be able to work with all possible format in result

src/formatDate.js
Rozdorozhnii: I wrote universal function which collect all possible format and should be able to work with all possible format in result

src/formatDate.js
Rozdorozhnii: When I tried slice type of Number, it wasn't work. Now try again and it's work fine. Didn't understood why yet.

src/formatDate.js
Elena-brv: you need to create a function in that way to avoid unnecessary operations

Elena-brv (approved):

PR #441, opened 2021-12-24
Solution date

YuriTaraday:


src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
+    YY = YYYY.slice(-2);
+  } else {
+    YY = year;
+
+    if (YY < 30) {
+      YYYY = '20' + YY;
+    } else {
+      YYYY = '19' + YY;
+    }
+  }
```
aliceblunt8, line 84: this looks weird, maybe you can do this inside one of the if-conditions?

aliceblunt8 (changes requested):

vkryvytskyy (approved):

PR #515, opened 2022-04-05
Solution

serhii-naumenko:


src/formatDate.js
```diff
@@ -50,7 +50,74 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const [,,, oldSeparator] = fromFormat;
+  const [,,, newSeparator] = toFormat;
```
YegorKochetkov, line 54:
```suggestion
  const oldSeparator = fromFormat.pop();
  const newSeparator = toFormat.pop();
```
it's easier that way, isn't it?

YegorKochetkov (approved):
interesting solution)

src/formatDate.js
serhii-naumenko: This mutates the original array. Maybe then it's better to use const oldSeparator = fromFormat[4] ?

YegorKochetkov: Ok)

src/formatDate.js
```diff
@@ -50,7 +50,74 @@
...
+    switch (fromFormat[i]) {
+      case 'YY': {
+        year = oldDateArr[i];
+        break;
+      }
+
+      case 'YYYY': {
+        year = oldDateArr[i];
+        break;
+      }
```
AlinaMatuschak, line 72:
```suggestion
      case 'YYYY',
      case 'YY': {
        year = oldDateArr[i];
        break;
      }
```

src/formatDate.js
```diff
@@ -50,7 +50,74 @@
...
+        newDateArr[i] = year;
+        break;
+      }
+
+      case 'YYYY': {
+        if (year.length === 2) {
+          if (+year < 30) {
+            year = '20' + year;
+          } else {
+            year = '19' + year;
```
AlinaMatuschak, line 100: What is 30, '20' and '19'? Magic values

AlinaMatuschak (approved):

src/formatDate.js
serhii-naumenko: It's from terms of task: "When converting from YYYY to YY just use 2 last digit (1997 -> 97).
 *   When converting from YY to YYYY use 20YY if YY < 30 and 19YY otherwise."

PR #527, opened 2022-04-08
solut

KostiaO:


src/formatDate.js
```diff
@@ -48,9 +48,107 @@
...
+      } else {
+        num1 = `19${startDate[0]}`;
+      }
+    } else {
+      num1 = startDate[0];
+      num3 = startDate[2];
+    }
+    num3 = startDate[2];
+  //   newStr.push(num1);
+  }
```
Bohdan-Maksymiuk, line 76: it is better way instead using so much if's everywhere

Bohdan-Maksymiuk (changes requested):
Your solution is too complicated.
You could try to use object in your solution, it should be easier to implement
```
  const dateObj = {
    'DD': null,
    'MM': null,
    'YY': null,
    'YYYY': null,
  };
  ```
  for example

PR #537, opened 2022-04-27
solution

LukaszGryziewicz:


59MrRobot (approved):
Good job, Lukasz!

src/formatDate.js
```diff
@@ -50,7 +50,66 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparator = fromFormat.pop();
+  const dateArr = date.split(fromSeparator);
+  let year = '';
+  let month = '';
+  let day = '';
```
meforoma, line 57: you could just declare all three without initializing:
```suggestion
  let year, month, day;
```

meforoma (changes requested):
great code. Already approvable.

Yet if you want - you could approach it using a dateObject (and benefit from `key: value` structure)

Hint:
```
  for (let i = 0; i < 3; i++) {
  }
```

ihor-jpeg (approved):

PR #538, opened 2022-04-27
solution

efdalyalcin:


src/formatDate.js
```diff
@@ -50,7 +50,31 @@
...
+        result[j] = dateArr[i];
+      } else if (fromFormat[i] === 'YYYY' && toFormat[j] === 'YY') {
+        result[j] = dateArr[i].slice(-2);
+      } else if (fromFormat[i] === 'YY' && toFormat[j] === 'YYYY') {
+        if (dateArr[i] < 30) {
+          result[j] = `20${dateArr[i]}`;
+        } else {
+          result[j] = `19${dateArr[i]}`;
+        }
+      }
```
59MrRobot, line 73: What about continuing using if statements here to make your code more readable?

src/formatDate.js
```diff
@@ -50,7 +50,31 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const iconSplit = fromFormat[3];
+  const iconJoin = toFormat[3];
```
59MrRobot, line 54: You and I know where this 3 comes from but what about someone else? Right now you are using a 'magic number'

59MrRobot (approved):
Good job, Efdal. Just a couple suggestions

59MrRobot (changes requested):
I made a mistake, my bad

meforoma (changes requested):
non-trivial logic - and that's a 👍 👍 
yet nested `for` loops increase algorithm complexity.

You can leave it as-is - it's approvable.
Or you may practice objects and use benefits of key-value pairs.

Hint:
```
  for (let i = 0; i < 3; i++) {
    dateObj[fromFormat[i]] = dateSplit[i];
  }
```

ihor-jpeg (approved):
well done!

ihor-jpeg (approved):

VinnieJ-2k20 (approved):

PR #539, opened 2022-04-28
Solution

iLebo-stack:


src/formatDate.js
```diff
@@ -50,7 +50,51 @@
...
+
+    if (fromFormat[i].includes('D')) {
+      day = dateArray[i];
+    }
+  }
+
+  for (let i = 0; i < dateArray.length; i++) {
+    if (toFormat[i].includes('Y')) {
+      if (toFormat[i].length === 4) {
+        if (parseInt(year.slice(-2)) < 30) {
```
huberciak, line 78: You can also use + sign what is a bit shorter but less visible, to be honest I dont know what is better option here

huberciak (approved):
One short comment but code clear. Good job! I am approving.

src/formatDate.js
iLebo-stack: Its a tiny detail but I'll change it

src/formatDate.js
```diff
@@ -50,7 +50,51 @@
...
+      }
+    }
+
+    if (toFormat[i].includes('M')) {
+      result[i] = month;
+    }
+
+    if (toFormat[i].includes('D')) {
+      result[i] = day;
+    }
```
VinnieJ-2k20, line 94: Could use switch/case here instead of first level of `if`s

VinnieJ-2k20 (approved):
Good job 👍

PR #540, opened 2022-04-28
Solution

amitpatiljc:


src/formatDate.js
```diff
@@ -50,7 +50,58 @@
...
+      year = dateArray[i];
+    }
+
+    if (fromFormat[i] === 'MM') {
+      month = dateArray[i];
+    }
+
+    if (fromFormat[i] === 'DD') {
+      day = dateArray[i];
+    }
```
59MrRobot, line 78: If you used switch statement below, why not also use it here? 🤔

59MrRobot (changes requested):
Good work, Amit - just one suggestion

src/formatDate.js
amitpatiljc: Yes... you are right!

59MrRobot (approved):
Great work! 🔥

meforoma (changes requested):
it's fine, yet takes a bit much lines of code.

Try approaching it using an object and benefits of it's keys.

Hint: 
```
  for (let i = 0; i < 3; i++) {
    dateObj[fromFormat[i]] = dateSplit[i];
  }
```

amitpatiljc:
> it's fine, yet takes a bit much lines of code.
> 
> Try approaching it using an object and benefits of it's keys. Essentially generate a date object (from `fromFormat`) and later assemble a new date from it.
> 
> Hint:
> 
> ```
>   for (let i = 0; i < 3; i++) {
>     dateObj[fromFormat[i]] = dateSplit[i];
>   }
> ```

meforoma (approved):
👍

PR #541, opened 2022-04-28
Ultrasonic automatic AI algorithm for date conversion

andreisucman:


WojciechCzarnowski:
Nice job. Looking good, if a bit long with all the explenations. Patching you through to the higher ups.

src/formatDate.js
```diff
@@ -50,7 +50,55 @@
...
+  // happy checking
+
+  const inputArray = date.split('');
+  const result = [];
+
+  for (let i = 0; i < inputArray.length; i++) {
+    if (isNaN(inputArray[i] / 2)) {
+      inputArray[i] = ' ';
+    }
+  }
```
ihor-jpeg, line 62: im not sure i understand what this is for. it looks like the way to remove the separator then you can simply pop it from the from format and split by that to get a clean array

src/formatDate.js
```diff
@@ -50,7 +50,55 @@
...
+    // change the current date to integer
+    threeElements[fromFormat.indexOf('YYYY')]
+    = parseInt(threeElements[fromFormat.indexOf('YYYY')]);
+
+    const currentDate = threeElements[fromFormat.indexOf('YYYY')];
+
+    if (currentDate < 30) {
+      threeElements[fromFormat.indexOf('YYYY')] += 2000;
+    } else if (currentDate < 100) {
+      threeElements[fromFormat.indexOf('YYYY')] += 1900;
```
ihor-jpeg, line 91: there is a lot of indexOf. each time that method is called it needs to traverse the entire array. accessing object values through keys is much more efficient. 

to me, a much simpler and faster solution would involve an object like 

{
  'DD': 13,
  'MM': 03,
  'YY': 2000,
}

ihor-jpeg (changes requested):

src/formatDate.js
ihor-jpeg: it's ontional tho

src/formatDate.js
andreisucman: Okay fixed

src/formatDate.js
andreisucman: Okay, I get your point. But changing to your proposed logic will require me to rethink this task all over again from scratch and for now, I would like to focus on learning other topics to not to lag behind

VinnieJ-2k20 (approved):

src/formatDate.js
```diff
@@ -50,7 +50,48 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  // happy checking
+  const fromSeparator = fromFormat.pop();
+  const toSeparator = toFormat.pop();
```
VinnieJ-2k20, line 55: Sorry, missed from first check.
Don't mutate incoming arrays.

VinnieJ-2k20 (changes requested):

src/formatDate.js
andreisucman: Ok, fixed.

ihor-jpeg (approved):

ihor-jpeg (approved):

ihor-jpeg (approved):

PR #542, opened 2022-04-28
Solution

huberciak:


andreisucman (approved):
Your code is fire, Hubert. 🥇

meforoma (changes requested):
@huberciak i love that you're utilizing a `dateObject`, yet it's a bit unproductive.
for `dateObject` keys you're using custom `month / day / year` - why not using those already provided in parameters? (YY | YYYY / MM / DD)

Hint:
```
  for (let i = 0; i < 3; i++) {
  }
```

VinnieJ-2k20 (approved):
Good job 👍

PR #547, opened 2022-04-30
Task complete

rHashTag:
d(0o,)b   >>>

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -51,6 +51,54 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const fromSeperator = fromFormat.pop();
+  const dateArray = date.split(fromSeperator);
+  let dateYear, dateMonth, dateDay;
+  const output = [];
+
+  for (let i = 0; i < fromFormat.length; i++) {
+    if (fromFormat[i] === 'YYYY' || fromFormat[i] === 'YY') {
+      dateYear = dateArray[i];
+    }
+
+    if (fromFormat[i] === 'MM') {
+      dateMonth = dateArray[i];
+    }
+
+    if (fromFormat[i] === 'DD') {
+      dateDay = dateArray[i];
+    }
+  }
+
+  const toSeperator = toFormat.pop();
+
+  if (dateYear.length === 2) {
+    if (parseInt(dateYear) < 30) {
+      dateYear = `20${dateYear}`;
+    } else {
+      dateYear = `19${dateYear}`;
+    }
+  }
+
+  for (let i = 0; i < toFormat.length; i++) {
+    if (toFormat[i] === 'YYYY') {
+      output[(toFormat.indexOf(toFormat[i]))] = dateYear;
+    }
+
+    if (toFormat[i] === 'YY') {
+      output[(toFormat.indexOf(toFormat[i]))] = dateYear.slice(2);
+    }
+
+    if (toFormat[i] === 'MM') {
+      output[(toFormat.indexOf(toFormat[i]))] = dateMonth;
+    }
+
+    if (toFormat[i] === 'DD') {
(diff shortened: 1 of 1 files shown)


src/formatDate.js
```diff
@@ -51,6 +51,54 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const fromSeperator = fromFormat.pop();
+  const dateArray = date.split(fromSeperator);
+  let dateYear, dateMonth, dateDay;
+  const output = [];
```
MabuzaM, line 57: I think output can be anything. Maybe a more meaningful variable, based on its value is better.

src/formatDate.js
```diff
@@ -51,6 +51,54 @@
...
+    if (fromFormat[i] === 'MM') {
+      dateMonth = dateArray[i];
+    }
+
+    if (fromFormat[i] === 'DD') {
+      dateDay = dateArray[i];
+    }
+  }
+
+  const toSeperator = toFormat.pop();
```
MabuzaM, line 73: I think this line can be moved to the top where other variables are defined.

MabuzaM (changes requested):
Hi Rashaad, Great work. Please check have left some suggestions.

src/formatDate.js
rHashTag: Done!

src/formatDate.js
rHashTag: Done.

MabuzaM: Great

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const toSeperator = toFormat[toFormat.length - 1];
+  const fromSeperator = fromFormat.pop();
```
VinnieJ-2k20, line 54: Don't mutate incoming array.

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
+  const dateArray = date.split(fromSeperator);
+  let dateYear, dateMonth, dateDay;
```
VinnieJ-2k20, line 56: Declare each variable on the separate line.
```suggestion
  let dateYear;
  let dateMonth;
  let dateDay;
```

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
+      dateOutput[dateOutput.length] = dateYear.slice(2);
+    }
+
+    if (toFormat[i] === 'MM') {
+      dateOutput[dateOutput.length] = dateMonth;
+    }
+
+    if (toFormat[i] === 'DD') {
+      dateOutput[dateOutput.length] = dateDay;
+    }
```
VinnieJ-2k20, line 96: Great place for switch case. You check only one parameter and you have limited number of outcomes.

VinnieJ-2k20 (changes requested):
Almost done, Some improvements can be made.

src/formatDate.js
rHashTag: Done.

src/formatDate.js
VinnieJ-2k20: Need to fix this as well.

src/formatDate.js
VinnieJ-2k20: Not fixed.

VinnieJ-2k20 (changes requested):
You fixed one comment, but forgot to fix two more.

src/formatDate.js
```diff
@@ -50,7 +50,55 @@
...
```
ihor-jpeg, line 98: gotta use that switch

ihor-jpeg (changes requested):

VinnieJ-2k20 (approved):

PR #548, opened 2022-04-30
Jstasks

MabuzaM:


src/formatDate.js
```diff
@@ -50,7 +50,63 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparator = fromFormat.pop();
+  const toSeparator = toFormat.pop();
+  const dateToArr = date.split(fromSeparator);
+  const newDate = [];
+  let year, month, dte;
```
59MrRobot, line 57: What is `dte`? You should use a more descriptive name imo

59MrRobot (changes requested):
Good job, just one comment

src/formatDate.js
MabuzaM: Sho thanx :)

src/formatDate.js
```diff
@@ -50,7 +50,63 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparator = fromFormat.pop();
+  const toSeparator = toFormat.pop();
```
VinnieJ-2k20, line 54: Don't mutate incoming arrays.

VinnieJ-2k20 (changes requested):
Almost done.
One small but important change left.

src/formatDate.js
MabuzaM: Thank you. I have made changes. I don't understand why it's failing to work with something like this  
```
 const toSeparator = toFormat[3];
```

Please help.

VinnieJ-2k20: Well, I suppose that you iterate through full length of the array. It reaches the last element and your `switch/case` structure throws an error.

VinnieJ-2k20 (approved):
Good job 👍

PR #553, opened 2022-05-01
too many if solution

ZUBOV-ILLIA:


valerii-derkach (approved):
Good job!

src/formatDate.js
```diff
@@ -50,7 +50,32 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateArr = date.split(fromFormat[3]);
+  let resultArr = [];
```
Anna-Tkachenko, line 54: Don't use name of data types in variable names

Anna-Tkachenko (approved):
Good job!
Will be better to make some checks or some calculations at the start of the function and then combine the results in needed format
One cycle and many if / else statement - not so good solution)
If you'll have time, try to rewrite it

PR #554, opened 2022-05-01
format date

Andriyk0:


yelnikov-andrii (approved):
Approved

Anna-Tkachenko (approved):
Will be better to get main values from the function arguments at the start of the function, make some checks or calculations and then combine the results in needed date
If you have time, try to rewrite it in some way

PR #556, opened 2022-05-01
solution

valerii-derkach:


vlasiuk-anatolii (approved):
Good job!

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+      fromDate[0] = '20' + fromDate[0];
+    } else {
+      fromDate[0] = '19' + fromDate[0];
+    }
+    fromFormat[0] = 'YYYY';
+  }
+
+  if (fromFormat[2] === 'YY'
+  && (toFormat[0] === 'YYYY' || toFormat[2] === 'YYYY')) {
+    if (fromDate[0].toString() > 31) {
```
Anna-Tkachenko, line 68: If you have some similar checks, you can create separate variable (for example for (toFormat[0] === 'YYYY' || toFormat[2] === 'YYYY') and use it in your code

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+  }
+
+  if (fromFormat[0] === 'YYYY'
+  && (toFormat[0] === 'YY' || toFormat[2] === 'YY')) {
+    fromDate[0] = fromDate[0].slice(-2);
+    fromFormat[0] = 'YY';
+  }
+
+  if (fromFormat[2] === 'YYYY'
+  && (toFormat[0] === 'YY' || toFormat[2] === 'YY')) {
```
Anna-Tkachenko, line 83: Same here with similar checks

Anna-Tkachenko, line 83: Same here with similar checks

Anna-Tkachenko (approved):
⭐️

PR #557, opened 2022-05-01
formatDate was done

vlasiuk-anatolii:


src/formatDate.js
```diff
@@ -49,8 +49,35 @@
...
  */
 
-function formatDate(date, fromFormat, toFormat) {
-  // write code here
+ function formatDate(date, fromFormat, toFormat) {
+  let string = '';
+  const arrDate = date.split(fromFormat[3]);
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    const a = fromFormat.indexOf(toFormat[i]);
```
valerii-derkach, line 57: you should use semantic names of variables to understand what do they mean

valerii-derkach (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let string = '';
+  const arrDate = date.split(fromFormat[3]);
```
Anna-Tkachenko, line 54: Don't use name of data types in variable names

Anna-Tkachenko (approved):
⭐️

PR #558, opened 2022-05-01
Solution

iwanttosmile4u:


valerii-derkach (approved):
Good job!

src/formatDate.js
```diff
@@ -50,7 +50,50 @@
...
+        day = splittedDate[i];
+        break;
+
+      case 'M':
+        month = splittedDate[i];
+        break;
+
+      case 'Y':
+        year = splittedDate[i];
+        break;
```
Anna-Tkachenko, line 71: Each switch / case statement should have default case

Anna-Tkachenko (approved):
⭐️

PR #559, opened 2022-05-02
format Date is ready

Uktam19980416:


src/formatDate.js
```diff
@@ -51,6 +51,52 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const arr = [];
+  const arrSplit = date.split(fromFormat.pop());
+  const arrToFormat = toFormat.pop();
+  let year = '',
+    month = '',
+    day = '';
```
meforoma, line 59: you can just declare without initializing:
```suggestion
  let year, month, day;
```

src/formatDate.js
```diff
@@ -51,6 +51,52 @@
...
+
+  for (let idx = 0; idx < fromFormat.length; idx++) {
```
meforoma, line 61: `idx` is bit weird, common iterators are `i / j`
although that's not an error, yet better get used to common conventions

meforoma (changes requested):
your implementation is decent.

Yet you might want to practice implementation using object benefits.

Hint:
```
  for (let i = 0; i < 3; i++) {
  }
```

src/formatDate.js
```diff
@@ -51,6 +51,52 @@
...
+    month,
+    day;
+
+  for (let i = 0; i < fromFormat.length; i++) {
+    if (fromFormat[i] === 'YY') {
+      if (+arrSplit[i] < 30) {
+        year = `20${arrSplit[i]}`;
+      } else {
+        year = `19${arrSplit[i]}`;
+      }
```
ihor-jpeg, line 67: maybe use ternary

src/formatDate.js
```diff
@@ -51,6 +51,52 @@
...
+      year = arrSplit[i];
+    }
+
+    if (fromFormat[i] === 'MM') {
+      month = arrSplit[i];
+    }
+
+    if (fromFormat[i] === 'DD') {
+      day = arrSplit[i];
+    }
```
ihor-jpeg, line 80: if you use an object to store data, you won't need so many checks

ihor-jpeg (approved):

VinnieJ-2k20 (approved):

PR #560, opened 2022-05-02
made my date format

uliton:


Anna-Tkachenko (approved):
Good job! ⭐️
If you'll have some time, you can try to rewrite your code
Just try to make some checks and calculations at the start of your function, and then using one cycle combine the results in needed format
Such way will be better than 2 cycles with many if / else statements

PR #561, opened 2022-05-02
Date adv

proznerka:


MabuzaM (changes requested):
Hi Karina. Good job. Have left a few suggestions below.

src/formatDate.js
```diff
@@ -50,7 +50,43 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let day = '';
+  let month = '';
+  let year = '';
+  let result = '';
+  const splits = date.split(fromFormat[3]);
+
```
MabuzaM, line 58: Let's assign this index of a "/" to a variable for it to be meaningful. And maybe use another variable instead of splits, maybe "dateParts" or "dateArray".

src/formatDate.js
```diff
@@ -50,7 +50,43 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let day = '';
+  let month = '';
+  let year = '';
+  let result = '';
```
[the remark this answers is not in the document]
proznerka: Variables like this shouldn't be declared in one line. Because it makes code less readable.

src/formatDate.js
proznerka: That I can change.

src/formatDate.js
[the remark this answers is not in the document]
MabuzaM: okay

src/formatDate.js
MabuzaM: okay

src/formatDate.js
```diff
@@ -50,7 +50,45 @@
...
+      } else if (toFormat[i].length === 2 && year.length === 4) {
+        year = year.slice(2);
+      }
+
+      result += year + toSeparator;
+    } else if (toFormat[i][0] === 'M') {
+      result += month + toSeparator;
+    } else {
+      result += day + toSeparator;
+    }
```
ihor-jpeg, line 88: a lot of nested ifs

ihor-jpeg (approved):

VinnieJ-2k20 (approved):

PR #562, opened 2022-05-02
seems finished

WojciechCzarnowski:


andreisucman (approved):
Good job.

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const arrDate = [];
+  const splitDate = date.split(fromFormat.pop());
```
VinnieJ-2k20, line 54: Don`t mutate incoming arrays.

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
+  const dateFormat = toFormat.pop();
+
+  let year,
+    month,
+    day;
```
VinnieJ-2k20, line 59: Better to declare variables separately.
```suggestion
  let year;
  let month;
  let day;
```

VinnieJ-2k20 (changes requested):
Almost done. Couple of small changes left.

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const arrDate = [];
+  const splitDate = date.split([...fromFormat].pop());
+  const dateFormat = toFormat.pop();
```
VinnieJ-2k20, line 55: This should not be mutated as well.

VinnieJ-2k20 (changes requested):
Small advice.
I can leave a comment referring to certain line of code, but that means you should fix this detail all over your project, not just in that line :)

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const arrDate = [];
+  const splitDate = date.split([...fromFormat].pop());
```
VinnieJ-2k20, line 54: Really hard to read. Split to separate operations.

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
+  const dateFormat = [...toFormat.pop()];
```
VinnieJ-2k20, line 55:
```suggestion
  const dateFormat = [...toFormat].pop();
```

VinnieJ-2k20 (changes requested):
Very small changes left.

PR #565, opened 2022-05-05
solution

Sardorbek1122:


src/formatDate.js
```diff
@@ -51,6 +51,53 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const result = [];
+
+  const fromSeparator = fromFormat.pop();
+  const inputDateArray = date.split(fromSeparator);
+  const newSeparator = toFormat.pop();
```
VinnieJ-2k20, line 58: Don't mutate incoming arrays.
```suggestion
  const fromSeparator = [...fromFormat].pop();
  const inputDateArray = date.split(fromSeparator);
  const newSeparator = [...toFormat].pop();
```

src/formatDate.js
```diff
@@ -51,6 +51,53 @@
...
+      if (toFormat[i].length === 4) {
+        if (+year.slice(-2) < 30) {
+          result[i] = `20${year.slice(-2)}`;
+        } else {
+          result[i] = `19${year.slice(-2)}`;
+        }
+      } else {
+        result[i] = year.slice(-2);
+      }
+    }
```
VinnieJ-2k20, line 89: Better to avoid heavily nested `if` structures.
For now - leave it like this.

VinnieJ-2k20 (changes requested):
Just one thing left to fix - avoid mutation of incoming data.

meforoma (approved):

PR #581, opened 2022-06-02
Solution

frontend112:


meforoma (changes requested):
the code logic is pretty hard to follow to be honest
may i suggest you to approach task using a dateObject (and benefit from `key: value` structure)

Hint:
```
const dateSplit = date.split(fromSeparator);
const dateObject = {};

  for (let i = 0; i < 3; i++) {
  }
```

src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
+  if (year.length === 2) {
+    // check is it before 2000 or after
+    const thousands = year >= 30 ? 19 : 20;
+
+    //  YY => YYYY
+    newYear = thousands + newYear;
+  }
+
+  return newYear;
 }
```
ihor-jpeg, line 99: write less code
```suggestion
function convertYear(year) {
  if (year.length === 4) {
    // YYYY => YY
    return year.charAt(2) + year.charAt(3);
  }
  
  const thousands = year >= 30 ? 19 : 20;

  //  YY => YYYY
  return thousands + year;
}
```

ihor-jpeg (approved):

meforoma (approved):

PR #585, opened 2022-06-04
Solution

ik-web:


src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
+      newFormat.push(year);
+    }
+
+    if (el[0] === 'M') {
+      newFormat.push(month);
+    }
+
+    if (el[0] === 'D') {
+      newFormat.push(day);
+    }
```
illiarerun, line 96: follow DRY - don't repeat yourself, you basically do the same operation in this 2 if statements, so combine them

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let year = '';
+  let month = '';
+  let day = '';
+
+  const oldFormat = date.split(fromFormat[3]);
```
illiarerun, line 57: It'd be better to create variable for `fromFormat[3]`, e.g. `oldSeparator`
```suggestion
  const oldFormat = date.split(fromFormat[3]);
```

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
+    if (el[0] === 'M') {
+      newFormat.push(month);
+    }
+
+    if (el[0] === 'D') {
+      newFormat.push(day);
+    }
+  });
+
+  return newFormat.join(toFormat[3]);
```
illiarerun, line 99: same here - `newSeparator`

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let year = '';
+  let month = '';
+  let day = '';
+
+  const oldFormat = date.split(fromFormat[3]);
+
+  fromFormat.forEach((el, i) => {
```
illiarerun, line 59: smth like `datePart` would be more fitting
```suggestion
  fromFormat.forEach((datePart, i) => {
```

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
+    }
+
+    if (el[0] === 'D') {
+      day = oldFormat[i];
+    }
+  });
+
+  const newFormat = [];
+
+  toFormat.forEach((el) => {
```
illiarerun, line 75: same here

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
+    if (el[0] === 'Y') {
+      if (year.length > el.length) {
+        year = year.slice(2);
+      } else if (year.length < el.length) {
+        if (+year < 30) {
+          year = '20' + year;
+        } else {
+          year = '19' + year;
+        }
```
illiarerun, line 84: you can create a variable for century value, so you wouldn't need to repeat yourself with `year = n + year`, you'll do it only once at the end

illiarerun: and also it'd be great to use ternary for that)

illiarerun (changes requested):
GJ , now let's work on some fixes:

src/formatDate.js
ik-web: Fixed

src/formatDate.js
ik-web: Fixed

src/formatDate.js
ik-web: Fixed
Thanks for your help once more!

PR #590, opened 2022-06-08
Solution

mira1zu:


src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+    dateObj[fromFormat[i][0]] = dateSplit[i];
+  }
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    currentUnit = toFormat[i];
+    currentElement = dateObj[currentUnit[0]];
+
+    if (currentUnit.length < currentElement.length) {
+      currentElement = currentElement.slice(-2);
+    } else if (currentUnit.length > currentElement.length) {
```
jstmpelowycz, line 71: if that's OK when `currentUnit.length === currentElement.length`, you can include it in one of the conditions, by this avoiding `else-if` statement, just `else` statement

src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+      if (currentElement < 30) {
+        currentElement = '20' + currentElement;
+      } else {
+        currentElement = '19' + currentElement;
+      }
```
illiarerun, line 76: could've used ternary operator for it

illiarerun (approved):
Great job 🔥

src/formatDate.js
jstmpelowycz, line 76: nesting of conditions always a bad idea, so let's make some var like
```
const prefix = currentElement < 30 ? 20 : 19;
```
and then add before `currentElement`

jstmpelowycz (approved):
good job! take a look at what I've proposed

PR #591, opened 2022-06-08
First commit/ solution

Kaunin1:


NikitaKarbazin (approved):

NikitaKarbazin:
Nice solution!

illiarerun (changes requested):
GJ, let's just apply one little fix:

Kaunin1:
Good sugestions are  appreciated !
Code was simplified  by implementation of additional variables , redability is improved.

src/formatDate.js
```diff
@@ -50,7 +50,44 @@
...
+
+      case 'YYYY':
+        if (fromFormat.includes('YYYY')) {
+          formatedDate.push(yearLong);
+        } else {
+          if (arrayDate[fromFormat.indexOf('YY')] >= 30) {
+            formatedDate.push('19' + yearShort);
+          } else {
+            formatedDate.push('20' + yearShort);
+          };
```
jstmpelowycz, line 85: I'd rather create a separate var to define our prefix (_19_ or _20_), and then apply it to `yearShort`
by doing this way, we avoid extra conditions nesting

jstmpelowycz (approved):

PR #609, opened 2022-06-12
Solution

AndriiNychaiuk:


MariiaNH (approved):
Great job, Andrii - everything's work well :) Approved!

As for me, it's easier when there is one cycle in function - but it's how you would prefer. 

As well I think it's ok, but just interested which practice is better for a code style: if the cycle's parameters are the same - is it better to write checks for different objects in the same cycle or in different ones? Hope mentor will answer.

src/formatDate.js
```diff
@@ -50,7 +50,65 @@
...
+      case 'YYYY':
+        correcOrderedDate.push(fixedYear);
+        break;
+    }
+  }
+
+  return correcOrderedDate.join(toFormat[3]);
 }
 
+function fixedYears(input, required) {
```
Olexandr6, line 103: How do we name functions?
```suggestion
function fixedYears(input, required) {
```

src/formatDate.js
```diff
@@ -50,7 +50,65 @@
...
+
+      case 'MM':
+        month = dateParts[i];
+        break;
+
+      case 'YY':
+      case 'YYYY':
+        unfixedYear = dateParts[i];
+        break;
+    }
```
Olexandr6, line 73: You do the same operation in all cases. Let's create an array `newDateFormat` and push into it some date parts, without this `switch cases `

src/formatDate.js
```diff
@@ -50,7 +50,65 @@
...
+      case 'MM':
+        correcOrderedDate.push(month);
+        break;
+
+      case 'YY':
+      case 'YYYY':
+        correcOrderedDate.push(fixedYear);
+        break;
+    }
+  }
```
Olexandr6, line 98: The same case here, you don't need switch case statements here, you do the same operation in all cases but with different values, let's simplify this solution

Olexandr6 (changes requested):

src/formatDate.js
AndriiNychaiuk: I have no ideas how to reorder dates without switch cases :(

Olexandr6 (approved):
Okay, it's looks much better👍

PR #610, opened 2022-06-12
Solution

illavodonis:


src/formatDate.js
```diff
@@ -50,7 +50,50 @@
...
+      }
+
+      result.push(year);
+    }
+  }
+
+  return result.join(newSeperator);
+}
+
+function foundIndex(oldFormatItems, partDate, newFormatItems) {
```
yanaost, line 91: not sure you need newFormatItems...

yanaost (approved):
nice)
but it was a bit hard to read the code

src/formatDate.js
```diff
@@ -50,7 +50,50 @@
...
-  // write code here
+  const oldFormat = [...fromFormat];
+  const newFormat = [...toFormat];
+  const oldSeperator = oldFormat.pop();
+  const newSeperator = newFormat.pop();
+  const dateItems = date.split(oldSeperator);
+  const result = [];
+
+  for (let i = 0; i < dateItems.length; i++) {
+    if (newFormat[i].includes('D')) {
```
Olexandr6, line 61: I would rather check a string value, include('D') not the best choice here

src/formatDate.js
```diff
@@ -50,7 +50,50 @@
...
+
+      if (oldFormatYear.length < newFormatYear.length) {
+        const millennium = year < 30 ? 20 : 19;
+
+        year = millennium + year;
+      }
+
+      result.push(year);
+    }
+  }
```
Olexandr6, line 86: In this case, would be better to use switch(...) case

Olexandr6 (changes requested):

Olexandr6 (approved):

PR #611, opened 2022-06-12
solution

AKomashchuk:


mykola-koval (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,45 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const separator = String(toFormat[3]);
+  const dateSplit = date.split(fromFormat[3]);
+  const changeFormat = fromFormat.splice(0, 3).reduce((acc, item, index) => {
```
illiarerun, line 55: don't use mutating methods with function params, it may lead to unexpected behavior, you probably wanted to use `slice` here

illiarerun (approved):
GJ, left one comment

PR #617, opened 2022-06-13
solved

Grodh06:


src/formatDate.js
```diff
@@ -48,9 +48,46 @@
...
+  for (let i = 0; i < divideDate.length; i++) {
+    dateMap[fromFormat[i]] = divideDate[i];
+  }
+
+  if (!dateMap['YY']) {
+    dateMap['YY'] = dateMap['YYYY'].slice(-2);
+  }
+
+  if (!dateMap['YYYY']) {
+    dateMap['YYYY'] = +dateMap['YY'] >= 30 ? `19${dateMap['YY']}` : `20${dateMap['YY']}`;
```
Olexandr6, line 67: Looks better, agree?
```suggestion
    dateMap['YYYY'] = +dateMap['YY'] >= 30
      ? `19${dateMap['YY']}` 
      : `20${dateMap['YY']}`;
```

src/formatDate.js
```diff
@@ -48,9 +48,46 @@
...
+
+      case 'YY':
+        newDate[i] = dateMap['YY'];
+        break;
+
+      default:
+        newDate[i] = dateMap['YYYY'];
+        break;
+    }
+  }
```
Olexandr6, line 88: You don't need to switch cases here. In all cases you do the same operation, but with different values. Let's simplify this and don't use any if or switch case clauses.

Hint: use `for in` loop to go though your `dateMap` object

Olexandr6 (changes requested):
Your solution is good👍, but let's work on improvements

Olexandr6 (approved):

PR #635, opened 2022-07-01
js_format_date_advanced

YevhenVeselov:


src/formatDate.js
```diff
@@ -50,7 +50,56 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dataArr = date.split(fromFormat[3]);
+  const result = [];
+  let dd;
+  let mm;
+  let yyyy;
```
Bohdan-Maksymiuk, line 57: better use object for these properties

Bohdan-Maksymiuk (changes requested):

PR #639, opened 2022-07-02
Solution

fledith:


src/formatDate.js
```diff
@@ -50,7 +50,44 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const pastSeparator = fromFormat.pop();
+  const futureSeparator = toFormat.pop();
```
VETALEK, line 54: don't mutate the input arrays

VETALEK (changes requested):
Good start 💪

VETALEK (approved):
🚀

PR #643, opened 2022-07-03
solution

oleksandr-kotliarov:


src/formatDate.js
```diff
@@ -50,7 +50,48 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const firstSeperator = fromFormat.pop();
+  const secondSeperator = toFormat.pop();
```
VETALEK, line 54: you change your arguments, and if, by some reason, we need to use them, they won't have their last elements 😭

VETALEK (changes requested):
Good start 👍

src/formatDate.js
oleksandr-kotliarov: pushed these values before return

src/formatDate.js
```diff
@@ -50,7 +50,51 @@
...
+        break;
+
+      case 'YYYY':
+        formated.push(values.YYYY);
+        break;
+    }
+  });
+
+  fromFormat.push(firstSeperator);
+  toFormat.push(secondSeperator);
```
VETALEK, line 95: nice, but you may do it much simpler, just make
`const firstSeperator = fromFormat[fromFormat.length - 1]`
or 
`const firstSeperator = fromFormat.slice().pop()`

VETALEK (changes requested):
Almost 💪

VETALEK (approved):
🔥

PR #658, opened 2022-07-07
solution

OlgaStakhova:


src/formatDate.js
```diff
@@ -50,7 +50,72 @@
...
+  const separFrom = fromFormat[fromFormat.length - 1];
+  const toFormatArr = toFormat;
+  const separTo = toFormat[toFormat.length - 1];
+  const dateArr = date.split(separFrom);
+
+  const result = [0, 0, 0];
+
+  let day;
+  let month;
+  let year;
```
AleksandrYuschenko, line 63: Можна перерахувати через кому

src/formatDate.js
```diff
@@ -50,7 +50,72 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromFormatArr = fromFormat;
```
AleksandrYuschenko, line 53: не розумію для чого копія

src/formatDate.js
```diff
@@ -50,7 +50,72 @@
...
+        result[idxTo] = year;
+
+        return;
+      }
+
+      if (year.length === 4) {
+        result[idxTo] = year[2] + year[3];
+      }
+    }
+  });
```
AleksandrYuschenko, line 114: Switch maybe better:)

AleksandrYuschenko (approved):
👍

src/formatDate.js
ivan-kadykalo: Really, it is useless. You just copy that value to new constant

src/formatDate.js
ivan-kadykalo: In that case, it will be less readable. Leave it as it is.

src/formatDate.js
```diff
@@ -50,7 +50,72 @@
...
+
+        return;
+      }
+
+      if (year.length === 2) {
+        if (+year < 30) {
+          year = '20' + year;
+        } else {
+          year = '19' + year;
+        }
```
ivan-kadykalo, line 104: avoid using nested if statements

ivan-kadykalo (changes requested):

Ihor-Karpyn (approved):
LGTM

PR #660, opened 2022-07-08
solution

AleksandrYuschenko:


LordAsmodey (approved):
Interesting solution

src/formatDate.js
```diff
@@ -51,6 +51,31 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
```
ivan-kadykalo, line 53: delete comments

src/formatDate.js
```diff
@@ -51,6 +51,31 @@
...
+  const dateInArr = date.split(fromFormat[fromFormat.length - 1]);
+  const dateOutArr = [];
+
+  for (const keyto of toFormat) {
+    for (const keyfrom of fromFormat) {
```
ivan-kadykalo, line 58: Loop-in-the-loop is a very resource-intensive decision.  In more complicated tasks, execution may overflow the stack.

ivan-kadykalo (approved):
GJ💪

PR #665, opened 2022-07-10
solution

roman-romanenko:


src/formatDate.js
```diff
@@ -50,7 +50,27 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateFormat = date.split(fromFormat[fromFormat.length - 1]);
```
yuliia-dronova, line 53: `fromFormat[fromFormat.length - 1] ` is usually recommended to replace into separate varriable) 
For example, `const oldSeparator =  fromFormat[fromFormat.length - 1] `

yuliia-dronova (approved):
Looks very neat, GJ :)

PR #666, opened 2022-07-10
develop

nikita-biriukov:


denys-danyliuk (approved):
good job🚀

src/formatDate.js
```diff
@@ -48,9 +48,46 @@
...
+        newDate += '19' + year + toFormat[3];
+      }
+    } else if (format === 'YY' && year.length === 4) {
+      newDate += year.slice(2) + toFormat[3];
+    } else if (format === 'MM') {
+      newDate += month + toFormat[3];
+    } else if (format === 'DD') {
+      newDate += day + toFormat[3];
+    } else {
+      newDate += year + toFormat[3];
```
ivan-kadykalo, line 86: I think it would be easier to solve this task with an object and its keys

ivan-kadykalo (approved):
You make a very complicated if-else statement, but anyway, good job💪

PR #667, opened 2022-07-10
formatDate was created

yuliia-dronova:


roman-romanenko (approved):
You can use switch instead if in last loop 😉 But also looks good 😊

src/formatDate.js
```diff
@@ -50,7 +50,56 @@
...
+  const month = currentDate[fromFormat.indexOf(monthMask)];
+  const shortYear = currentDate[yearIndex].slice(-2);
+
+  let longYear = '';
+
+  if (+shortYear < 30) {
+    longYear = `20${shortYear}`;
+  } else {
+    longYear = `19${shortYear}`;
+  }
```
ivan-kadykalo, line 82:
```suggestion
  const longYear = +shortYear < 30 ? `20${shortYear}` : `19${shortYear}`;
```

ivan-kadykalo (changes requested):

ivan-kadykalo (approved):
GJ🚀

PR #671, opened 2022-07-11
Solution

yaroslav-lizogub:


src/formatDate.js
```diff
@@ -50,7 +50,51 @@
...
+      return `20${year}`;
+    }
+
+    return `19${year}`;
+  };
+
+  const getShortYear = (year) => {
+    return year.slice(2);
+  };
+
```
ivan-kadykalo, line 71: It would be better to take these functions out from formatDate function so that they are not created every time the formatDate is called.
```suggestion
```

src/formatDate.js
```diff
@@ -50,7 +50,51 @@
  */
```
ivan-kadykalo, line 51:
```suggestion
  const getFullYear = (year) => {
    return +year < 30 ? `20${year}` : `19${year}`
  };

  const getShortYear = (year) => {
    return year.slice(2);
  };

```

ivan-kadykalo (changes requested):

Ihor-Karpyn (approved):

PR #672, opened 2022-07-11
Solution_format_date_advanced

kata96451:


src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
+      formatedDate.push(splitedDate[dayIndex]);
+    }
+
+    if (toFormat[i] === 'MM') {
+      formatedDate.push(splitedDate[monthIndex]);
+    }
+
+    if (toFormat[i] === 'YYYY') {
+      if (fromFormat.includes('YY')) {
+        if (year < 30) {
```
illnino380, line 80: Three nested if's a little bit complicated to read) But it works)

illnino380 (approved):
It is a little bit complicated to read years' conditions in the loop, but there are a lot of conditions for them) Good job)

src/formatDate.js
ivan-kadykalo, line 80: Double nested if😱

> [CODE STYLE] - avoid using nested if statements, we are sure that you can do it without them :)

src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
+          year = '19' + year;
+        }
+      }
+
+      formatedDate.push(year);
+    }
+
+    if (toFormat[i] === 'YY') {
+      if (fromFormat.includes('YYYY')) {
+        year = year.slice(2);
```
ivan-kadykalo, line 92: Here the same. Fix it

ivan-kadykalo (changes requested):

ivan-kadykalo (approved):
🔥

PR #674, opened 2022-07-11
Solution

blcts:


src/formatDate.js
```diff
@@ -51,6 +51,34 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const newdate = date.split(`${fromFormat[fromFormat.length - 1]}`);
```
chernobrovin, line 54: you shood use 'camel case'
```suggestion
  const newDate = date.split(`${fromFormat[fromFormat.length - 1]}`);
```

chernobrovin (changes requested):
Your solution is perfect, but you need to fix your variable's names

chernobrovin (approved):
Perfect!

Ihor-Karpyn (approved):

PR #676, opened 2022-07-12
Solution

Danil-Hodytskyi:


ivan-kadykalo (changes requested):
What about student reviews?

blcts (approved):
nice! 👌

src/formatDate.js
```diff
@@ -50,7 +50,45 @@
...
+      case 'YY':
+        formatedDate.push(shortYear);
+
+        break;
+
+      case 'YYYY':
+        formatedDate.push(longYear);
+
+        break;
+    }
```
ivan-kadykalo, line 88: You should always have default case

ivan-kadykalo (approved):
GJ🤘
Fix that comment

src/formatDate.js
Danil-Hodytskyi: I added default case.

PR #691, opened 2022-07-21
advanced date

Blenemy:


src/formatDate.js
```diff
@@ -50,7 +50,42 @@
...
+
+    objDate[index] = splittedDate[i];
+  }
+
+  if (objDate.YYYY.length === 0) {
+    if (objDate.YY >= 30) {
+      objDate.YYYY = '19' + objDate.YY;
+    } else {
+      objDate.YYYY = '20' + objDate.YY;
+    }
```
valerii-derkach, line 75: better use ternary operator here

PR #700, opened 2022-08-01
Solution

aleksandra-prognimak:


src/formatDate.js
```diff
@@ -51,6 +51,35 @@
...
+    objDate[fromFormat[i]] = arrDate[i];
+  }
+
+  if (objDate['YY'] < 30) {
+    objDate['YYYY'] = 20 + objDate['YY'];
+  }
+
+  if (objDate['YY'] >= 30) {
+    objDate['YYYY'] = 19 + objDate['YY'];
+  }
```
sergii-nosachenko, line 72:
```suggestion
  if (objDate['YY']) {
    objDate['YYYY'] = (objDate['YY'] < 30 ? 20 : 19) + objDate['YY'];
  }
```

This part can be simplified under one common condition.

src/formatDate.js
```diff
@@ -51,6 +51,35 @@
...
+
+  if (objDate['YYYY'] !== undefined) {
+    objDate['YY'] = objDate['YYYY'].slice(2, 4);
+  }
```
sergii-nosachenko, line 76: This part will rewrite your existed property with the same value. I'll explain: before this lines you check if (oldDateObject['YY']) exists and if true create oldDateObject['YYYY']. Then here you check if (oldDateObject['YYYY']) exists - of course it exists as you've already created it.
You need to improve the logic of these blocks to avoid unnecessary runs.

Simply use else statement.

src/formatDate.js
```diff
@@ -51,6 +51,35 @@
...
   // write code here
+  const newDate = [];
+  const oldSeparator = fromFormat[3];
+  const newSeparator = toFormat[3];
+
+  const arrDate = date.split(oldSeparator);
+
+  const objDate = {};
+
+  for (let i = 0; i <= 2; i++) {
```
sergii-nosachenko, line 62:
```suggestion
  for (let i = 0; i < 3; i++) {
```

Try to use same logic across your code. Or you loop through array keys before some index, or inclusive (in identical use cases).

sergii-nosachenko (changes requested):
Hi, Sasha)
As always, good work. But can be even better 🦾

sergii-nosachenko (changes requested):
Review your condition - it has unwanted turn.

sergii-nosachenko (approved):
Everything's fine 👌

AndriiRut (approved):
🚀

PR #707, opened 2022-08-06
Solution

mariabramnik:


src/formatDate.js
```diff
@@ -51,6 +51,84 @@
...
+    }
+  } else if (indMonFirstTo === 0) {
+    res += dataMon + substTo;
+
+    if (toFormatStr[toFormatStr.length - 1] === 'D') {
+      res += dataYearNew + substTo + dataDay;
+    } else {
+      res += dataDay + substTo + dataYearNew;
+    }
+  }
```
Bohdan-Maksymiuk, line 130: very complicated and hard to read

Bohdan-Maksymiuk (changes requested):
You can create object

```
 const dateObj = {
	DD: '',
	MM: '',
	YY: '',
	YYYY: '',
  }; 
```

 Then 2 loops:
  
  ```
    const result = [];  
    
    result.push(dateObj[toFormat[i]])
   ```
   
   and then we join array element with separator

stanislav-topikha (approved):

PR #710, opened 2022-08-08
Solution

Black3942:


src/formatDate.js
```diff
@@ -50,7 +50,33 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldDateArr = date.split(fromFormat[3]);
+  const dateObj = {
+    [fromFormat[0]]: oldDateArr[0],
+    [fromFormat[1]]: oldDateArr[1],
+    [fromFormat[2]]: oldDateArr[2],
```
ArtemTopolskyi, line 57: lets avoid hardcoding indices, what if we will replace different parts of date?

ArtemTopolskyi (changes requested):

src/formatDate.js
Black3942: corrected

alexandra-protyanova (approved):

PR #716, opened 2022-08-09
Develop

KorniychukLesya:


src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+    fromFormat[index] = 'YY';
+    dataArray[index] = dataArray[index].slice(2);
+  }
+
+  if (fromFormat.includes('YY') && toFormat.includes('YYYY')) {
+    const index = fromFormat.indexOf('YY');
+
+    fromFormat[index] = 'YYYY';
+
+    if (dataArray[index] < 22) {
```
sergii-nosachenko, line 69:
```suggestion
    if (dataArray[index] < 30) {
```
Comparison doesn't met task condition.

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const separator = toFormat[3];
+  const dataArray = date.split(fromFormat[3]);
+  let toDate = '';
+
+  if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
+    const index = fromFormat.indexOf('YYYY');
+
+    fromFormat[index] = 'YY';
```
sergii-nosachenko, line 60:
```suggestion
    // fromFormat[index] = 'YY'; <-- This is a bad practice!
```
Here you've modified function argument. It's forbidden. Also it's an array - so your change affected original object too.

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+    dataArray[index] = dataArray[index].slice(2);
+  }
+
+  if (fromFormat.includes('YY') && toFormat.includes('YYYY')) {
+    const index = fromFormat.indexOf('YY');
+
+    fromFormat[index] = 'YYYY';
```
sergii-nosachenko, line 67:
```suggestion
```
Not acceptable too.

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+  let toDate = '';
+
+  if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
+    const index = fromFormat.indexOf('YYYY');
+
+    fromFormat[index] = 'YY';
+    dataArray[index] = dataArray[index].slice(2);
+  }
+
+  if (fromFormat.includes('YY') && toFormat.includes('YYYY')) {
```
sergii-nosachenko, line 64:
```suggestion
  } else {
```
No need to write inverted conditions - just use else.

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+
+  for (let i = 0; i < dataArray.length; i++) {
+    const index = fromFormat.indexOf(toFormat[i]);
+
+    toDate += dataArray[index];
+
+    if (i !== 2) {
+      toDate += separator;
+    }
+  }
```
sergii-nosachenko, line 84:
```suggestion
  toDate = dataArray.join(separator);
```

Can be simplified a lot. If you have array and separator you can use join method without writing it from scratch.

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+    const index = fromFormat.indexOf('YY');
+
+    fromFormat[index] = 'YYYY';
+
+    if (dataArray[index] < 22) {
+      dataArray[index] = '20' + dataArray[index];
+    } else {
+      dataArray[index] = '19' + dataArray[index];
+    }
+  }
```
sergii-nosachenko, line 74:
```suggestion
   dataArray[index] = ( <ternary logic here> ) + dataArray[index];
```
Try to rewrite this in one simple line using ternary operator.

sergii-nosachenko (changes requested):
Hi, Lesya! ✌
Your code is working but has some issues and can be optimized a lot.

src/formatDate.js
KorniychukLesya: corrected

src/formatDate.js
KorniychukLesya: corrected

src/formatDate.js
KorniychukLesya: corrected

src/formatDate.js
KorniychukLesya: corrected

sergii-nosachenko (approved):
Looks better 👌

danylo-onishchuk (approved):

PR #719, opened 2022-08-10
solution

Githi54:


src/formatDate.js
```diff
@@ -50,7 +50,59 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dataObj = {
+    day: null,
+    month: null,
+    year: null,
```
OlehMy, line 56: It seems no need for that. Enough just create object)

OlehMy (changes requested):
Not bad. Only one little comment))
And as whole, it isn't the easiest solution. As well as mine)))

OlehMy (approved):

src/formatDate.js
```diff
@@ -50,7 +50,55 @@
...
+        if (toFormatYear.length === fromFormatYear.length) {
+          newDate.push(dataObj.year);
+        } else if (toFormatYear.length < fromFormatYear.length) {
+          newDate.push(dataObj.year.slice(2));
+        } else {
+          if (fromFormatYear >= 30) {
+            newDate.push(`19${dataObj.year}`);
+          } else {
+            newDate.push(`20${dataObj.year}`);
+          }
```
AlinaMatuschak, line 96: you can use ternary operator

src/formatDate.js
```diff
@@ -50,7 +50,55 @@
...
+        dataObj.month = arrDate[i];
+        break;
+
+      default:
+        dataObj.year = arrDate[i];
+    }
+  }
+
+  const newDate = [];
+  const fromFormatYear = dataObj.year;
```
AlinaMatuschak, line 72: destructuring )

AlinaMatuschak (approved):
Good job 👏

PR #726, opened 2022-08-12
Solution

OlehMy:


src/formatDate.js
```diff
@@ -50,7 +50,33 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const resArr = Array(3);
```
Githi54, line 53: understandable names!👍

src/formatDate.js
```diff
@@ -50,7 +50,33 @@
...
+  const monthNewPosition = toFormat.findIndex(e => e.startsWith('M'));
+  const yearNewPosition = toFormat.findIndex(e => e.startsWith('Y'));
+
+  if (fromFormat[yearPosition].length === 2
+    && toFormat[yearNewPosition].length === 4) {
+    dateArr[yearPosition] = (dateArr[yearPosition] < 30 ? 20 : 19)
+    + dateArr[yearPosition];
+  }
+
+  resArr[dayNewPosition] = dateArr[dayPosition];
```
Githi54, line 70: I think resArr it's doesn't matter varriable. Maybe, can you use only dateArr?

Githi54 (approved):
Good job!🦾 But small recommendation 😊

src/formatDate.js
```diff
@@ -50,7 +50,33 @@
...
+  const resultArr = Array(3);
+  const oldSeparator = fromFormat[3];
+  const newSeparator = toFormat[3];
+  const dateArr = date.split(oldSeparator);
+  const dayPosition = fromFormat.findIndex(e => e.startsWith('D'));
+  const monthPosition = fromFormat.findIndex(e => e.startsWith('M'));
+  const yearPosition = fromFormat.findIndex(e => e.startsWith('Y'));
+  const dayNewPosition = toFormat.findIndex(e => e.startsWith('D'));
+  const monthNewPosition = toFormat.findIndex(e => e.startsWith('M'));
+  const yearNewPosition = toFormat.findIndex(e => e.startsWith('Y'));
```
dellachaise, line 62: You could try to rewrite your code.

Create empty objectData for example and then add there this values using `for` loop.

This way would be kind of shorter.

dellachaise (approved):
Well done 👍

src/formatDate.js
OlehMy: Thanks) This is really a good idea))

PR #727, opened 2022-08-12
Solution

Anastasiia-Khmelovska:


src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+    const index = copyFromFormat.indexOf('YY');
+
+    copyFromFormat[index] = 'YYYY';
+
+    if (arrDate[index] < 30) {
+      arrDate[index] = 20 + arrDate[index];
+    } else {
+      arrDate[index] = 19 + arrDate[index];
+    }
+  }
```
AndreyJigijalkin, line 74: Instead of doing all these actions you can create object with keys from "from format" and values from your "arrDate". Then you can make only 3 test for your "obj.YYYY" or "obj.YY", which will start only if fromFormat and toFormat year records are different

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+
+  for (let i = 0; i < toFormat.length; i++) {
+    for (let j = 0; j < toFormat.length - 1; j++) {
+      if (copyFromFormat[j] === toFormat[i]) {
+        result += arrDate[j] + toFormat[3];
+      }
+    }
+  }
```
AndreyJigijalkin, line 82: This loops may also be simplified a bit

AndreyJigijalkin (commented):
I will not either reject or approve it yet. I'm sure, that you can make your code a bit more elegant:)

Anastasiia-Khmelovska:
> I will not either reject or approve it yet. I'm sure, that you can make your code a bit more elegant:)

Thank you very much for such a detailed comment!
yes, I agree with you, this solution would be easier, but I don't take the easy way out)))
I saw such solutions on github and I wanted to solve it without objects) I was specifically looking for such a way) if it is fundamental, it will not be difficult (I guess) to remake it, but I think this solution also might be.

your approve will be a good present for my birthday))
I really hope for your understanding)

AndreyJigijalkin (approved):
Happy b-Day🥰

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+  } else {
+    const index = copyFromFormat.indexOf('YY');
+
+    copyFromFormat[index] = 'YYYY';
+
+    if (arrDate[index] < 30) {
+      arrDate[index] = 20 + arrDate[index];
+    } else {
+      arrDate[index] = 19 + arrDate[index];
+    }
```
AlinaMatuschak, line 73: Ternary operator )

AlinaMatuschak (changes requested):
Agree with @AndreyJigijalkin 😉

AlinaMatuschak (approved):
Nice attempt 👏

PR #728, opened 2022-08-13
function formatDate was added

Andrii-Stiebalov:
I enjoyed this task immensely. And I really want to know how to improve my solution. 🐥

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -48,9 +48,35 @@
  *
  * @returns {string}
  */
-
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const arrDate = date.split(fromFormat.pop());
+
+  if (toFormat.includes('YYYY') && fromFormat.includes('YY')) {
+    const indexY = fromFormat.indexOf('YY');
+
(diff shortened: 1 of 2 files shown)


src/formatDate.js
```diff
@@ -48,9 +48,35 @@
...
+
+  if (toFormat.includes('YY') && fromFormat.includes('YYYY')) {
+    const indexY = fromFormat.indexOf('YYYY');
+
+    arrDate[indexY] = arrDate[indexY].slice(2);
+    fromFormat[indexY] = 'YY';
+  }
+
+  const obj = arrDate.reduce(function(target, key, index) {
+    const rigihtIndex = toFormat.indexOf(fromFormat[index]);
```
plemiannik, line 72: right

plemiannik (approved):
I don`t know what is reduce, so i google it, but i still never used it, so it looks quite complicated 4 me. But, generally, in my point of view, i think ur solution rather right, then false. Aspecialy, given that the tests are correct.

dellachaise (approved):
🚀

PR #737, opened 2022-08-15
Solution

kir8mir:


Andrii-Stiebalov (approved):
Nice job. Cool that you used objects. 👏

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
+
+    from['YYYY'] = from['YYYY'] < 30
+      ? '20' + from['YYYY'] : '19' + from['YYYY'];
+  } else if (toFormat.includes('YY')
+  && fromFormat.includes('YYYY')) {
+    from['YY'] = from['YYYY'];
+    delete from['YYYY'];
+
+    from['YY'] = from['YY'].slice(2);
+  }
```
dellachaise, line 74: Try to simply check if `from.YYYY` instead of includes and 2 conditions and then `else`. For this task it would be enough.

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const arr = [];
+  const newDate = date.split(fromFormat[fromFormat.length - 1]);
+
+  const from = {
+    [fromFormat[0]]: newDate[0],
+    [fromFormat[1]]: newDate[1],
+    [fromFormat[2]]: newDate[2],
+  };
```
dellachaise, line 60: You could generate this object, using `for` loop.

dellachaise (approved):
Great job 👍

PR #740, opened 2022-08-15
Solution

max-kravchenko:


src/formatDate.js
```diff
@@ -50,7 +50,44 @@
...
+        newDate += rawDate[i].toString().slice(-2);
+      }
+
+      if (i === oldFormat.length - 1
+        && newFormat[j] !== oldFormat[i]
+        && newFormat[j] === 'YYYY') {
+        if (+rawDate[0] < 30) {
+          newDate += '20' + rawDate[0];
+        } else {
+          newDate += '19' + rawDate[0];
```
Ohiiko, line 80: Try to use ternary operator here, instead of if else.

src/formatDate.js
```diff
@@ -50,7 +50,44 @@
...
+
+  for (let j = 0; j < newFormat.length; j++) {
+    for (let i = 0; i < oldFormat.length; i++) {
+      if (newFormat[j] === oldFormat[i]) {
+        newDate += rawDate[i];
+        break;
+      }
+
+      if (i === oldFormat.length - 1
+        && newFormat[j] !== oldFormat[i]
```
Ohiiko, line 69: I think, given that the condition is repeated, it is better to make a separate if

max-kravchenko: I am not sure if I got you right, but I managed to shorten the code :) Thanks!

Ohiiko (approved):
Good job.

src/formatDate.js
```diff
@@ -50,7 +50,38 @@
...
+  const oldSeparator = fromFormat[3];
+  const newSeparator = toFormat[3];
+  const oldFormat = fromFormat.splice(0, 3);
+  const newFormat = toFormat.splice(0, 3);
+  const rawDate = date.split(oldSeparator);
+
+  let newDate = '';
+
+  for (let j = 0; j < newFormat.length; j++) {
+    for (let i = 0; i < oldFormat.length; i++) {
```
AlinaMatuschak, line 62: loop inside loop 😬
I'm sure that you can better )

src/formatDate.js
```diff
@@ -50,7 +50,38 @@
...
+        newDate += rawDate[i];
+        break;
+      }
+
+      if (i === oldFormat.length - 1
+        && newFormat[j] === 'YY') {
+        newDate += rawDate[i].toString().slice(-2);
+      }
+
+      if (i === oldFormat.length - 1
```
AlinaMatuschak, line 73: Can you save in variable `oldFormat.length - 1` value?

AlinaMatuschak: You calculate three times 🥲

src/formatDate.js
```diff
@@ -50,7 +50,38 @@
...
+        && newFormat[j] === 'YYYY') {
+        newDate += +rawDate[0] < 30 ? '20' + rawDate[0] : '19' + rawDate[0];
```
AlinaMatuschak, line 75:
```suggestion
        newDate += (+rawDate[0] < 30 ? '20' : '19') + rawDate[0];
```

AlinaMatuschak (approved):
Good attempt, but I believe that you can better

you can create date object with format keys and values. And use one loop to set it:

{
  'YYYY': 2020,
  'MM': 02,
  ...
}

And after use one more loop to generate result

src/formatDate.js
max-kravchenko: Thanks:)

PR #741, opened 2022-08-15
format date solution

roma-shafihullin:


src/formatDate.js
```diff
@@ -50,7 +50,29 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateToArr = date.split(fromFormat[3]);
```
Andrii-Stiebalov, line 53: The variable's name should say **what** is stored in it, and not **"how"** or **" where"**.
For example `arrOfDate`; 😄

Andrii-Stiebalov (approved):
Cool job!

PR #742, opened 2022-08-15
Solution

Andrii-Dunets:


Andrii-Stiebalov (approved):
Good job 👏

src/formatDate.js
```diff
@@ -50,7 +50,37 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateArr = date.split(`${fromFormat[3]}`);
```
AlinaMatuschak, line 53: I suggest that it anyway is a string 

```suggestion
  const dateArr = date.split(fromFormat[3]);
```

src/formatDate.js
```diff
@@ -50,7 +50,37 @@
...
+  for (let i = 0; i <= dateArr.length; i++) {
+    dateObj[fromFormat[i]] = dateArr[i];
+  }
+
+  if (toFormat.includes('YY')) {
+    dateObj['YY'] = dateObj['YYYY'].slice(2);
+  }
+
+  if (toFormat.includes('YYYY')) {
+    if (!dateObj.hasOwnProperty('YYYY')) {
```
AlinaMatuschak, line 66: And will be nice to create a separate constant for it

```suggestion
  if (!dateObj.hasOwnProperty('YYYY') && toFormat.includes('YYYY')) {
```

src/formatDate.js
```diff
@@ -50,7 +50,37 @@
...
+      if (dateObj['YY'] < 30) {
+        dateObj['YYYY'] = `20${dateObj['YY']}`;
+      } else {
+        dateObj['YYYY'] = `19${dateObj['YY']}`;
+      }
```
AlinaMatuschak, line 71: Looks like a good place for the ternary operator

src/formatDate.js
```diff
@@ -50,7 +50,37 @@
...
+    }
+  }
+
+  for (let a = 0; a < toFormat.slice(0, 3).length; a++) {
```
AlinaMatuschak, line 75: When you slice elements from 0 to 3 index, you know exactly length that will be here 
```suggestion
  for (let a = 0; a < 3; a++) {
```

But should create a separate variable for 3. Now It will be a magic number

src/formatDate.js
```diff
@@ -50,7 +50,37 @@
...
+    finalDate += dateObj[toFormat[a]];
+
+    if (a !== toFormat.slice(0, 3).length - 1) {
```
AlinaMatuschak, line 78: Don’t need two times use `toFormat.slice(0, 3).length`. You do the same on 75 and here. Should create variable

AlinaMatuschak (changes requested):
Good work 👏, just small issues 👇

dellachaise (approved):
Well done 👍

PR #749, opened 2022-08-19
format_date_advanced Done

sanyokmalyshev:


src/formatDate.js
```diff
@@ -50,7 +50,50 @@
...
+      }
+
+      if (obj.Y.length === 2) {
+        if (obj.Y < 30) {
+          newArr.push('20' + obj.Y);
+        } else {
+          newArr.push('19' + obj.Y);
+        }
+      }
+    }
```
Bohdan-Maksymiuk, line 83: avoid such nesting `else if` conditions

Bohdan-Maksymiuk (changes requested):
You can create object
 ```   
   const dateObj = {
	DD: '',
	MM: '',
	YY: '',
	YYYY: '',
  };
 ``` 
 Then 2 loops: 
  
  ``` 
    const result = [];  result.push(dateObj[toFormat[i]]) 
  ```
  
    and then we join array element with separator

src/formatDate.js
sanyokmalyshev: Thanks for advice. Change solution

liliia-livandovskaya (approved):
👍  GJ

PR #761, opened 2022-09-04
working

illia-kyselov:


src/formatDate.js
```diff
@@ -50,7 +50,37 @@
...
+        newDateObj[toFormat[i]] = `20${dateObj.YY}`;
+      } else {
+        newDateObj[toFormat[i]] = `19${dateObj.YY}`;
+      }
+    } else if (toFormat[i] === 'YY' && !dateObj[toFormat[i]]) {
+      newDateObj[toFormat[i]] = dateObj.YYYY.slice(2);
+    } else {
+      newDateObj[toFormat[i]] = dateObj[toFormat[i]];
+    }
+  }
```
SanyaBratashchuk, line 78: I think it is better to use if with break statement instead of many else if

src/formatDate.js
```diff
@@ -50,7 +50,37 @@
...
+
+  const newDate = Object.values(newDateObj);
+  const result = newDate.join(toFormat[toFormat.length - 1]);
+
+  return result;
```
SanyaBratashchuk, line 83:
```suggestion
  return newDate.join(toFormat[toFormat.length - 1]);
```

SanyaBratashchuk (changes requested):

AlisaVlasova (approved):

PR #763, opened 2022-09-05
Develop

smiyka:


src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
+
+    if (format === 'MM') {
+      dateObj.MM = dayArrayOld[i];
+    };
+
+    if (format === 'DD') {
+      dateObj.DD = dayArrayOld[i];
+    };
+
+    i++;
```
alexandra-protyanova, line 89: it would be great to use switch here

alexandra-protyanova (changes requested):

AlisaVlasova (approved):

PR #767, opened 2022-09-07
Solution

Taras-1:


src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+      if (year >= 22) {
+        dateOnNewPlaces[i] = +year + 1900;
+      } else {
+        dateOnNewPlaces[i] = +year + 2000;
+      }
+    }
+
+    if (toFormat[i] === 'YY' && year.length === 4) {
+      if (+year <= 2022 && +year >= 2000) {
+        dateOnNewPlaces[i] = +year - 2000;
```
liliia-livandovskaya, line 84:
```suggestion
        dateOnNewPlaces[i] = Number(year) - 2000;
```

liliia-livandovskaya, line 84: Let's add switch case for this whole block

liliia-livandovskaya (changes requested):
GJ 👍 Pay attention to the comments

SanyaBratashchuk (approved):

PR #768, opened 2022-09-08
Solution

MaxManis:


src/formatDate.js
```diff
@@ -49,8 +49,51 @@
...
+// - I ve tried to use minimum of magic methods
+// |-- like: x.split().map().reduce().filter().sort().join() =).
+// |-- SO here is the result:
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  // Base setup:
+  const dateArr = date.split(fromFormat[3]);
+  const separator = toFormat[3];
+  const dateObj = {};
+  const res = [];
```
Anna-Tkachenko, line 60: Just let think a little about meaningful variable names)
It's not a great idea to use name of data types in variable names (arr, obj), try to change them)

Anna-Tkachenko (changes requested):
Logic - awesome!
Need just check variables)

PR #771, opened 2022-09-08
Solution

HelenaVer:


src/formatDate.js
```diff
@@ -51,6 +51,32 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const [,,, oldSeparator] = fromFormat;
+  const [,,, newSeparator] = toFormat;
```
SanyaBratashchuk, line 55:
```suggestion
  const oldSeparator = fromFormat[3];
  const newSeparator = toFormat[3];
```

src/formatDate.js
```diff
@@ -51,6 +51,32 @@
...
+    if (dateObj['YY'] < 30) {
+      dateObj['YYYY'] = 20 + dateObj['YY'];
+    } else {
+      dateObj['YYYY'] = 19 + dateObj['YY'];
+    }
+  } else {
+    dateObj['YY'] = dateObj['YYYY'].slice(2);
+  }
+
+  for (let j = 0; j < 3; j++) {
```
SanyaBratashchuk, line 75: `3` looks like a magic number move it to the variable or use toFormat.length - 1

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #779, opened 2022-09-13
Develop

grafit-off:
feat: date formatter interpretation

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -48,9 +48,38 @@
  *
  * @returns {string}
  */
-
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const oldSeparator = fromFormat[3];
+  const newSeparator = toFormat[3];
+
+  const oldDateArr = date.split(oldSeparator);
+  const newDateArr = new Array(3);
+
+  for (let i = 0; i < newDateArr.length; i++) {
+    let oldDateItem = oldDateArr[i];
+    const oldFormatItem = fromFormat[i];
+    let newFormatIndex = toFormat.indexOf(oldFormatItem);
+
+    if (newFormatIndex < 0) {
+      newFormatIndex = oldFormatItem === 'YY' ? toFormat.indexOf('YYYY')
+        : toFormat.indexOf('YY');
+
+      oldDateItem = formatYearToAlternative(oldDateItem);
+    }
+
+    newDateArr[newFormatIndex] = oldDateItem;
+  }
+
+  return newDateArr.join(newSeparator);
+}
+
+function formatYearToAlternative(year) {
+  if (year.length !== 2) {
+    return year.slice(-2);
+  }
+
+  return year < 30 ? 20 + year : 19 + year;
 }
 
 module.exports = formatDate;


src/formatDate.js
```diff
@@ -48,9 +48,38 @@
...
  * @returns {string}
  */
-
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const oldSeparator = fromFormat[3];
+  const newSeparator = toFormat[3];
+
+  const oldDateArr = date.split(oldSeparator);
+  const newDateArr = new Array(3);
```
TheVovchik, line 57: like for decision to create array of defined length

src/formatDate.js
```diff
@@ -48,9 +48,38 @@
...
+
+  for (let i = 0; i < newDateArr.length; i++) {
+    let oldDateItem = oldDateArr[i];
+    const oldFormatItem = fromFormat[i];
+    let newFormatIndex = toFormat.indexOf(oldFormatItem);
+
+    if (newFormatIndex < 0) {
+      newFormatIndex = oldFormatItem === 'YY' ? toFormat.indexOf('YYYY')
```
TheVovchik, line 65: if you separate them on different lines 
(condition)
  ? true
  : false 
 
will be more readable

TheVovchik (approved):
beautiful resolve of task using storage array. also good job with creating a side function to transform year format.

leave few suggestions ;)

src/formatDate.js
```diff
@@ -48,9 +48,38 @@
...
+
+  return newDateArr.join(newSeparator);
+}
+
+function formatYearToAlternative(year) {
+  if (year.length !== 2) {
+    return year.slice(-2);
+  }
+
+  return year < 30 ? 20 + year : 19 + year;
```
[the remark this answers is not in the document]
grafit-off: Yes it's shorter, but affects the readability

src/formatDate.js
```diff
@@ -48,9 +48,38 @@
...
+  const newDateArr = new Array(3);
+
+  for (let i = 0; i < newDateArr.length; i++) {
+    let oldDateItem = oldDateArr[i];
+    const oldFormatItem = fromFormat[i];
+    let newFormatIndex = toFormat.indexOf(oldFormatItem);
+
+    if (newFormatIndex < 0) {
+      newFormatIndex = oldFormatItem === 'YY' ? toFormat.indexOf('YYYY')
+        : toFormat.indexOf('YY');
```
Denchepornyuk, line 66: if you move one part of the ternary operator to the new line then move another part also
```suggestion
      newFormatIndex = oldFormatItem === 'YY'
        ? toFormat.indexOf('YYYY')
        : toFormat.indexOf('YY');
```

src/formatDate.js
Denchepornyuk, line 82: usually, we use ternary operator for choosing one value depending on condidtion
but here you make math operations inside ternary operator it isn't the best way to use this operator

Denchepornyuk (approved):

src/formatDate.js
grafit-off: Got it, I'll take it into account!

src/formatDate.js
grafit-off: Thanks, will change this tomorrow!

PR #780, opened 2022-09-13
format date task finished

TheVovchik:


src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+  }
+
+  const yearFormatCheck = dataObj.hasOwnProperty('YY');
+
+  switch (true) {
+    case yearFormatCheck:
+      dataObj.YYYY = +dataObj.YY < 30
+        ? '20' + dataObj.YY
+        : dataObj.YYYY = '19' + dataObj.YY;
+      break;
```
grafit-off, line 69: Looks really short, but it affects the readability

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+
+  for (let i = 0; i < formatLength - 1; i++) {
+    convertedData += dataObj[toFormat[i]];
+
+    if (i !== formatLength - 2) {
+      convertedData += separatorTo;
+    }
+  }
+
+  return convertedData;
```
grafit-off, line 85: Is it join method?

grafit-off (approved):
Good job 🔥 Keep going

src/formatDate.js
TheVovchik: dataObj.YYYY = '19' + dataObj.YY; - I find mistake)) I don't need to write here in variable

src/formatDate.js
TheVovchik: it's join method before it was invented)) cause we need string in the end I choosed to collect data in string, I will rework it on join method. probably it will save me few lines;

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const formatLength = fromFormat.length;
+  const separatorFrom = fromFormat[formatLength - 1];
+  const splitedDate = date.split(separatorFrom);
+  const dataObj = {};
```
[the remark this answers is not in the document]
TheVovchik: will fix it

PR #782, opened 2022-09-15
Solution

jojnajbude:


myroslav-diiak (approved):

src/formatDate.js
```diff
@@ -49,8 +49,34 @@
...
+    dataObj['YY'] = `${dataObj['YYYY']}`.slice(-2);
+  } else if (toFormat.includes('YYYY')) {
+    if (!dataObj['YYYY']) {
+      if (dataObj['YY'] < 22) {
+        dataObj['YYYY'] = '20' + dataObj['YY'];
+      } else {
+        dataObj['YYYY'] = '19' + dataObj['YY'];
+      }
+    }
+  }
```
OleksiiBlezniuk, line 71: Avoid this nested else/if blocks, make it more simple

src/formatDate.js
```diff
@@ -49,8 +49,34 @@
...
+
+  for (const item of toFormat.slice(0, -1)) {
+    dataToArr.push(dataObj[item]);
+  }
+
+  const newData = dataToArr.join(toFormat[3]);
```
OleksiiBlezniuk, line 77: Save separator to variable, now it's hard to understand what is `toFormat[3]`

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+  const dataToArr = [];
+  const separator = toFormat[3];
+
+  const fromYYtoYYYY = (obj) => {
+    if (obj['YY'] < 30) {
+      obj['YYYY'] = '20' + obj['YY'];
+    } else {
+      obj['YYYY'] = '19' + obj['YY'];
+    }
+  };
```
OleksiiBlezniuk, line 64: When you creating helper function - move it outside parent functions (there is no need to create this function every time when parent function is called)

OleksiiBlezniuk (approved):

PR #785, opened 2022-09-15
Develop

Anna-Borodaienko:


imaksweb (approved):
Great job 👍

src/formatDate.js
```diff
@@ -50,7 +50,46 @@
...
+
+  if (year[1].length === 2 && year[2].length === 4) {
+    const yearShort = year[0];
+
+    if (yearShort < 30) {
+      year[0] = '20' + yearShort;
+    } else {
+      year[0] = '19' + yearShort;
+    }
+  }
```
OleksiiBlezniuk, line 79: Move this year format converting logic to separate function

OleksiiBlezniuk: Also would be nice to simplify it, it looks too complicated now. I think you can do it even without array creating

OleksiiBlezniuk (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,50 @@
...
+    && toFormat[indexToYear] === 'YY') {
+    year = year.substring(2);
+  }
+
+  if (fromFormat[indexFromYear] === 'YY'
+    && toFormat[indexToYear] === 'YYYY') {
+    year = (year < 30) ? '20' + year : '19' + year;
+  }
+
+  return year;
```
OleksiiBlezniuk, line 96:
```suggestion
  if (fromFormat[indexFromYear] === toFormat[indexToYear]) {
    return year;
  }

  if (fromFormat[indexFromYear] === 'YYYY') { // after first if fromYear and toYear are not equal
    return year.substring(2);
  }
    // fromFormat is 'YY' and toFormat is 'YYYY', 'cause of previous conditions 
    
    return (year < 30)   // wrap ternary operators conditions to new line, it's more readable
      ? '20' + year
      : '19' + year;  
```

OleksiiBlezniuk: Conditions in if looks complicated because of their length, let's make it simpler

src/formatDate.js
```diff
@@ -50,7 +50,50 @@
...
+
+  for (let i = 0; i < 3; i++) {
+    const part = toFormat[i];
+    const index = fromFormat.indexOf(part);
+
+    newDate[i] = dateArray[index];
+
+    if (!newDate[i]) {
+      newDate[i] = getYear(dateArray, fromFormat, toFormat);
+    }
```
OleksiiBlezniuk, line 64: You assigned value to array element and then change it. This solution looks strange

OleksiiBlezniuk: Let's check if this value need to be converted and assign value only once (you can use `continue` if you need it)

src/formatDate.js
```diff
@@ -50,7 +50,50 @@
...
+  }
+
+  return newDate.join(toFormat[3]);
```
OleksiiBlezniuk, line 67: save separator to variable to make it clear what have you used in this join method

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):
👍

PR #787, opened 2022-09-15
solution

myroslav-diiak:


jojnajbude (approved):

src/formatDate.js
```diff
@@ -51,6 +51,60 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const inputArray = date.split(fromFormat[fromFormat.length - 1]);
+
+  let d, m, y, separator, inputYearLength;
```
OleksiiBlezniuk, line 56: Instead of this variables (except length) you can use one object with format elements as keys

src/formatDate.js
```diff
@@ -51,6 +51,60 @@
...
+          outputArray.push(y[y.length - 2] + y[y.length - 1]);
+        } else if (toFormat[i].length > inputYearLength) {
+          if (y >= 30) {
+            outputArray.push('19' + y);
+          } else {
+            outputArray.push('20' + y);
+          }
+        } else {
+          outputArray.push(y);
+        }
```
OleksiiBlezniuk, line 98: Too complicated, let's avoid this nested if/else blocks

src/formatDate.js
```diff
@@ -51,6 +51,60 @@
...
+        outputArray.push(d);
+        break;
+
+      case 'M':
+        outputArray.push(m);
+        break;
+
+      case 'Y':
+        if (toFormat[i].length < inputYearLength) {
+          outputArray.push(y[y.length - 2] + y[y.length - 1]);
```
OleksiiBlezniuk, line 89: Use `slice`

src/formatDate.js
OleksiiBlezniuk: Also move year converting logic to separate function

src/formatDate.js
```diff
@@ -51,6 +51,60 @@
...
+            outputArray.push('20' + y);
+          }
+        } else {
+          outputArray.push(y);
+        }
+        break;
+
+      default:
+        separator = toFormat[i];
+        break;
```
OleksiiBlezniuk, line 103: What if there will be something else instead of separator?

OleksiiBlezniuk: Also you can save separator outside switch/case block, it's always a last element

src/formatDate.js
OleksiiBlezniuk: Also create more meaningful variables names instead of `d`, `m` or `y`

OleksiiBlezniuk (changes requested):

src/formatDate.js
```diff
@@ -51,6 +51,68 @@
...
+    } else if (inputLength < outputLength) {
+      if (year >= 30) {
+        return `19${year}`;
+      } else {
+        return `20${year}`;
+      }
+    }
+
+    return year;
+  }
```
OleksiiBlezniuk, line 115: Good, but better to move this function outside parent function

OleksiiBlezniuk (approved):
👍

PR #788, opened 2022-09-15
Solution

Yevheniya-Deryka:


src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateArray = date.split(fromFormat[3]);
+
+  const fromDateObject = {};
+  const toDateArray = [];
+
+  let index = 0;
+
+  for (let i = 0; i <= 2; i++) {
```
dmytristepanov, line 60: I think you can replace '2' with 'fromFormat.length - 1' , because it looks like a magic number:)

src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+        fromDateObject['YYYY'] = '19' + dateArray[index];
+      }
+    } else {
+      fromDateObject[fromFormat[i]] = dateArray[index];
+    }
+
+    index++;
+  }
+
+  for (let i = 0; i <= 2; i++) {
```
dmytristepanov, line 74: Same here

src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+    if (toFormat[i] === 'YY') {
+      toDateArray.push(fromDateObject.YYYY.slice(2));
+    } else {
+      toDateArray.push(fromDateObject[toFormat[i]]);
+    }
+  }
```
dmytristepanov, line 80: Btw very interesting solution!

dmytristepanov (approved):

src/formatDate.js
Yevheniya-Deryka: agree

src/formatDate.js
Yevheniya-Deryka: Your solution seems better to me)

src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+
+  let index = 0;
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    if (fromFormat[i] === 'YY') {
+      if (dateArray[index] < 30) {
+        fromDateObject['YYYY'] = '20' + dateArray[index];
+      } else {
+        fromDateObject['YYYY'] = '19' + dateArray[index];
+      }
```
OleksiiBlezniuk, line 66: Move year converting logic to separate function

src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    if (toFormat[i] === 'YY') {
+      toDateArray.push(fromDateObject.YYYY.slice(2));
+    } else {
+      toDateArray.push(fromDateObject[toFormat[i]]);
+    }
+  }
+
+  return toDateArray.join(toFormat[3]);
```
OleksiiBlezniuk, line 82: Better to store separator in variable, it's hard to understand here what is `toFormat[3]`

src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+  const dateArray = date.split(fromFormat[3]);
+
+  const fromDateObject = {};
+  const toDateArray = [];
+
+  let index = 0;
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    if (fromFormat[i] === 'YY') {
+      if (dateArray[index] < 30) {
```
OleksiiBlezniuk, line 62: Avoid nested `if`'s

OleksiiBlezniuk (changes requested):

src/formatDate.js
Yevheniya-Deryka: done

src/formatDate.js
Yevheniya-Deryka: done

src/formatDate.js
Yevheniya-Deryka: done

OleksiiBlezniuk (approved):
👍

PR #790, opened 2022-09-15
Solution

imaksweb:


Anna-Borodaienko (approved):
GJ!

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+      let fullYear;
+
+      if (isCurrentCentury) {
+        fullYear = dateItem.padStart(4, 20);
+      } else {
+        fullYear = dateItem.padStart(4, 19);
+      }
+
+      dateObj['YYYY'] = fullYear;
+    }
```
vmamchur, line 77: it'll be cool if you rewrite this part without using the nested if statement, but on the other hand, i don't know if it's a big mistake 🦧

vmamchur (approved):

src/formatDate.js
vmamchur: i'm late, i guess xd

imaksweb: > it'll be cool if you rewrite this part without using the nested if statement, but on the other hand, i don't know if it's a big mistake 🦧

Honestly I've no idea how to do that 😄

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparator = fromFormat.slice(-1);
+  const toSeparator = toFormat.slice(-1);
+  const formatLength = fromFormat.length;
+  const splitedDate = date.split(fromSeparator);
+  const dateObj = {};
```
Anna-Tkachenko, line 57: don't use data type names in variable names, let's create more meaningful variable name here

Anna-Tkachenko (approved):
Good job!

PR #791, opened 2022-09-15
Solution

iam-datsenko:


iPanda177 (approved):
Nice work

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
+      }
+    }
+
+    if (toFormat[i] === 'YY' && year.length === 2) {
+      dateFinal.push(year);
+    }
+
+    if (toFormat[i] === 'YY' && year.length === 4) {
+      dateFinal.push(year.slice(2));
+    }
```
OleksiiBlezniuk, line 92: Move year converting logic to separate function

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
+      year = dateSplit[i];
+    }
+
+    if (fromFormat[i] === 'MM') {
+      month = dateSplit[i];
+    }
+
+    if (fromFormat[i] === 'DD') {
+      day = dateSplit[i];
+    }
```
OleksiiBlezniuk, line 70: You can use object instead of 3 variables for this values

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
+      dateFinal.push(year.slice(2));
+    }
+
+    if (toFormat[i] === 'MM') {
+      dateFinal.push(month);
+    }
+
+    if (toFormat[i] === 'DD') {
+      dateFinal.push(day);
+    }
```
OleksiiBlezniuk, line 100: If you'll use object instead variables you'll be able to get rid of this two `if`s

OleksiiBlezniuk (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,29 @@
...
+
+  if (dateObject['YYYY']) {
+    dateObject['YY'] = dateObject['YYYY'].slice(-2);
+  }
+
+  if (dateObject['YY'] < 30) {
+    dateObject['YYYY'] = `20${dateObject['YY']}`;
+  } else {
+    dateObject['YYYY'] = `19${dateObject['YY']}`;
+  }
```
OleksiiBlezniuk, line 69: Better to move year converting to separate function

OleksiiBlezniuk (approved):

PR #794, opened 2022-09-16
Done!

iPanda177:


src/formatDate.js
```diff
@@ -50,7 +50,31 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateStr = date.split(fromFormat[3]);
```
iam-datsenko, line 53: Magic number 3?

src/formatDate.js
```diff
@@ -50,7 +50,31 @@
...
+    dateObj[fullYear] = 20 + dateObj[shortYear];
+  } else {
+    dateObj[fullYear] = 19 + dateObj[shortYear];
+  }
+
+  for (let i = 0; i < 3; i++) {
+    newDate.push(dateObj[toFormat[i]]);
+  }
+
+  return newDate.join(toFormat[3]);
```
iam-datsenko, line 77: To be continued...

src/formatDate.js
```diff
@@ -50,7 +50,31 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateStr = date.split(fromFormat[3]);
+  const fullYear = 'YYYY';
+  const shortYear = 'YY';
+  const dateObj = {};
+  const newDate = [];
+
+  for (let i = 0; i < 3; i++) {
```
iam-datsenko, line 59: Again magic?

src/formatDate.js
```diff
@@ -50,7 +50,31 @@
...
+    dateObj[shortYear] = dateObj[fullYear].slice(-2);
+  }
+
+  if (dateObj[shortYear] < 30) {
+    dateObj[fullYear] = 20 + dateObj[shortYear];
+  } else {
+    dateObj[fullYear] = 19 + dateObj[shortYear];
+  }
+
+  for (let i = 0; i < 3; i++) {
```
iam-datsenko, line 73: Still magic?

iam-datsenko (approved):
GJ! Object is a good idea

src/formatDate.js
iPanda177: Date always have the same length, so there's no need to create a variable or .length method)

src/formatDate.js
OleksiiBlezniuk: Ok, but you also can use array length here instead of 3

src/formatDate.js
OleksiiBlezniuk, line 77: Save separator in variable, now it's not clear what is `toFormat[3]`

OleksiiBlezniuk (approved):

PR #795, opened 2022-09-16
first quick solution

fatihsamur:


kamal-tufekcic (approved):

ihor-jpeg (approved):
It's an optional task, so I approve. But try to use just one for loop

PR #798, opened 2022-09-16
js_format_date_advanced

xiimiik:


Dikey945:
Good Job! Keep going 🚀

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+    newDate.push(dateObj[toFormat[i]]);
+
+    if (!dateObj.hasOwnProperty(toFormat[i])) {
+      if (toFormat[i].length === 2) {
+        year = year.slice(2);
+      }
+
+      if (toFormat[i].length === 4) {
+        year = yearOfCentury + year;
+      }
```
OleksiiBlezniuk, line 80: Move this logic to separated function

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+
+    dateObj[fromFormat[i]] = oldDate[i];
+  }
+
+  const yearOfCentury = (year < 30) ? 20 : 19;
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    newDate.push(dateObj[toFormat[i]]);
+
+    if (!dateObj.hasOwnProperty(toFormat[i])) {
```
OleksiiBlezniuk, line 73: It's strange that you have to overwrite element in array after adding, you can wrap `push` in `if` and break loop after element adding. Then there will be no need to wrap other logic of loop in `if` and you'll be able to push modified year to array

OleksiiBlezniuk (changes requested):

src/formatDate.js
```diff
@@ -49,8 +49,42 @@
  * @returns {string}
  */
 
+function formatedYear(year) {
```
OleksiiBlezniuk, line 52:
```suggestion
function formatYear(year) {
```

OleksiiBlezniuk: format - action, formatted - result of this action

OleksiiBlezniuk (approved):

PR #799, opened 2022-09-16
Solution of format date advanced

Dikey945:


xiimiik (approved):
Great, nothing to add 🔥

src/formatDate.js
```diff
@@ -51,6 +51,49 @@
...
+  const oldDateFormat = fromFormat.slice(0, -1);
+  const newDateFormat = toFormat.slice(0, -1);
+  const oldDateObject = { normalize: toNormalizeDate };
+  const newDate = [];
+
+  for (let i = 0; i < oldDate.length; i++) {
+    oldDateObject[oldDateFormat[i]] = oldDate[i];
+  }
+
+  oldDateObject.normalize(newDateFormat);
```
Anna-Tkachenko, line 66: what's the reason of adding this function as a method for new object? let's create universal function, that will accept 2 objects (old date and new date), and make needed logic

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #800, opened 2022-09-16
Solution

anastasiia-nahaiska:


artem-musii (approved):
GJ!

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dayMonthYear = [];
+  const dateArr = date.split(fromFormat[3]);
```
Anna-Tkachenko, line 54: don't use data type names in variable names

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+    && dayMonthYear[2] < 30) {
+      newDate[i] = `20${dayMonthYear[2]}`;
+    } else if (toFormat[i] === 'YYYY' && dayMonthYear[2].length === 2
+    && dayMonthYear[2] >= 30) {
+      newDate[i] = `19${dayMonthYear[2]}`;
+    } else if (toFormat[i] === 'YYYY' && dayMonthYear[2].length === 4) {
+      newDate[i] = dayMonthYear[2];
+    } else {
+      newDate[i] = dayMonthYear[2].slice(2);
+    }
```
Anna-Tkachenko, line 82: looks too complicated
can you split this code to some separate function for defining needed formats? and use them to define needed values for result data

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #801, opened 2022-09-16
solution

artem-musii:


anastasiia-nahaiska (approved):
Good solution!!!

src/formatDate.js
```diff
@@ -50,7 +50,33 @@
...
+
+  if (receivedData.hasOwnProperty('YYYY')) {
+    receivedData.YY = receivedData.YYYY.slice(-2);
+  } else {
+    const century = receivedData.YY < 30
+      ? 20
+      : 19;
+
+    receivedData.YYYY = `${century}${receivedData.YY}`;
+  }
```
Anna-Tkachenko, line 69: it can be separate function for defining needed year

Anna-Tkachenko (approved):

PR #802, opened 2022-09-16
solution to format date advanced

dsydorenko534:


src/formatDate.js
```diff
@@ -50,7 +50,42 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const separator = fromFormat[3];
+  const newJointer = toFormat[3];
+  const result = [];
```
iPanda177, line 55:
```suggestion
  const newDate = [];
```

src/formatDate.js
```diff
@@ -50,7 +50,42 @@
...
+  const splitDate = date.split(separator);
+
+  const obj = {};
+
+  for (let i = 0; i < splitDate.length; i++) {
+    obj[fromFormat[i]] = splitDate[i];
+  }
+
+  if (fromFormat.includes('YY') && toFormat.includes('YYYY')) {
+    const tmp = obj.YY;
```
iPanda177, line 66: what is 'tmp'? make more understandable variable.
```suggestion
    const temporary = obj.YY;
```

iPanda177 (changes requested):

iPanda177 (approved):

src/formatDate.js
```diff
@@ -50,7 +50,42 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const separator = fromFormat[3];
+  const newJointer = toFormat[3];
+  const newDate = [];
+
+  const splitDate = date.split(separator);
+
+  const obj = {};
```
silchyn, line 59: Same here

silchyn (changes requested):

silchyn (changes requested):
"object" and "temporary" are by no means better than "obj" and "tmp"

silchyn (approved):
Cool!

PR #805, opened 2022-09-17
solution

artem-nasadchuk:


src/formatDate.js
```diff
@@ -50,7 +50,29 @@
...
+
+  if (objDate.hasOwnProperty('YY')) {
+    objDate.YYYY = objDate.YY < 30
+      ? '20' + objDate.YY
+      : '19' + objDate.YY;
+  }
+
+  if (objDate.hasOwnProperty('YYYY')) {
+    objDate.YY = objDate.YYYY.slice(2);
+  }
```
MaxManis, line 69: i belive you could use `if \ else if` statement here)
```suggestion
  if (objDate.hasOwnProperty('YY')) {
    objDate.YYYY = objDate.YY < 30
      ? '20' + objDate.YY
      : '19' + objDate.YY;
  } else if (objDate.hasOwnProperty('YYYY')) {
    objDate.YY = objDate.YYYY.slice(2);
  }
```

src/formatDate.js
```diff
@@ -50,7 +50,29 @@
...
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    resultDate.push(objDate[toFormat[i]]);
+  }
+
+  return resultDate.join(`${toFormat[toFormat.length - 1]}`);
```
MaxManis, line 75: new separator is already a string, so you dont need to use formated str `${x}` )
```suggestion
  return resultDate.join(toFormat[toFormat.length - 1]);
```

MaxManis (commented):
GJ!)
left a few comments)

MaxManis (approved):
Good job!)

Anna-Tkachenko (approved):
Good job!

PR #807, opened 2022-09-18
Solution

Yeliseiev:


src/formatDate.js
```diff
@@ -50,7 +50,29 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldDateFormat = date.split(fromFormat[3]);
```
glugovsky, line 53: It is not clear, what `fromFormat[3]` is referring to here and `toFormat[3]` on the 75th line. Probably, would be better to add extra variables for separators

glugovsky (approved):

src/formatDate.js
Yeliseiev: Yep, you rigth. Better name them)

silchyn (approved):
Cool!

PR #808, opened 2022-09-18
Solution

Oleksandra-Shevchenko:


src/formatDate.js
```diff
@@ -51,6 +51,53 @@
...
+    if (toFormat[i] === 'YYYY') {
+      newDate[i] = newYear;
+    }
+
+    if (toFormat[i] === 'YY') {
+      newDate[i] = newYear.slice(-2);
+    }
+  }
+
+  return newDate.join(toFormat[3]);
```
roman-zherebetskyi, line 100: again number 3, you can make it like a variable

src/formatDate.js
```diff
@@ -51,6 +51,53 @@
...
+    if (fromFormat[i] === 'YY' && +splitDate[i] < 30) {
+      newYear = '20' + splitDate[i];
+    }
+
+    if (fromFormat[i] === 'YY' && +splitDate[i] >= 30) {
+      newYear = '19' + splitDate[i];
+    }
+  }
+
+  for (let i = 0; i < 3; i++) {
```
roman-zherebetskyi, line 82: here again, I would rather use more more understandable parameter

src/formatDate.js
```diff
@@ -51,6 +51,53 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const newDate = [];
+  const splitDate = date.split(fromFormat[3]);
+  let newDay = '';
+  let newMonth = '';
+  let newYear = '';
+
+  for (let i = 0; i < 3; i++) {
```
roman-zherebetskyi, line 60: It's okay but, what is 3 here? I mean why you use exactly 3? you can use splitDate.length for example

roman-zherebetskyi (approved):
Nice job!!!

silchyn (changes requested):
Take into account these suggestions

silchyn (approved):
Cool!

PR #809, opened 2022-09-18
Solution

Apollonia5:


Arsen-Shpak (approved):

silchyn (changes requested):
You can do it simpler, without nested loops

silchyn (approved):
Cool!

PR #811, opened 2022-09-18
Function formatDate was created

mykyta-krainik:


src/formatDate.js
```diff
@@ -51,6 +51,36 @@
...
+  const newDate = [];
+  const newSeparator = toFormat[3];
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    let posInNewFormat = toFormat.indexOf(fromFormat[i]);
+
+    if (posInNewFormat !== -1) {
+      newDate[posInNewFormat] = parsedDate[i];
+
+      continue;
```
liliia-livandovskaya, line 64: do you really need `continue` there?)

src/formatDate.js
```diff
@@ -51,6 +51,36 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
```
liliia-livandovskaya, line 53: remove comments

src/formatDate.js
```diff
@@ -51,6 +51,36 @@
...
+  const parsedDate = date.split(fromFormat[3]);
+  const newDate = [];
+  const newSeparator = toFormat[3];
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    let posInNewFormat = toFormat.indexOf(fromFormat[i]);
```
liliia-livandovskaya, line 59: posInNewFormat - make the name understandable

liliia-livandovskaya (changes requested):
GJ! Pay attention to the comments

src/formatDate.js
mykyta-krainik: Yes, I do, because in that if condition I treat with the situation when indexOf has found the needed element in the toFormat array and I don't want the algorithm to deal with the part of the code where the exclusion situation is handled

liliia-livandovskaya (approved):
GJ 👍

PR #812, opened 2022-09-18
solution

Naveed-Shoukat:


Bibicutu (approved):
Good job!

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const splitedDate = date.split(fromFormat[3]);
+
+  const dateIndexFrom = fromFormat.indexOf('DD');
```
fatihsamur, line 55: I guess this variable should named as `dayIndexFrom` and `dayIndexTo`

fatihsamur (approved):
Well done. Approved 👍

src/formatDate.js
Naveed-Shoukat: In my opinion day is like weekday and date is like month-date. That is why I chose date

Naveed-Shoukat: I have revised my solution using Object, please re-review

fatihsamur (approved):
I think I will shorten my solution like this 👍

ihor-jpeg (approved):

PR #833, opened 2022-09-26
function formatDate was created

diakovMax:


src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+      case 'MM':
+        dateOb.month = arrDateFrom[i];
+        break;
+      case 'DD':
+        dateOb.day = arrDateFrom[i];
+    }
+  }
+
+  for (let i = 0; i < 3; i++) {
+    if ((toFormat[i] === 'YYYY') || (toFormat[i] === 'YY')) {
```
TarasHoliuk, line 87: `swich` would look better here too

TarasHoliuk (changes requested):
Well done! left 2 comments, check it, please

liliia-livandovskaya (approved):
🔥

PR #840, opened 2022-10-04
formatDate was implemented

Andrii-Medintsev:


src/formatDate.js
```diff
@@ -50,7 +50,37 @@
...
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    const index = toFormat.indexOf(fromFormat[i]);
+
+    newArray[index] = temporaryArray[i];
+  }
+
+  return newArray.join(toFormat[toFormat.length - 1]);
+}
+
+function convertYear(str) {
```
liliia-livandovskaya, line 74:
```suggestion
function convertYear(year) {
```
let's use appropriate names for variables

liliia-livandovskaya (changes requested):
Pay attention to the comments

PR #845, opened 2022-10-07
solved

Deniso4ka23:


qoa11a (changes requested):
Your code returns incorrect value in this case `formatDate(
  '97/02/18',
  ['YY', 'MM', 'DD', '/'],
  ['DD', 'MM', 'YYYY', '.'],
)`

It returns 18.02.97 and should return 18.02.1997

Deniso4ka23:
I tried to code differently. I hope in this task it was possible to use array iter meth.

qoa11a (approved):

PR #852, opened 2022-10-14
Solution

bokorzita:


src/formatDate.js
```diff
@@ -50,7 +50,51 @@
...
+    }
+
+    if (fromFormat[i].charAt(fromFormat[i].length - 1) === 'M') {
+      dateObj.month = dateArr[i];
+    }
+
+    if (fromFormat[i].charAt(fromFormat[i].length - 1) === 'D') {
+      dateObj.day = dateArr[i];
+    }
+  }
```
amitpatiljc, line 68: instead of switch case for optimisation you can use `switch` case here instead of if statements

amitpatiljc (approved):
Good Job!

PR #854, opened 2022-10-16
Solution

Vasyliy22:


src/formatDate.js
```diff
@@ -50,7 +50,37 @@
...
+        if (toFormat[i] !== shortYear) {
+          result.push(year < 30 ? `20${year}` : `19${year}`);
+        } else {
+          result.push(year.slice(2));
+        }
+      } else {
+        result.push(year);
+      }
+    } else {
+      result.push(dateObj[toFormat[i]]);
```
polosanya, line 79: Try to avoid nesting if...else. It can be difficult to understand such code.

polosanya (changes requested):
Good job! But I think you can make your code a lit bit cleaner. Please check my comments

src/formatDate.js
```diff
@@ -50,7 +50,35 @@
...
+        ? dateObj[shortYear]
+        : dateObj['YYYY'];
+
+      if (!dateObj.hasOwnProperty(toFormat[i])) {
+        result.push(year < 30
+          ? `20${year}`
+          : `${year.slice(2)}` || `19${year}`);
+      } else {
+        result.push(year);
+      }
```
loralevitska, line 75: one more solution to avoid this `if else`

```suggestion
     const formattedYear = year < 30
       ? `20${year}`
       : `${year.slice(2)}` || `19${year}`;

     const needYear = !dateObj.hasOwnProperty(toFormat[i]) ? formattedYear : year

     result.push(needYear);
```

loralevitska (approved):
great!

PR #857, opened 2022-10-20
Format date

maksym-kryvolap:


src/formatDate.js
```diff
@@ -50,7 +50,30 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const array = date.split(fromFormat[3]);
```
TarasHoliuk, line 53: `array` is not a good name. Rename it to something that explains what exactly it is.

src/formatDate.js
```diff
@@ -50,7 +50,30 @@
...
+  const newDate = [];
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    const lengthTrue = fromFormat[i].length !== toFormat[i].length;
```
TarasHoliuk, line 57: rename it, please
at least `true` in the name when `!==` in comparison sounds weird)

src/formatDate.js
```diff
@@ -50,7 +50,30 @@
...
+    const inc = fromFormat[i].includes('Y') && toFormat[i].includes('Y');
```
TarasHoliuk, line 58: the same, do not use short forms, add a name which will better describe value

TarasHoliuk (changes requested):
GJ! But pay attention to naming

TarasHoliuk (approved):

PR #860, opened 2022-10-23
Solution

Illia-Skladnik:


src/formatDate.js
```diff
@@ -50,7 +50,58 @@
...
+
+  for (let i = 0; i < 3; i++) {
+    if (fromFormat.includes(toFormat[i])) {
+      index = fromFormat.indexOf(toFormat[i]);
+    } else if (fromFormat.includes('YYYY')) {
+      index = fromFormat.indexOf('YYYY');
+    } else {
+      index = fromFormat.indexOf('YY');
+    }
+
```
Svitlana-Yudina, line 65: I'm confused what is this index for... It might be better to name the variable `indexFrom` or `indexTo`. but it's up to you))

src/formatDate.js
```diff
@@ -50,7 +50,58 @@
...
+      if (fromFormat[index] === 'YY') {
+        if (dateArr[index] < 30) {
+          newDate += '20';
+          newDate += dateArr[index];
+        } else {
+          newDate += '19';
+          newDate += dateArr[index];
+        }
+      }
```
Svitlana-Yudina, line 94: Perhaps you can move  `newDate += dateArr[index];`  outside the nested if, so that the same action does not repeat)
```suggestion
      if (fromFormat[index] === 'YY') {
        if (dateArr[index] < 30) {
          newDate += '20';
        } else {
          newDate += '19';
        }
      newDate += dateArr[index];
      }
```

Svitlana-Yudina (approved):
GJ! 🔥 It's a very interesting solution)) 👍

iryna-lelyk (approved):

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let newDate = '';
+  let index = 0;
+  const dateArr = date.split(fromFormat[3]);
+
+  for (let i = 0; i < 3; i++) {
+    if (fromFormat.includes(toFormat[i])) {
+      index = fromFormat.indexOf(toFormat[i]);
```
illiarerun, line 59: This means that format doesn't change, better add your value right here and use `continue` to move to the next iteration

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
+      index = fromFormat.indexOf('YYYY');
+    } else {
+      index = fromFormat.indexOf('YY');
+    }
+
+    if (toFormat[i] === 'YY') {
+      switch (fromFormat[index]) {
+        case 'YY':
+          newDate += fromFormat[index];
+          break;
```
illiarerun, line 70: then you wouldn't need this

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
+
+        default:
+          break;
+      }
+    }
+
+    if (toFormat[i] === 'YYYY') {
+      if (fromFormat[index] === 'YYYY') {
+        newDate += dateArr[index];
+      }
```
illiarerun, line 84: and this

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
+
+    if (toFormat[i] === 'YY') {
+      switch (fromFormat[index]) {
+        case 'YY':
+          newDate += fromFormat[index];
+          break;
+
+        case 'YYYY':
+          newDate += dateArr[index].slice(2);
+          break;
```
illiarerun, line 74: and you just can check it in your if to o

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
+      if (fromFormat[index] === 'YYYY') {
+        newDate += dateArr[index];
+      }
+
+      if (fromFormat[index] === 'YY') {
+        if (dateArr[index] < 30) {
+          newDate += '20';
+        } else {
+          newDate += '19';
+        }
```
illiarerun, line 91: Better use ternary operator here

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
+        newDate += dateArr[index];
+      }
+    }
+
+    if (toFormat[i] === 'DD' || toFormat[i] === 'MM') {
+      newDate += dateArr[index];
+    }
+
+    newDate += toFormat[3];
```
illiarerun, line 100: just don't add it if it's the last element of your array, then you wouldn't need `slice`

illiarerun (changes requested):

mykola-koval (approved):

PR #861, opened 2022-10-24
solution

MaxFedak:


Kry-Gin (approved):
GJ! Really interesting solution

src/formatDate.js
```diff
@@ -51,6 +51,50 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const oldSeparator = fromFormat[fromFormat.length - 1];
+  const newSeparator = toFormat[toFormat.length - 1];
+  const dateArray = date.split(oldSeparator);
```
denys-danyliuk, line 56:
```suggestion
  const splittedDate = date.split(oldSeparator);
```

src/formatDate.js
```diff
@@ -51,6 +51,50 @@
...
+  const dateFormattedArray = [];
+
+  const dateObj = {};
+
+  dateArray.forEach((item, i) => {
+    dateObj[fromFormat[i][0]] = item;
```
denys-danyliuk, line 62:
```suggestion
    dateObj[fromFormat[i]] = item;
```
it's pretty hard to read/understand if you do it [i][0], you can just use [i], no need to get only first digit

src/formatDate.js
```diff
@@ -51,6 +51,50 @@
...
+  });
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    switch (toFormat[i][0]) {
```
denys-danyliuk, line 66: same here, it's just unobvious why you need to to this

src/formatDate.js
```diff
@@ -51,6 +51,50 @@
...
+            }
+          }
+
+          if (dateObj.Y.length === 4) {
+            yearFormatted = yearFormatted.slice(2);
+          }
+          dateFormattedArray.push(yearFormatted);
+        } else {
+          dateFormattedArray.push(dateObj.Y);
+        }
```
denys-danyliuk, line 93: add default case for you 'switch', in default case you could add logic for situation when you don't need to change format of the year

src/formatDate.js
```diff
@@ -51,6 +51,50 @@
...
+        if (dateObj.Y.length !== toFormat[i].length) {
+          let yearFormatted = dateObj.Y;
+
+          if (dateObj.Y.length === 2) {
+            if (+yearFormatted < 30) {
+              yearFormatted = '20' + ('' + yearFormatted);
+            } else {
+              yearFormatted = '19' + ('' + yearFormatted);
+            }
+          }
```
denys-danyliuk, line 85: 😵‍💫
avoid nested if statements, you can manage this task without them

src/formatDate.js
```diff
@@ -51,6 +51,50 @@
...
+
+          if (dateObj.Y.length === 4) {
+            yearFormatted = yearFormatted.slice(2);
```
denys-danyliuk, line 88: just push it here immediately

denys-danyliuk (changes requested):
you have three nested if statements, try to find better solution, but overall good job!🚀

src/formatDate.js
MaxFedak: if i get rid of [0] it will break all down

it's not for digits, but for letters. I've got object keys 'm', 'd', 'y' to not be concerned about which format of year (YY or YYYY) we've got. So,  i need only a first letter of the format for this solution

i can change that, but i'll need to implement new logic for the cycle, as i see..

denys-danyliuk: I see, then you can leave it just as it is

src/formatDate.js
```diff
@@ -50,7 +50,44 @@
...
+          if (dateObj.Y.length === 2) {
+            yearFormatted = +yearFormatted < 30
+              ? '20' + ('' + yearFormatted)
+              : '19' + ('' + yearFormatted);
+          } else {
+            yearFormatted = yearFormatted.slice(2);
+          }
+        }
+
+        dateFormattedArray.push(yearFormatted);
```
denys-danyliuk, line 86: do this in else statement

src/formatDate.js
```diff
@@ -50,7 +50,44 @@
...
+      case 'Y':
+        let yearFormatted = dateObj.Y;
+
+        if (dateObj.Y.length !== toFormat[i].length) {
+          if (dateObj.Y.length === 2) {
+            yearFormatted = +yearFormatted < 30
+              ? '20' + ('' + yearFormatted)
+              : '19' + ('' + yearFormatted);
+          } else {
+            yearFormatted = yearFormatted.slice(2);
```
denys-danyliuk, line 82:
```suggestion
      dateFormattedArray.push(yearFormatted.slice(2));
```

denys-danyliuk (changes requested):
add default case for your switch statement, also pay attention to the comment

denys-danyliuk (approved):

PR #862, opened 2022-10-24
Solution

Kry-Gin:


src/formatDate.js
```diff
@@ -50,7 +50,48 @@
...
+    } else if (desired === 'YYYY' && !fromFormat.includes(desired)) {
+      desired = 'YY';
+    }
+
+    yearIndex = fromFormat.indexOf(desired);
+
+    newDateArray[i] = dateArray[yearIndex];
+  }
+
+  if (fromFormat[3] !== toFormat[3]) {
```
MaxFedak, line 88: do we really need check here?
i guess, if they equal or different u can just always use new separator

MaxFedak (approved):

src/formatDate.js
```diff
@@ -50,7 +50,44 @@
...
+    } else if (desired === 'YYYY' && !fromFormat.includes(desired)) {
+      desired = 'YY';
+    }
+
+    yearIndex = fromFormat.indexOf(desired);
+
+    newDateArray[i] = dateArray[yearIndex];
+  }
+
+  newDate = newDateArray.join(toFormat[3]);
```
denys-danyliuk, line 88:
```suggestion
 return newDateArray.join(toFormat[3]);
```

src/formatDate.js
```diff
@@ -50,7 +50,44 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let newDate = date;
```
denys-danyliuk, line 53: this variable is redundant, you almost don't use it

src/formatDate.js
```diff
@@ -50,7 +50,44 @@
...
+    if (year < 30) {
+      dateArray[yearIndex] = `20${year}`;
+    } else {
+      dateArray[yearIndex] = `19${year}`;
+    }
+  } else if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
+    yearIndex = fromFormat.indexOf('YYYY');
+
+    dateArray[yearIndex] = dateArray[yearIndex].slice(-2);
+  }
```
denys-danyliuk, line 72: avoid nested if statements

src/formatDate.js
```diff
@@ -50,7 +50,44 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let newDate = date;
+  const dateArray = newDate.split(fromFormat[3]);
+  const newDateArray = [];
+  let yearIndex = 0;
```
denys-danyliuk, line 56: this variable also seems redundant, it only adds unnecessary code where you use it

denys-danyliuk (changes requested):
I like your solution, just think if you can avoid making nested if statements, but overall good job🚀

denys-danyliuk (approved):

PR #863, opened 2022-10-25
format-date solution

Svitlana-Yudina:


Illia-Skladnik (approved):
Very neat and readable code!

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const curentDateArr = date.split(fromFormat[3]);
+  const newFormatArr = Array(3);
+
+  const fromIndexD = fromFormat.indexOf('DD');
+  const fromIndexM = fromFormat.indexOf('MM');
+  const fromIndexY = 3 - fromIndexD - fromIndexM;
```
illiarerun, line 58: 🔥

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+    if (toFormat[i].includes('D')) {
+      newFormatArr[i] = curentDateArr[fromIndexD];
+    }
+
+    if (toFormat[i].includes('M')) {
+      newFormatArr[i] = curentDateArr[fromIndexM];
+    }
+
+    if (toFormat[i].includes('Y')) {
+      newFormatArr[i] = curentDateArr[fromIndexY];
```
illiarerun, line 70: These are the same actions just based on different values and you repeat yourself three times. Try to create object like:
```
{
  D: value,
  M: value,
  Y: value,
}
```
so you can get the first character in `toFormat[i]` and find needed element by this character from object by default

illiarerun (approved):
Approve, but please pay attention to comment and try to improve your solution a bit

PR #864, opened 2022-10-25
Solution

Vlad-Fedorishchev:


src/formatDate.js
```diff
@@ -50,7 +50,62 @@
...
+    if (fromFormat[i].includes('M')) {
+      monthOfDate = splittedDate[i];
+    }
+
+    if (fromFormat[i].includes('D')) {
+      dayOfDate = splittedDate[i];
+    }
+  }
+
+  for (let i = 0; i < 3; i++) {
```
polosanya, line 79: Why do you have the same loop twice?
Try to avoid it

src/formatDate.js
```diff
@@ -50,7 +50,62 @@
...
+      if (toFormat[i].length === 2) {
+        result[i] = yearOfDate.slice(2);
+      }
+
+      if (toFormat[i].length === 4) {
+        if (+yearOfDate < 30) {
+          result[i] = '20' + yearOfDate;
+        } else {
+          result[i] = '19' + yearOfDate;
+        }
```
polosanya, line 95: You can use ternary operator instead

polosanya (changes requested):
Good job! But please check my comments

Vlad-Fedorishchev:
@polosanya Already fixed mistakes according to your comments, can u check please?)

qoa11a (approved):

PR #865, opened 2022-10-25
Format_date was created

iryna-lelyk:


taraselnik (approved):
Like it! i have no to add because of 'if code works don't touch it' :) interesting solution.

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
+  const [, , , newOperator] = toFormat;
+
+  const oldDate = date.split(oldOperator);
+  const oldDayPlace = fromFormat.indexOf('DD');
+  const oldMonthPlace = fromFormat.indexOf('MM');
+  let oldYearPlace = fromFormat.indexOf('YY');
+
+  if (oldYearPlace === -1) {
+    oldYearPlace = fromFormat.indexOf('YYYY');
+  }
```
denys-danyliuk, line 63: use ternary operator or || to get value of this variable

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
+
+  const newDayPlace = toFormat.indexOf('DD');
+  const newMonthPlace = toFormat.indexOf('MM');
+  let newYearPlace = toFormat.indexOf('YY');
+
+  if (newYearPlace === -1) {
+    newYearPlace = toFormat.indexOf('YYYY');
+  }
```
denys-danyliuk, line 71: same here

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
+
+  const newDate = [];
+
+  newDate[newDayPlace] = oldDate[oldDayPlace];
+  newDate[newMonthPlace] = oldDate[oldMonthPlace];
+
+  switch (true) {
```
denys-danyliuk, line 78: don't use true as condition for switch

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
+      newDate[newYearPlace] = oldDate[oldYearPlace];
+      break;
+
+    case fromFormat[oldYearPlace] === 'YY' && oldDate[oldYearPlace] < 30:
+      newDate[newYearPlace] = 20 + oldDate[oldYearPlace];
+      break;
+
+    case fromFormat[oldYearPlace] === 'YY' && oldDate[oldYearPlace] >= 30:
+      newDate[newYearPlace] = 19 + oldDate[oldYearPlace];
+      break;
```
denys-danyliuk, line 89: combine this two cases in to one

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
+
+    case fromFormat[oldYearPlace] === 'YYYY':
+      newDate[newYearPlace] = oldDate[oldYearPlace].slice(2);
+      break;
```
denys-danyliuk, line 93: you can do this as your default case

denys-danyliuk (changes requested):
great solution🚀

denys-danyliuk (approved):

PR #866, opened 2022-10-26
solution

taraselnik:


src/formatDate.js
```diff
@@ -50,7 +50,41 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const separator = fromFormat[fromFormat.length - 1];
+  const splittedDate = date.split(separator);
+  const lengthSplitedDate = splittedDate.length;
```
iryna-lelyk, line 55: Maybe name your variables in other way. Something like dateArray or something.

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+
+    if (keyFromFormat === 'YY') {
+      if (+valueFromSplitteDate < 30) {
+        storeDateObj['YYYY'] = '20' + valueFromSplitteDate;
+      } else {
+        storeDateObj['YYYY'] = '19' + valueFromSplitteDate;
+      }
+    }
+
+    storeDateObj[keyFromFormat] = valueFromSplitteDate;
```
iryna-lelyk, line 74: Maybe you should check if this is a month or a day, but not a year, because you already assigned values for year (your code line 62-72).

iryna-lelyk (approved):
Very interesting solution!

src/formatDate.js
taraselnik: no need i think,  because we have sure keys (formats) in input datas (they are in the conditions of task. in case Datas will bad we will never understand what is year and what is month and what is day).

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+  const separator = fromFormat[fromFormat.length - 1];
+  const splittedDate = date.split(separator);
+  const lengthSplitedDate = splittedDate.length;
+  const storeDateObj = {};
+
+  for (let i = 0; i < lengthSplitedDate; i++) {
+    const keyFromFormat = fromFormat[i];
+    const valueFromSplitteDate = splittedDate[i];
+
+    if (keyFromFormat === 'YYYY') {
```
illiarerun, line 62: add check that it's not present in `toFormat` because you do redundant action otherwise

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+      storeDateObj['YY'] = valueFromSplitteDate.slice(2);
+    }
+
+    if (keyFromFormat === 'YY') {
```
illiarerun, line 66: same here

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+    }
+
+    storeDateObj[keyFromFormat] = valueFromSplitteDate;
+  }
+
+  const separatorOut = toFormat[toFormat.length - 1];
+  const resultDateArr = [];
+  const toFormatDate = toFormat.slice(0, -1);
+
+  for (const i of toFormatDate) {
```
illiarerun, line 81: that's not index, but a datePart
```suggestion
  for (const datePart of toFormatDate) {
```

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+    if (keyFromFormat === 'YYYY') {
+      storeDateObj['YY'] = valueFromSplitteDate.slice(2);
+    }
+
+    if (keyFromFormat === 'YY') {
+      if (+valueFromSplitteDate < 30) {
+        storeDateObj['YYYY'] = '20' + valueFromSplitteDate;
+      } else {
+        storeDateObj['YYYY'] = '19' + valueFromSplitteDate;
+      }
```
illiarerun, line 71: basically you do almost the same operations here, try to use ternary operator or create a variable for `20` or `19` and then add it to your obj, follow DRY

illiarerun (changes requested):

src/formatDate.js
taraselnik: ok. fix redundance with condition

src/formatDate.js
taraselnik: fixed

src/formatDate.js
taraselnik: okay. fixed with ternary

mykola-koval (approved):

PR #870, opened 2022-10-26
Develop

AlexTsimba:


src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
+
+  const newSeparator = toFormat.slice(-1);
+  const toFormatOrder = toFormat.slice(0, 3);
+  const formattedDate = [];
+
+  const YEAR_BREAKPOINT = 30;
+  const THIS_CENTURY = 20;
+  const LAST_CENTURY = 19;
+  const SHORT_DATE = 2;
+  const LONG_DATE = 4;
```
andrii-bassarab, line 64: Why are you declaring these variables in upper case?

andrii-bassarab:
GJ!

src/formatDate.js
AlexTsimba: I  just left inside only variables that cannot be taken out. I'm not 100% sure where is the best place to declare these 4, but I hope, next reviewer will describe the best options

mykola-koval (approved):

PR #872, opened 2022-10-26
Solution

andrii-bassarab:


src/formatDate.js
```diff
@@ -50,7 +50,31 @@
...
+  || inputDate[fromFormat.indexOf('YY')];
+
+  const copyToFormat = [...toFormat];
+
+  copyToFormat[copyToFormat.indexOf('DD')] = number;
+  copyToFormat[copyToFormat.indexOf('MM')] = month;
+
+  copyToFormat[copyToFormat.indexOf('YYYY')] = (year.length === 4)
+    ? year
+    : (year < 23)
```
AlexTsimba, line 66: according to the conditions, breakout year should be 30

src/formatDate.js
```diff
@@ -50,7 +50,31 @@
...
+      ? '20' + year
+      : '19' + year;
```
AlexTsimba, line 68: let's move all static numbers to variables

AlexTsimba (commented):
Great job, but with few small things to improve

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const inputDate = date.split(fromFormat[3]);
+  const number = inputDate[fromFormat.indexOf('DD')];
```
denys-danyliuk, line 54: why 'number'? I thought it should be 'day'

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
+  const month = inputDate[fromFormat.indexOf('MM')];
+  const year = inputDate[fromFormat.indexOf('YYYY')]
+  || inputDate[fromFormat.indexOf('YY')];
```
denys-danyliuk, line 57: good knowledge of '||' operator🔥

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
+
+  const copyToFormat = [...toFormat];
+  const dateBreakPoint = 30;
+  const currentCentury = 20;
+  const lastCentury = 19;
```
denys-danyliuk, line 62: I mean, this is unnecessary cause you used them only once and probably they won't change, but okay

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
+    : (year < dateBreakPoint)
+      ? currentCentury + year
+      : lastCentury + year;
+
+  copyToFormat[copyToFormat.indexOf('YY')] = (year.length === 2)
+    ? year
+    : year.slice(2);
+
+  const resultData = copyToFormat[0] + copyToFormat[3] + copyToFormat[1]
+    + copyToFormat[3] + copyToFormat[2];
```
denys-danyliuk, line 78: 😵‍💫
it's hard to understand what you are doing here, would be better if you will do this with 'copyToFormat' and in the end just 'join' him

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
+  const lastCentury = 19;
+
+  copyToFormat[copyToFormat.indexOf('DD')] = number;
+  copyToFormat[copyToFormat.indexOf('MM')] = month;
+
+  copyToFormat[copyToFormat.indexOf('YYYY')] = (year.length === 4)
+    ? year
+    : (year < dateBreakPoint)
+      ? currentCentury + year
+      : lastCentury + year;
```
denys-danyliuk, line 71: try to avoid nested ternary/if statements

denys-danyliuk (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
+
+  copyToFormat[copyToFormat.indexOf('YYYY')] = (year.length === 4)
+    ? year
+    : century + year;
+
+  copyToFormat[copyToFormat.indexOf('YY')] = (year.length === 2)
+    ? year
+    : year.slice(2);
+
+  const resultData = copyToFormat.slice(0, 3).join(copyToFormat[3]);
```
denys-danyliuk, line 76:
```suggestion
  return copyToFormat.slice(0, 3).join(copyToFormat[3]);
```

denys-danyliuk (approved):

PR #881, opened 2022-10-31
solution

Anastasiia-Smirnova:


taraselnik (approved):
interesting solution. a bit hard to read (( enjoy

src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
+  const newDate = new Array(toFormat.length - 1);
+  let divider = '';
+  let newDivider = '';
+
+  for (const char of date) {
+    if (isNaN(char)) {
+      divider = char;
+      break;
+    }
+  }
```
denys-danyliuk, line 62: it's redundant

src/formatDate.js
```diff
@@ -50,7 +50,52 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newDate = new Array(toFormat.length - 1);
+  let divider = '';
```
denys-danyliuk, line 54:
```suggestion
  let divider = fromFormat[fromFormat.length - 1];
```
or something like that instead of creating cycle

src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
+  for (let i = 0; i < 3; i++) {
+    const key = toFormat[i][0];
+
+    if (toFormat[i].startsWith(key)) {
+      newDate[i] = currentDateObj[key];
+    }
+
+    if (key === 'Y') {
+      if (toFormat[i].length > newDate[i].length) {
+        if (+newDate[i] < 30) {
```
denys-danyliuk, line 84: avoid nested cycles, try to do this without them

denys-danyliuk (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,46 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newDate = new Array(toFormat.length - 1);
+  const divider = fromFormat[fromFormat.length - 1];
+
+  ;
+
```
denys-danyliuk, line 57: ?😅

src/formatDate.js
```diff
@@ -50,7 +50,46 @@
...
+  let newDivider = '';
+
+  const dateChanged = date.split(divider);
```
denys-danyliuk, line 60:
```suggestion
  const splittedDate = date.split(divider);
```

denys-danyliuk (approved):

PR #884, opened 2022-11-01
format_date

SashaKornieiev:


src/formatDate.js
```diff
@@ -50,7 +50,33 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
-}
+  const arrDate = date.split(`${fromFormat[3]}`);
+  let str = '';
```
TarasHoliuk, line 54: don't use short forms for variable names
+ name should explain what is it

src/formatDate.js
```diff
@@ -50,7 +50,33 @@
...
+
+  const obj = {
```
TarasHoliuk, line 56: the same about variable name
and it could be just an empty object, because this `obj[fromFormat[i]] = arrDate[i];` will define new property even if it doesn't exist in your object before

TarasHoliuk (changes requested):

src/formatDate.js
AlisaVlasova: Hi, I see you haven't fixed previous comments, maybe you did not push your changes?
If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review 🙂

AlisaVlasova (changes requested):

qoa11a (approved):

PR #886, opened 2022-11-01
Solution

webdevw:


src/formatDate.js
```diff
@@ -50,7 +50,63 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const result = [];
+  const [, , , separateOld] = fromFormat;
```
Anastasiia-Svintsova, line 54:
```suggestion
  const separateOld = fromFormat[3];
```

src/formatDate.js
```diff
@@ -50,7 +50,63 @@
...
+  if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
+    reNewYear = oldData[indYear].slice(2);
+  }
+
+  if (fromFormat.includes('YY') && toFormat.includes('YYYY')) {
+    if (reNewYear < 30) {
+      reNewYear = `20${reNewYear}`;
+    } else {
+      reNewYear = `19${reNewYear}`;
+    }
```
Anastasiia-Svintsova, line 77: use a ternary operator here

Anastasiia-Svintsova (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,63 @@
...
+
+  if (
+    (toFormat.indexOf('YY' || 'YYYY') === 1 && toFormat.indexOf('MM') === 0)) {
+    result.push(oldData[indDay], reNewYear, oldData[indMonth]);
+
+    return result.reverse().join(separateNew);
+  }
+
+  if (toFormat.indexOf('MM') === 0 && toFormat.indexOf('DD') === 1) {
+    result.push(oldData[indMonth], oldData[indDay], reNewYear);
```
Bohdan-Maksymiuk, line 106: I believe it could be better solutions then 4 if's for all cases
you may create object with date, where keys will be date elements (`MM`, `YY` and so on), and access values later

Bohdan-Maksymiuk (commented):
feel free to ask in `fe_chat` for help

src/formatDate.js
```diff
@@ -50,7 +50,45 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const result = [];
+  const separateOld = fromFormat[3];
+  const separateNew = toFormat[3];
+
+  const oldData = date.split(separateOld);
+  const indDay = fromFormat.indexOf('DD');
+  const indMonth = fromFormat.indexOf('MM');
+  let indYear = fromFormat.indexOf('YY');
```
qoa11a, line 60:
```suggestion
  const idxDay = fromFormat.indexOf('DD');
  const idxMonth = fromFormat.indexOf('MM');
  let idxYear = fromFormat.indexOf('YY');
```
It's more common to use this short form of "index"

qoa11a (approved):

PR #890, opened 2022-11-03
Sol

TarasV123:


src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
+  }
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    if (toFormat[i] === 'DD') {
+      dateConvert.push(dateObject.day);
+    }
+
+    if (toFormat[i] === 'MM') {
+      dateConvert.push(dateObject.month);
+    }
```
AlisaVlasova, line 82: об'єднай це все в один цикл і використай switch case замість if

AlisaVlasova (changes requested):

tetiana-ivanchenko (changes requested):
did you really fix prev comment? did you combine the loops?

src/formatDate.js
TarasV123: I have to make one out of two cycles, and don't make other cycles???
Because I have no idea how to solve this task with one cycles

TarasV123: If this is possible, give me a hint

liliia-livandovskaya (approved):
Great job! 🔥

PR #893, opened 2022-11-05
Solution

s3rhiienko:


src/formatDate.js
```diff
@@ -50,7 +50,66 @@
...
+        year = `20${param}`;
+        break;
+
+      case format.length === 2:
+        year = year.slice(2);
+        break;
+    }
+
+    return year;
+  }
```
qoa11a, line 94: Move helper functions outside main function.

qoa11a (changes requested):

PR #895, opened 2022-11-06
Solution

natalia-maryshko:


src/formatDate.js
```diff
@@ -50,7 +50,30 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldFormat = date.split(fromFormat[3]);
+  const newFormat = [];
+
+  for (let i = 0; i < 3; i++) {
+    for (let k = 0; k < 3; k++) {
```
polosanya, line 57: Using nested loops is a bad practice.
It makes your code too unoptimized

src/formatDate.js
```diff
@@ -50,7 +50,30 @@
...
+      if (fromFormat[i] === 'YYYY' && toFormat[k] === 'YY') {
+        newFormat[k] = oldFormat[i].slice(2);
+      }
+
+      if (fromFormat[i] === 'YY' && toFormat[k] === 'YYYY') {
+        if (oldFormat[i] < 30) {
```
polosanya, line 63: It is also better to avoid nesting if...else, try to use ternary operator in this case
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator

polosanya (changes requested):
Good job! 
Please pay attention to my comments and try to optimize your code

src/formatDate.js
natalia-maryshko: thanks for advice)

liliia-livandovskaya (approved):
Looks good to me 🔥

PR #896, opened 2022-11-07
Added the solution

zynkro:


amitpatiljc (changes requested):
Good job , just try to accomplish this with `switch` statement, instead of `if`

zbyn3k:
Hi Amit, should be good now. Thanks for the feedback!

amitpatiljc (approved):
Good work 👍

PR #901, opened 2022-11-13
v01 Solution

home2ego911:


src/formatDate.js
```diff
@@ -50,7 +50,46 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateSplited = date.split(fromFormat[fromFormat.length - 1]);
+  const glue = toFormat[toFormat.length - 1];
+
+  const objDate = {};
```
vlad-demchuk, line 56: don't cut names of variables

src/formatDate.js
```diff
@@ -50,7 +50,46 @@
...
+
+  for (const part of toFormat) {
+    switch (part) {
+      case 'DD':
+        result.push(objDate[part]);
+        break;
+
+      case 'MM':
+        result.push(objDate[part]);
+        break;
```
vlad-demchuk, line 72: you do the same twice. just combine cases

Virass (approved):

PR #902, opened 2022-11-13
solution

olenitut:


src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+  if (year.length !== toFormat[newYearIndex].length) {
+    if (year.length === 2) {
+      if (year < 30) {
+        year = `20${year}`;
+      } else {
+        year = `19${year}`;
+      }
+    } else {
+      year = year.slice(2);
+    }
```
Bohdan-Maksymiuk, line 77: avoid such nesting `else if` conditions

src/formatDate.js
```diff
@@ -50,7 +50,36 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateArr = date.split(fromFormat[3]);
+  const day = dateArr[fromFormat.indexOf('DD')];
+  const month = dateArr[fromFormat.indexOf('MM')];
+  const oldYearIndex = fromFormat.indexOf('YY') === -1
+    ? fromFormat.indexOf('YYYY') : fromFormat.indexOf('YY');
+  let year = dateArr[oldYearIndex];
```
Bohdan-Maksymiuk, line 58: try to use object to store date values
for example
```
  const dateObj = {
	  DD: '',
	  MM: '',
	  YY: '',
	  YYYY: '',
  };
```

Bohdan-Maksymiuk (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,38 @@
...
+  newDateArr[toFormat.indexOf('DD')] = dateObj.day;
+  newDateArr[toFormat.indexOf('MM')] = dateObj.month;
+
+  const newYearIndex = toFormat.indexOf('YY') === -1
+    ? toFormat.indexOf('YYYY') : toFormat.indexOf('YY');
+
+  const currentYearFormat = dateObj.year.length;
+  const expectedYearFormat = toFormat[newYearIndex].length;
+
+  if (currentYearFormat > expectedYearFormat) {
```
Viktor-Kost, line 74: Good!
You avoid nesting `else if `conditions. Fine)

And also you can make separate variables for `if` conditions to make it again more readable

Viktor-Kost (approved):

PR #904, opened 2022-11-13
commit

OleksChernikov:


src/formatDate.js
```diff
@@ -49,8 +49,56 @@
...
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    const current = fromFormat[i];
+
+    if (current.includes('Y') && toFormat.indexOf(current) === -1) {
+      year = dateArrFrom[i];
+
+      if (current === 'YYYY') { 
+        year = year.slice(2); 
+      } else if (current === 'YY') {
```
polosanya, line 68: Try to avoid using else if 
Use switch...case and/or ternary operator instead

polosanya (changes requested):
Good job! Please pay attention to my comments

qoa11a (approved):

PR #905, opened 2022-11-14
dateFormatted

Mykola-Busarhin:


src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const firstDivider = fromFormat[3];
+  const secondDivider = toFormat[3];
+  const originalDate = date.split(firstDivider);
+
+  const reverseDate = [];
+  let year;
+  let month;
+  let day;
```
Bohdan-Maksymiuk, line 60: You can create object  for this purpose
```   
  const dateObj = {
	DD: '',
	MM: '',
	YY: '',
	YYYY: '',
    }; 
  ```

File: src/formatDate.js [hunk 2]
```diff
@@ -50,7 +50,59 @@
...
+        break;
+
+      case 'YY':
+        if (+originalDate[i] < 30) {
+          year = 20 + originalDate[i];
+        } else {
+          year = 19 + originalDate[i];
+        }
+        break;
+    }
```
Bohdan-Maksymiuk, line 83: and operate with object here

src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+      case 'MM':
+        reverseDate.push(month);
+        break;
+      case 'YYYY':
+        reverseDate.push(year);
+        break;
+      case 'YY':
+        reverseDate[i] = year.slice(2);
+        break;
+    }
```
Bohdan-Maksymiuk, line 100: and here `toFormat[i]` will be the key of your object

Bohdan-Maksymiuk (changes requested):
Good job 👍 
Lets refactor our code a little bit

src/formatDate.js
Mykola-Busarhin: I have made the changes you suggested, but, anyway, what is the point of the whole thing? It is supposed to make my code more readable and understandable, but does it in this case? Could you explain that, please?

src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+        dateObj.DD = originalDate[i];
+        break;
+
+      case 'MM':
+        dateObj.MM = originalDate[i];
+        break;
+
+      case 'YYYY':
+        dateObj.YYYY = originalDate[i];
+        break;
```
Bohdan-Maksymiuk, line 74:
```suggestion
  dateObj[fromFormat[i]] = originalDate[i];
```
smth like that instead of all cases

src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+        reverseDate.push(dateObj.DD);
+        break;
+      case 'MM':
+        reverseDate.push(dateObj.MM);
+        break;
+      case 'YYYY':
+        reverseDate.push(dateObj.YYYY);
+        break;
+      case 'YY':
+        reverseDate[i] = dateObj.YYYY.slice(2);
```
Bohdan-Maksymiuk, line 98: and rewrite logic here
toFormat[i] - is a key of your date object

Bohdan-Maksymiuk (changes requested):

src/formatDate.js
Bohdan-Maksymiuk: Yes. You store all values in 1 object and then dynamically (based on key (YY, DD, etc)) can write and access the value. This code will be more optimised and fast

src/formatDate.js
Mykola-Busarhin: One more PR. Spent a lot of time to complete it though can`t get rid of those if-blocks. How else can I manipulate the 'year' fotmat!? Anyway, this is my final version for now.

SanyaBratashchuk (approved):

PR #906, opened 2022-11-15
Solution

anton-sosonskyi:
I think, I created a monster, but I did my best)))

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -51,6 +51,37 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const dataArr = date.split(fromFormat[3]);
+  const updatedDate = [];
+  const updatedYear = (lengthDiff, year) => {
+    if (lengthDiff > 0) {
+      return year < 30 ? `20${year}` : `19${year}`;
+    } else if (lengthDiff < 0) {
+      return year.slice(2);
+    }
+
+    return year;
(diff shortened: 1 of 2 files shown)


src/formatDate.js
```diff
@@ -51,6 +51,37 @@
...
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const dataArr = date.split(fromFormat[3]);
+  const updatedDate = [];
+  const updatedYear = (lengthDiff, year) => {
+    if (lengthDiff > 0) {
+      return year < 30 ? `20${year}` : `19${year}`;
+    } else if (lengthDiff < 0) {
+      return year.slice(2);
+    }
```
SanyaBratashchuk, line 61:
```suggestion
    } 
    
    if (lengthDiff < 0) {
      return year.slice(2);
    }
```

src/formatDate.js
```diff
@@ -51,6 +51,37 @@
...
+
+    return year;
+  };
+
+  const yearData = (format) => {
```
SanyaBratashchuk, line 66:
```suggestion
  const yearData = (format) => {
```
Function's name should be verb so fix it everywhere

SanyaBratashchuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #908, opened 2022-11-19
solution 1.0

aliliolek:


src/formatDate.js
```diff
@@ -51,6 +51,54 @@
...
+  const newObject = {};
+
+  newObject[`${fromFormat[0]}`] = fromDayArray[0];
+  newObject[`${fromFormat[1]}`] = fromDayArray[1];
+  newObject[`${fromFormat[2]}`] = fromDayArray[2];
+
+  const toDayArray = [];
+
+  for (let i = 0; i < 3; i++) {
+    for (const key in newObject) {
```
TarasHoliuk, line 65: It's better to do 2 separate loops (with their own `switch`) instead of nested loops

src/formatDate.js
```diff
@@ -51,6 +51,54 @@
...
+      if (toFormat[i] === key) {
+        toDayArray[i] = newObject[key];
+      }
+
+      switch (key) {
+        case 'YY':
+          switch (toFormat[i]) {
```
TarasHoliuk, line 72: don't use nested `switch`

TarasHoliuk (changes requested):

PR #910, opened 2022-11-20
Solution

OleksandrOse:


src/formatDate.js
```diff
@@ -51,6 +51,52 @@
...
+    switch (true) {
+      case fromFormat[i] === 'YY' || fromFormat[i] === 'YYYY':
+        year = dateArr[i];
+        break;
+      case fromFormat[i] === 'MM':
+        mounth = dateArr[i];
+        break;
+      case fromFormat[i] === 'DD':
+        day = dateArr[i];
+        break;
```
ivanchuk-yurii, line 72: if separator will be `DD` or 'MM' ..., it break logic)

ivanchuk-yurii (approved):

PR #913, opened 2022-11-22
Solution

vadiimvooo:


src/formatDate.js
```diff
@@ -50,7 +50,43 @@
...
-  // write code here
+  const splitedDate = date.split(fromFormat[3]);
+  const dateObject = {};
+  const dateResult = [];
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    if (fromFormat[i] === 'DD') {
+      dateObject.DD = splitedDate[i];
+    } else if (fromFormat[i] === 'MM') {
+      dateObject.MM = splitedDate[i];
```
shyptia, line 61: I think you don't need to check if fromFormat[i] is equal to 'DD' or 'MM', you can just add it to dateObject

src/formatDate.js
```diff
@@ -50,7 +50,43 @@
...
+    } else {
+      if (splitedDate[i].length === 2) {
+        if (splitedDate[i] < 25) {
```
shyptia, line 64: When converting from YY to YYYY use 20YY if YY < 30

shyptia (approved):

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+  const splitedDate = date.split(fromFormat[3]);
+  const dateObject = {};
+  const dateResult = [];
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    dateObject[fromFormat[i]] = splitedDate[i];
+
+    if (fromFormat[i].includes('Y')) {
+      if (splitedDate[i].length === 2) {
+        if (splitedDate[i] < 30) {
```
solaryasha, line 62: No, let's avoid such nesting. Let's check exact value `YY`

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+          dateObject.YYYY = '20' + splitedDate[i];
+        } else {
+          dateObject.YYYY = '19' + splitedDate[i];
+        }
```
solaryasha, line 66: It would be great if you create function for this formatting

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+        dateObject.YYYY = splitedDate[i];
+      }
+    }
+  }
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    if (toFormat[i] === 'DD') {
+      dateResult.push(dateObject.DD);
+    } else if (toFormat[i] === 'MM') {
+      dateResult.push(dateObject.MM);
```
solaryasha, line 77: You could use dynamic keys, something like this
```suggestion
      dateResult.push(dateObject[toFormat[i]]);
```
And you will have only one exception - `YY`

solaryasha (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    if (toFormat[i] !== 'YY') {
+      dateResult.push(dateObject[toFormat[i]]);
+    } else {
+      if (toFormat[i].length === 2) {
+        dateResult.push(dateObject.YYYY.slice(2, 4));
+      } else {
+        dateResult.push(dateObject.YYYY);
+      }
```
solaryasha, line 73: Do we need this `else` statement here?
If you remove it, your code still should work

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const splitedDate = date.split(fromFormat[3]);
+  const dateObject = {};
```
solaryasha, line 54:
```suggestion
  const dateMapping = {};
```

solaryasha (changes requested):

solaryasha (approved):
It looks better now, good job!

PR #914, opened 2022-11-22
WWWWW

yurii-filiushkin:


src/formatDate.js
```diff
@@ -50,7 +50,43 @@
...
+  }
+
+  for (let u = 0; u < toFormat.length; u++) {
+    if (toFormat[u] === 'YY') {
+      year = year[2] + year[3];
+    }
+  }
+
+  for (let y = 0; y <= 2; y++) {
+    for (let t = 0; t <= 2; t++) {
```
TarasHoliuk, line 78: you can do it without a nested loop
and 2 loops could be enough

TarasHoliuk: try to simplify it, feel free to ask help in chat if you need)

TarasHoliuk (changes requested):

vlad-demchuk (approved):

PR #916, opened 2022-11-24
Solution

UmizDemud:


src/formatDate.js
```diff
@@ -48,9 +48,57 @@
...
+      }
+    } else if (toFormat[i] === 'YYYY') {
+      if ('YYYY' in dictionary) {
+        result += dictionary['YYYY'];
+      } else {
+        if (dictionary['YY'][0] < '3') {
+          result += '20' + dictionary['YY'];
+        } else {
+          result += '19' + dictionary['YY'];
+        }
```
mgrinko, line 92: Too many `if/else`!
Please simplify
I think the whole solution can ve written in 10 lines

mgrinko (changes requested):

src/formatDate.js
```diff
@@ -48,9 +48,25 @@
  *
  * @returns {string}
  */
-
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateSplit = date.split(fromFormat[3]);
+
+  return toFormat.slice(0, 3).reduce((accumulator, current, i) => {
```
meforoma, line 54: let's destructure:
```suggestion
  const [fromValue1, fromValue2, fromDelimiter] = fromFormat;
  const [toValue1, toValue2, toDelimiter] = toFormat;
  const dateSplit = date.split(fromDelimiter);

  return toFormat.slice(0, 3).reduce((accumulator, current, i) => {
```

src/formatDate.js
```diff
@@ -48,9 +48,25 @@
...
+    const index = fromFormat.findIndex(field => current === field);
+    let running; // variable to add to compilation for each date field
+
+    if (index !== -1) { // exact value taken from the initial date
+      running = dateSplit[index];
+    } else if (current === 'YYYY') { // value is asked in YYYY format
+      const years = dateSplit[fromFormat.indexOf('YY')]; // date in YY format
+
+      running = years[0] < 3 ? '20' + years : '19' + years; // century added
```
meforoma, line 63: multi-line all ternaries

src/formatDate.js
UmizDemud: But I'd need the indexes in the ```fromFormat```.. I'll spend some time on it.

UmizDemud: I looked for a spread operation like 
```js
const [...fromValues, fromDelimiter] = fromFormat;
```
Apparently can't do that, let me push the solution according to the rest of the comments. Haven't been able to figure out a cool solution here, yet

naziamoff (changes requested):
let's create variables for all the magic numbers. Now it is not clear to me why you have `-1` or `3`. It is a bit difficult to read the code this way, so would be great to improve)

naziamoff (approved):
🔥

PR #918, opened 2022-11-25
Solution

halynarydasova:


src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
-  // write code here
+  const oldFormat = [];
+  const oldFormatArray = date.split(fromFormat[3]);
+  const newFormat = [];
+
+  for (let i = 0; i < (fromFormat.length - 1); i++) {
+    if (fromFormat[i] === 'YY') {
+      if (oldFormatArray[i] > 22) {
+        oldFormatArray[i] = '19' + oldFormatArray[i];
+      } else {
```
polosanya, line 61: You can use the ternary operator instead to avoid nesting `if...else` statements.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator

polosanya (changes requested):
Good job! Please check my comment to make your code a bit better 😉

polosanya (approved):
Well done!

PR #920, opened 2022-11-25
solution

Lun4trick:


src/formatDate.js
```diff
@@ -50,7 +50,66 @@
...
+
+  for (const partTo of toFormat) {
+    switch (partTo) {
+      case 'YY':
+        if (partTo.length === dateParts.year.length) {
+          toFormat[toFormat.indexOf(partTo)] = dateParts.year;
+          break;
+        } else {
+          toFormat[toFormat.indexOf(partTo)]
+          = dateParts.year.slice(2, dateParts.year.length);
```
nurdanokcu, line 87: if you write slice(2) only it will work still. Because it starts from 2 and goes till the end anyway.

src/formatDate.js
```diff
@@ -50,7 +50,66 @@
...
+  const dateParts = {
+    separator: fromFormat[fromFormat.length - 1],
+  };
+
+  const separateDate = date.split(dateParts.separator);
+
+  for (const partFrom of fromFormat) {
+    switch (partFrom) {
+      case 'YY':
+        dateParts.year = separateDate[fromFormat.indexOf(partFrom)];
```
nurdanokcu, line 62: Since you need indexes and indexOf(partFrom) is repeating , using for loop with  indexes could be a better idea.

nurdanokcu (approved):
Good usage of objects.

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+    MM: separateDate[fromFormat.indexOf('MM')],
+    DD: separateDate[fromFormat.indexOf('DD')],
+  };
+
+  const toYear = (toFormat.includes('YYYY'))
+    ? 'YYYY'
+    : 'YY';
+
+  switch (true) {
+    case (toYear.length < dateParts.year.length):
```
naziamoff, line 68: now it is totally unclear to me what happens in the code, some magic 😄
Let's create some variables for the conditions to make it more clear

naziamoff (changes requested):

naziamoff (approved):
🔥

PR #921, opened 2022-11-25
format day Valentos

mr-valentos:


GabrielMaksimovich (changes requested):
Its better to use switch in the second loop

GabrielMaksimovich (approved):
Nice!

valerii-derkach (changes requested):

PR #923, opened 2022-11-25
Solution

nurdanokcu:


src/formatDate.js
```diff
@@ -50,7 +50,78 @@
...
+function yearFormat(year, format) {
+  if (year.length === 4 && format === 'YY') {
+    return year.slice(2);
+  }
+
+  if (year.length === 2 && format === 'YYYY') {
+    if (parseInt(year) < 30) {
+      return '20' + year;
+    } else {
+      return '19' + year;
```
Lun4trick, line 120:
```suggestion
return (parseInt(year) < 30)
    ?  '20' + year
    :  '19' +year;
```
i think its better to use ternary situations like this :)

Lun4trick (approved):
Good job :) 👍

src/formatDate.js
nurdanokcu: Thank you, will do so. 👍

src/formatDate.js
```diff
@@ -50,7 +50,76 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const currentSep = fromFormat[3];
```
naziamoff, line 53: what is sep?

src/formatDate.js
```diff
@@ -50,7 +50,76 @@
...
+
+      default:
+        break;
+    }
+  }
+
+  return dateAfter.join(finalSep);
+}
+
+function yearFormat(year, format) {
```
naziamoff, line 111: any ideas on improving the name of this function?

naziamoff (approved):
Well done, just some naming comments

src/formatDate.js
nurdanokcu: I wanted to name it currentSeperator but then it was gonna be very long.

src/formatDate.js
nurdanokcu: I think formatYear or formatTheYear is better because it is a function. I will improve it.

naziamoff: Well done)

src/formatDate.js
naziamoff: it is much better to keep it long but clear than to have short unclear variables

PR #924, opened 2022-11-25
solution

lefteris-L:


src/formatDate.js
```diff
@@ -50,7 +50,67 @@
...
+
+  const format = {
+    from: fromFormat,
+    to: toFormat,
+  };
+  const positions = ['start', 'mid', 'end'];
+
+  for (const pos of positions) {
+    for (const source in format) {
+      const ind = positions.indexOf(pos);
```
ysrckr, line 74: i or index would be more understandable.

src/formatDate.js
```diff
@@ -50,7 +50,67 @@
...
+
+      if (format[`${source}`][ind].includes('Y')) {
+        year[`${source}`][`${pos}`] = true;
+        year[`${source}`][`${format[`${source}`][ind].length}`] = true;
```
ysrckr, line 78: you don't need to use interpolation. 
`format[source]`
would work too.

src/formatDate.js
ysrckr: Also when you need access index just use standard for loop. 
`
for (let i = 0; i < positions.length; i++) {...}
`
'i' is directly index

ysrckr (approved):
It passes the test.

src/formatDate.js
lefteris-L: Would be cleaner too

src/formatDate.js
```diff
@@ -50,7 +50,65 @@
...
+  const positions = ['start', 'mid', 'end'];
+
+  for (let i = 0; i < positions.length; i++) {
+    for (const source in format) {
+      if (format[source][i].includes('Y')) {
+        year[source][positions[i]] = true;
+        year[source][format[source][i].length] = true;
+      }
+    }
+  }
```
naziamoff, line 79: nested loops are evil 🙅

naziamoff (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,64 @@
...
+
+  const index = fromFormat.findIndex((element) => element.includes('Y'));
+  const y = year.from[2]
+    ? (newDate[index] < 30 ? '20' : '19') + newDate[index]
+    : newDate[index].slice(2);
+
+  switch (true) {
+    case (year.from.mid && year.to.end)
+      || (year.from.start && year.to.mid):
```
naziamoff, line 87: would be better if you created separate variables for conditions. Now it is quite difficult to read

naziamoff (approved):

PR #925, opened 2022-11-27
Solution

Deniskakaka:


OlEzhA0 (changes requested):
Good job! let's optimise your solution

PR #928, opened 2022-11-28
Format Date Advanced Added

serge-kontakt:


src/formatDate.js
```diff
@@ -50,7 +50,47 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  // if (fromFormat[0] === 'YYYY' && toFormat[0] === 'YYYY') {
+  //   return date.split(fromFormat[3]).join(toFormat[3]);
+  // }
+
+  // if (fromFormat[0] === 'YYYY' && toFormat[0] === 'DD') {
+  //   return date.split(fromFormat[3]).reverse().join(toFormat[3]);
+  // }
```
vlad-demchuk, line 59: remove comments

src/formatDate.js
```diff
@@ -50,7 +50,47 @@
...
+
+  const [data1, data2, data3] = date.split(fromFormat[3]);
+
+  fromFormat[0] = `${fromFormat[0]}` + `${data1}`;
+  fromFormat[1] = `${fromFormat[1]}` + `${data2}`;
+  fromFormat[2] = `${fromFormat[2]}` + `${data3}`;
```
vlad-demchuk, line 65: don't mutate input parameters

src/formatDate.js
```diff
@@ -50,7 +50,47 @@
...
+        toFormat[i] = `${fromFormat[y].slice(0, 2)}` + `${fromFormat[y].slice(6)}`;
+      } else {
+        if (toFormat[i].slice(0, 1) === fromFormat[y].slice(0, 1)) {
+          toFormat[i] = fromFormat[y];
+        }
+      }
+    }
+  }
+
+  for (let i = 0; i < 3; i++) {
```
vlad-demchuk, line 85: let's simplify it. avoid one loop.

vlad-demchuk (changes requested):

src/formatDate.js
serge-kontakt: It does not work with one loop. Please give me a hint.

loralevitska: use array methods `includes`, `split`, `indexOf`.  You can split you date and check values by array indexes. 
image

loralevitska (changes requested):
hi) left a tip in the thread of the previous comment from the mentor 
Try to improve you solution and fill free to ask in fe_chat

src/formatDate.js
serge-kontakt: Completely new solution were added

serge-kontakt: Solution #3 using includes, split, indexOf were added.

src/formatDate.js
```diff
@@ -50,7 +50,33 @@
...
+
+      if (arrDate[oldIndex] < 30) {
+        newElement = `20` + `${arrDate[oldIndex]}`;
+      } else {
+        newElement = `19` + `${arrDate[oldIndex]}`;
+      }
+    } else {
+      oldIndex = fromFormat.indexOf(key);
+      newElement = arrDate[oldIndex];
+    }
```
Bohdan-Maksymiuk, line 74: avoid such nesting `else if` conditions

Bohdan-Maksymiuk: do your really need all this code inside loop?
you can use different loops for your purposes

Bohdan-Maksymiuk (changes requested):
You can create object 
```   
const dateObj = {
	DD: '',
	MM: '',
	YY: '',
	YYYY: '',
  };
 

 Then 2 loops: 
 
``` const result = [];  result.push(dateObj[toFormat[i]]) ```

PR #932, opened 2022-11-29
Solution

vlad-boikov:
feels like a shitcode))))

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -50,34 +50,35 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
+  let result = '';
   const splitted = date.split(fromFormat[3]);
-  const result = [];
+  const dateObj = {
+    'DD': '',
+    'MM': '',
+    'YY': '',
+    'YYYY': '',
+  };
 
   for (let i = 0; i < fromFormat.length - 1; i++) {
-    for (let j = 0; j < toFormat.length - 1; j++) {
-      if (fromFormat[i] === toFormat[j]) {
-        result[j] = splitted[i];
-      };
+    if (fromFormat[i] === 'YY' && +splitted[i] < 30) {
+      dateObj['YYYY'] = `20${splitted[i]}`;
+    } else if (fromFormat[i] === 'YY' && +splitted[i] >= 30) {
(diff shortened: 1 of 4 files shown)


src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
+        && +splitted[i] >= 30) {
+        result[j] = `19${splitted[i]}`;
+        continue;
+      }
+
+      if (fromFormat[i] === 'YY' && toFormat[j] === 'YYYY') {
+        result[j] = `20${splitted[i]}`;
+      }
+    }
+  }
```
Bohdan-Maksymiuk, line 78: ohh, try to avoid such nested loops
it increases time complexity of your code

Bohdan-Maksymiuk (changes requested):
You can create object 

```
const dateObj = {
	DD: '',
	MM: '',
	YY: '',
	YYYY: '',
  };
```

Then 2 loops: 

src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+    'YY': '',
+    'YYYY': '',
+  };
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    if (fromFormat[i] === 'YY' && +splitted[i] < 30) {
+      dateObj['YYYY'] = `20${splitted[i]}`;
+    } else if (fromFormat[i] === 'YY' && +splitted[i] >= 30) {
+      dateObj['YYYY'] = `19${splitted[i]}`;
+    }
```
Bohdan-Maksymiuk, line 67: try to avoid `else if`
end lets remove such check from the loop

Bohdan-Maksymiuk: you may leave in the loop only
`dateObj[fromFormat[i]] = splitted[i];`

Virass (approved):

src/formatDate.js
Bohdan-Maksymiuk: after it you can calculate `dateObj.YYYY` and `dateObj.YY`

Bohdan-Maksymiuk (changes requested):

AlisaVlasova (approved):

PR #934, opened 2022-11-29
solution

Maks-Tsarenko:


src/formatDate.js
```diff
@@ -50,7 +50,42 @@
...
+  } else {
+    dateObj.YY = dateObj.YYYY.slice(2);
+  }
+
+  for (let i = 0; i < 3; i++) {
+    switch (toFormat[i]) {
+      case 'YYYY':
+      case 'YY':
+      case 'MM':
+      case 'DD':
```
DarkMistyRoom, line 83: Why do you need this switch?

src/formatDate.js
```diff
@@ -50,7 +50,42 @@
...
+
+  if (dateObj.YYYY === '') {
+    if (dateObj.YY >= 30) {
+      dateObj.YYYY = `19${dateObj.YY}`;
+    } else {
+      dateObj.YYYY = `20${dateObj.YY}`;
+    }
+  } else {
+    dateObj.YY = dateObj.YYYY.slice(2);
+  }
```
DarkMistyRoom, line 76: Try to avoid `if else` nesting. It is not good for reading and maintaining

DarkMistyRoom (changes requested):

PR #935, opened 2022-11-29
solution

superpooperxxx:


src/formatDate.js
```diff
@@ -50,7 +50,28 @@
...
+  const formattedDate = [];
+  const initialDateFormat = {};
+  const initialDateToArray = date.split(fromFormat[3]);
+
+  for (let i = 0; i < (fromFormat.length - 1); i++) {
+    if (fromFormat[i] === 'YY') {
+      initialDateFormat.YYYY = initialDateToArray[i] < 30
+        ? `20${initialDateToArray[i]}`
+        : `19${initialDateToArray[i]}`;
+    }
```
GabrielMaksimovich, line 62: Im not sure about this part, maybe its p4 or I m wrongm really hard to say. But whole solution very good!

GabrielMaksimovich (approved):

src/formatDate.js
valerii-derkach: yes, it is possible to do it without `if` including,

valerii-derkach (approved):

PR #936, opened 2022-11-29
Solution

IrynaK-a:


src/formatDate.js
```diff
@@ -50,7 +50,65 @@
...
+  let oldSeparator = '';
+  let newSeparator = '';
+  let oldYearIndex = '';
+  let oldYearLength = '';
+  let oldMonthIndex = '';
+  let oldDayIndex = '';
+  const newDate = [];
+
+  for (const itemFrom of fromFormat) {
+    if (itemFrom.includes('Y')) {
```
polosanya, line 62: You can use [switch...case](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) to provide all the checks properly

src/formatDate.js
```diff
@@ -50,7 +50,65 @@
...
+      oldSeparator = itemFrom;
+    }
+  }
+
+  const oldDateArray = date.split(`${oldSeparator}`);
+  const shortYear = oldDateArray[oldYearIndex].slice(-2);
+  const longYear = (shortYear < 30) ? `20${shortYear}` : `19${shortYear}`;
+
+  for (const itemTo of toFormat) {
+    if (itemTo.includes('D')) {
```
polosanya, line 85: Same here

src/formatDate.js
```diff
@@ -50,7 +50,65 @@
...
+      newDate[toFormat.indexOf(itemTo)] = oldDateArray[oldDayIndex];
+    }
+
+    if (itemTo.includes('M')) {
+      newDate[toFormat.indexOf(itemTo)] = oldDateArray[oldMonthIndex];
+    }
+
+    if (itemTo.includes('Y')) {
+      switch (oldYearLength) {
+        case '2':
```
polosanya, line 95: Do you really need switch here only for one case?

polosanya (changes requested):
Good job! Please check the comments to improve your code a bit

polosanya (approved):
Good job!

PR #940, opened 2022-11-30
solution

Mykhailyk-Vadym:


superpooperxxx (approved):
Good job, but I'd recommend using object here instead of 4 different variables, because you would not need to have the switch construction then. Those 'DD', 'MM' could be keys of the object.

Ihor-Karpyn (approved):
Good job

PR #943, opened 2022-12-01
finished

jackelashkar:


src/formatDate.test.js
```diff
@@ -53,6 +53,19 @@ describe('formatDate', () => {
         .toBe('21-12-2012');
     });
 
+    it(`for non standard parts order '12-2012-21'
```
naziamoff, line 56: when you write tests it is better to have more clear description
But good initiative to learn writing tests 👍

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+  }
+
+  // we check if we don't already have the format we need
+  if (toFormat.includes('YYYY') && ourFormat.YYYY === null) {
+    if (+ourFormat.YY < 30) {
+      ourFormat.YYYY = 20 + ourFormat.YY;
+    } else {
+      ourFormat.YYYY = 19 + ourFormat.YY;
+    }
+  }
```
naziamoff, line 79: nested `if's` are hell, it is better to avoid them by all costs

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+    YY: null,
+    YYYY: null,
+  };
+
+  for (let i = 0; i < 3; i++) {
+    ourFormat[fromFormat[i]] = ourDate[i];
+  }
+
+  // we check if we don't already have the format we need
+  if (toFormat.includes('YY') && ourFormat.YY === null) {
```
naziamoff, line 68: is there a reason for checking for exactly `null` value? Why not we just checking whether it is not falsy?
```suggestion
  if (toFormat.includes('YY') && !ourFormat.YY) {
```

naziamoff (approved):
LGTM!
Just a couple of suggestions

PR #953, opened 2022-12-04
Solution

illiashcherbak:


src/formatDate.js
```diff
@@ -50,7 +50,61 @@
...
+      case shortYears:
+        if (sortedDate[fullYears]) {
+          const shortenedYear = sortedDate[fullYears].slice(-2);
+
+          formatedDate.push(shortenedYear);
+        } else {
+          formatedDate.push(sortedDate[shortYears]);
+        }
+        break;
+    }
```
Anastasia4sik, line 104: I think you need to add `default`

Anastasia4sik (approved):

src/formatDate.js
```diff
@@ -50,7 +9,64 @@
...
+        break;
+
+      case months:
+        formatedDate.push(sortedDate[months]);
+        break;
+
+      case fullYears:
+        let lengthenedYears;
+
+        if (sortedDate[shortYears] < 30) {
```
ivan-kadykalo, line 41: This part looks like a nested if statement (case + if) 
image

ivan-kadykalo (changes requested):
You are almost there 🙃

ivan-kadykalo (approved):
You still have nested `if` inside `case`. This is the same as nested ifs. Try to update your year value before `for` and later in your case use prepared value

PR #956, opened 2022-12-08
Solution

nzrii-kysel:


src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
+        ? fromFormat[i] + 'YY'
+        : fromFormat[i].slice(-2);
+      index = toFormat.indexOf(fromFormat[i]);
+      dateArray[i] = formatYear(dateArray[i]);
+    }
+
+    formatedArray[index] = dateArray[i];
+  }
+
+  return formatedArray.join(toFormat[3]);
```
Malva37, line 80: You use toFormar[3] twice, maybe it's better create separate variable, it will do your code more understandable

Malva37 (approved):

src/formatDate.js
nzrii-kysel: good catch, thanks

ArturErmolenko (approved):
Well done 👍

PR #958, opened 2022-12-10
Solution

Iksich:


src/formatDate.js
```diff
@@ -50,7 +50,58 @@
...
+    if (fromFormat[i] === 'MM') {
+      month = dateArr[i];
+    }
+
+    if (fromFormat[i] === 'YY' || fromFormat[i] === 'YYYY') {
+      year = dateArr[i];
+    }
+  }
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
```
naziamoff, line 74: let's try to refactor this block and avoid nested `if/ if/else`

naziamoff (changes requested):

meforoma (approved):
now it's more concise 👍🏼

PR #960, opened 2022-12-11
add solution

constantweb:


src/formatDate.js
```diff
@@ -50,7 +50,55 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let day;
+  let month;
+  let year;
```
vlad-demchuk, line 55: assign variable at once. use `null` if you don't need any values yet

vlad-demchuk (changes requested):

TarasHoliuk (approved):

AlisaVlasova (approved):

PR #965, opened 2022-12-14
solution

Nikoramo:


src/formatDate.js
```diff
@@ -50,7 +50,37 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
-}
+  const [ ,,, oldSeparator ] = fromFormat;
+  const [ ,,, newSeparator ] = toFormat;
+  const splitDate = date.split(oldSeparator);
```
DarkMistyRoom, line 55:
```suggestion
  const splitedDate = date.split(oldSeparator);
```

src/formatDate.js
```diff
@@ -50,7 +50,37 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
-}
+  const [ ,,, oldSeparator ] = fromFormat;
+  const [ ,,, newSeparator ] = toFormat;
```
DarkMistyRoom, line 54: It looks a bit strange. I'd prefer to use `const oldSeparator = fromFormat[3];`

DarkMistyRoom (approved):

PR #977, opened 2022-12-25
Solution

Malva37:


NVyrsta (approved):
Interesting solution👍

src/formatDate.js
```diff
@@ -50,7 +50,35 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let separ = fromFormat.splice(-1);
```
ArturErmolenko, line 53:
```suggestion
  let separator = fromFormat.splice(-1);
```
interesting naming :)

ArturErmolenko (approved):

PR #981, opened 2022-12-27
Solution

andrieieva-yuliia:


src/formatDate.js
```diff
@@ -50,7 +50,38 @@
...
+    const to = toFormat[i];
+
+    if (to === 'YY') {
+      finallDate.push(fullDate['YYYY'].slice(2));
+    } else {
+      finallDate.push(fullDate[to]);
+    }
+  }
+
+  return finallDate.join(toFormat[3]);
```
Malva37, line 84: You can use separate variable to toFormat[3], because you use it twice,it will make your code more understandable

Malva37 (approved):
well done)

src/formatDate.js
Denchepornyuk: agree

Denchepornyuk (approved):

PR #982, opened 2022-12-27
Solution

undepend:


src/formatDate.js
```diff
@@ -50,7 +50,38 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const prevDelim = fromFormat.pop();
```
witflash, line 53: What does mean `delim`? Is this `delimeter`? Don't use shorthand for naming. And it should name `saparator` I think.

src/formatDate.js
```diff
@@ -50,7 +50,38 @@
...
+  const curDelim = toFormat.pop();
+  const curDate = date.split(prevDelim);
```
witflash, line 55: And here will be `currentDate` and `currentSeparator`

witflash (changes requested):
Good job!
But pay attention to comments.

OlEzhA0 (approved):
Good Job!

PR #987, opened 2022-12-29
Solution

oleh-holovnykh:


sv-dubov (approved):
Interesting solution with switch :)

src/formatDate.js
```diff
@@ -50,7 +50,47 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateValues = date.split(fromFormat[3]);
+  const dateObj = {};
```
Anna-Tkachenko, line 54: don't use data type name in variable name

Anna-Tkachenko (approved):
Good job! 🔥

PR #991, opened 2022-12-30
Solution was realised

anna-inozemets:


andrieieva-yuliia (approved):

src/formatDate.js
```diff
@@ -50,7 +50,45 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparate = fromFormat[fromFormat.length - 1];
+  const toSeparate = toFormat[toFormat.length - 1];
+  const oldDateArr = date.split(fromSeparate);
+  const oldDateObj = {};
```
Anna-Tkachenko, line 56: don't use data type name in variable name

src/formatDate.js
```diff
@@ -50,7 +50,45 @@
...
+      if (oldDateObj[yearFormatShort] < 24) {
+        yearOfDate = `20${oldDateObj[yearFormatShort]}`;
+        formatedDate.push(yearOfDate);
+      }
+
+      if (oldDateObj[yearFormatShort] > 23) {
+        yearOfDate = `19${oldDateObj[yearFormatShort]}`;
+        formatedDate.push(yearOfDate);
+      }
+    }
```
Anna-Tkachenko, line 83: can we simplify this logic for not using nested ifs?

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! ⭐️

PR #993, opened 2022-12-31
Result

svitjojo:


OlhaYevstifieieva (approved):
I think, you can do your code shorter, but anyway good job!

PR #1002, opened 2023-01-03
commit message

jaxxxon78:


src/formatDate.js
```diff
@@ -50,7 +50,30 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
-}
+  const currentDateOld = date.split(fromFormat[3]);
```
Viktor-Kost, line 53: so is it `current` or an `old` one? Update naming

src/formatDate.js
```diff
@@ -50,7 +50,30 @@
...
+  const res = [];
+  const dateInfo = {};
+
+  for (let i = 0; i < 3; i++) {
+    if (fromFormat[i] === 'YY') {
+      if (currentDateOld[i] < 30) {
+        dateInfo['YYYY'] = `20${currentDateOld[i]}`;
+      } else {
+        dateInfo['YYYY'] = `19${currentDateOld[i]}`;
+      }
```
Viktor-Kost, line 63: Try to avoid using `if else `statement
I would rather use ternary operator here. It looks clear

src/formatDate.js
```diff
@@ -50,7 +50,30 @@
...
+
+    dateInfo[fromFormat[i]] = currentDateOld[i];
+  }
 
+  for (let j = 0; j < 3; j++) {
+    if (toFormat[j] === 'YY') {
+      res.push(dateInfo['YYYY'].slice(2));
+    } else {
+      res.push(dateInfo[toFormat[j]]);
+    }
```
Viktor-Kost, line 74: the same with ternary

src/formatDate.js
```diff
@@ -50,7 +50,30 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
-}
+  const currentDateOld = date.split(fromFormat[3]);
+  const res = [];
+  const dateInfo = {};
+
+  for (let i = 0; i < 3; i++) {
```
Viktor-Kost, line 57: It is better not use "magic" numbers as `3`. It's hard to understand what does it mean. 
I would recommend to use additional constants for these numbers
Ex: 
`const MY_VARIABLE = 3`

Viktor-Kost (changes requested):
Nice!

Left some recommendations

src/formatDate.js
```diff
@@ -50,7 +50,26 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
-}
+  const myIndex = 3;
```
Viktor-Kost, line 53: Ok, yet for constants we use uppercase.
And it's gonna be meaningful naming
`const MY_CONSTANT = "some-value";`

Viktor-Kost (approved):
Nice!)

I left one comment for you to know

PR #1004, opened 2023-01-03
Solution

Coyoote:


PolinaSukhorukova (approved):

src/formatDate.js
```diff
@@ -50,7 +50,35 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const objDate = {};
```
Anna-Tkachenko, line 53: don't use data type name in variable name

Anna-Tkachenko (approved):
Good job! 🔥

PR #1005, opened 2023-01-03
Solution

vanst1:


src/formatDate.js
```diff
@@ -50,7 +50,33 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparator = fromFormat[fromFormat.length - 1];
+  const toSeparator = toFormat[toFormat.length - 1];
```
Serghii29, line 54: in my opinion, you can not use this variables

src/formatDate.js
```diff
@@ -50,7 +50,33 @@
...
+          newFormat.push(`20${dateObject.YY}`);
+        } else {
+          newFormat.push(`19${dateObject.YY}`);
+        }
+      } else {
+        newFormat.push(`${dateObject.YYYY.slice(-2)}`);
+      }
+    } else {
+      newFormat.push(dateObject[toFormat[i]]);
+    }
```
Serghii29, line 76: it's bad idea to write if in if and this code is not readable

Serghii29:
Good job🙌

ArturErmolenko (approved):

PR #1012, opened 2023-01-03
Solution

BrandonSunrise:


src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+  const finallDate = [];
+  let sep = fromFormat[3];
+  const dateArr = date.split(sep);
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    const from = fromFormat[i];
+    const initialDate = dateArr[i];
+
+    if (from === 'YY') {
+      if (initialDate < 30) {
```
anna-inozemets, line 63: Maybe it is possible to realise this part without nested if

anna-inozemets (approved):
Pretty nice job!🐇

ArturErmolenko (approved):
It will be great not to use nested if

PR #1013, opened 2023-01-03
solution

ann-yurkevych:


src/formatDate.js
```diff
@@ -51,6 +51,43 @@
...
   // write code here
+  const newDate = [];
+  const oldDate = date.split(fromFormat[fromFormat.length - 1]);
+
+  function findSameElement(index) {
+    for (const oldElement of fromFormat) {
+      const indexOfOld = fromFormat.indexOf(oldElement);
+      let newElement = oldDate[indexOfOld];
+
+      if (oldElement === toFormat[index]) {
```
oleh-holovnykh, line 62: with **for** (let i = 0; i < fromFormat - 1; i++)
there will be no need to use indexOf

indexOfOld === i
oldElement === fromFormat[i]
newElement === oldDate[i]

and fromFormat[i] === toFormat[index] looks clearer

src/formatDate.js
```diff
@@ -51,6 +51,43 @@
...
+      } else if (isBothYears && isNewLengthBigger) {
+        newElement = 19 + oldDate[indexOfOld];
+        newDate.push(newElement);
+      }
+    }
+  }
+
+  findSameElement(0);
+  findSameElement(1);
+  findSameElement(2);
```
oleh-holovnykh, line 88: the code repeats three times from 0 to 2,
you can use loop instead of a function

src/formatDate.js
```diff
@@ -51,6 +51,43 @@
...
+      }
+
+      const isBothYears = oldElement.includes('Y')
+      && toFormat[index].includes('Y');
+      const isNewLengthBigger = oldElement.length < toFormat[index].length;
+
+      if (isBothYears && !isNewLengthBigger) {
+        newElement = oldDate[indexOfOld].slice(2);
+        newDate.push(newElement);
+      }
```
oleh-holovnykh, line 74: oldDate[indexOfOld] already in newElement, perhaps it will be more clear to change newElement

`newElement = newElement.slice(2)` or `newDate.push(newElement.slice(2))`

oleh-holovnykh (changes requested):
Good job! Interesting solution with function and years definition. But I think some points can be simplified

PR #1017, opened 2023-01-09
Finish task

U-Ra-All:


src/formatDate.js
```diff
@@ -51,6 +51,32 @@
...
+          = Number(dateArray[i]) < 30 ? 20 + dateArray[i] : 19 + dateArray[i];
+      }
+
+      if (fromFormat[i].length === 4) {
+        dateObject['YY'] = dateArray[i].slice(2);
+      }
+    }
+
+    newDate += dateObject[toFormat[i]] + toFormat[3];
+  }
```
IvanFesenko, line 77: fromFormat[i] - is repeating several times. You can create seperate variable for it, for example const oldFormat = fromFormat[i] and than use it.
Better check on string contain instead length
if (fromFormat[i] === 'YYYY')

src/formatDate.js
```diff
@@ -51,6 +51,32 @@
...
+    dateObject[fromFormat[i]] = dateArray[i];
+  }
+
+  for (let i = 0; i < 3; i++) {
+    const index = toFormat.indexOf(fromFormat[i]);
+
+    if (index === -1) {
+      if (fromFormat[i].length === 2) {
+        dateObject['YYYY']
+          = Number(dateArray[i]) < 30 ? 20 + dateArray[i] : 19 + dateArray[i];
```
IvanFesenko, line 68: Also fix ternary operator style

```suggestion
        dateObject['YYYY']  = Number(dateArray[i]) < 30 
          ? 20 + dateArray[i] 
          : 19 + dateArray[i];
```

IvanFesenko (changes requested):

src/formatDate.js
U-Ra-All: Thanks, done!

src/formatDate.js
```diff
@@ -51,6 +51,34 @@
...
+    dateObject[fromFormat[i]] = dateArray[i];
+  }
+
+  for (let i = 0; i < 3; i++) {
+    const oldFormat = fromFormat[i];
+    const index = toFormat.indexOf(oldFormat);
+
+    if (index === -1) {
+      if (oldFormat === 'YY') {
+        dateObject['YYYY'] = Number(dateArray[i]) < 30
```
OlEzhA0, line 68:
```suggestion
        dateObject['YYYY'] = +dateArray[i] < 30 
```

src/formatDate.js
```diff
@@ -51,6 +51,34 @@
...
+
+      if (oldFormat === 'YYYY') {
+        dateObject['YY'] = dateArray[i].slice(2);
+      }
+    }
+
+    newDate += dateObject[toFormat[i]] + toFormat[3];
+  }
+
+  return newDate.slice(0, -1);
```
OlEzhA0, line 81: why do we need to add `toFormat[3]` each times?
you can define newDate like array, push formatted element, and return joined array with separator

OlEzhA0 (approved):
Hey! Good job
I approve your work but check please my suggestions

src/formatDate.js
U-Ra-All: Thanks, done!

PR #1018, opened 2023-01-09
Solution

AV1788:


loralevitska (changes requested):
Your solution is too big. Try to optimize it
You already has 4 element in array. The fourth is always separator and second always month. Date format can be only from 2 to 4 length)

PR #1019, opened 2023-01-09
DateFormate function was made

SamirMamedov1010:


src/formatDate.js
```diff
@@ -50,7 +50,46 @@
...
+  const years = 'YYYY';
+  const shortYears = 'YY';
+
+  const daysFromIndex = fromFormat.indexOf(days);
+  const monthsFromIndex = fromFormat.indexOf(months);
+  let yearsFromIndex = fromFormat.indexOf(years);
+
+  const daysToIndex = toFormat.indexOf(days);
+  const monthsToIndex = toFormat.indexOf(months);
+  let yearsToIndex = toFormat.indexOf(years);
```
loralevitska, line 64: It can be simplify. Divider index is always 3. Month index is always 1. So you need only to find daysIndex, and yearsIndex will be last of 4 elements automatically

loralevitska (changes requested):
Almost done) Let's improve

src/formatDate.js
SamirMamedov1010: Hi, thanks for check) But tests failed when month index is not one. And also i don't know how to find last element index(of year) can you help me with this?
Attach screen with failed test below.)
Знімок екрана 2023-01-12 о 13 09 30

src/formatDate.js
```diff
@@ -50,7 +50,46 @@
...
+  const splitedDateArr = date.split(fromFormat[3]);
+  const newArr = Array(3);
+
+  newArr[daysToIndex] = splitedDateArr[daysFromIndex];
+  newArr[monthsToIndex] = splitedDateArr[monthsFromIndex];
+
+  newArr[yearsToIndex] = splitedDateArr[yearsFromIndex];
+
+  if (fromFormat.includes(years)
+  && toFormat.includes(shortYears)) {
```
DarkMistyRoom, line 75:
```suggestion
  if (fromFormat.includes(years)
    && toFormat.includes(shortYears)) {
```

DarkMistyRoom (approved):

PR #1020, opened 2023-01-09
resolve

Bir3791:


src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
+  const rezult = [];
+
+  const oldData = date.split(oldSpliter);
+
+  for (let i = 0; i < 3; i++) {
+    time[fromFormat[i]] = oldData[i];
+  }
+
+  if (time.YY) {
+    if (time.YY < 30) {
```
loralevitska, line 65: What does it mean 30, 20, 19?  Looks like magic numbers. Move it to variables with meaningful names

loralevitska (changes requested):
Almost done) Let's improve

Virass (approved):

PR #1024, opened 2023-01-10
solution

vasyliev-anton:


src/formatDate.js
```diff
@@ -50,7 +50,44 @@
...
+    }
+
+    if (element === 'YYYY') {
+      result.push(year);
+    }
+
+    if (element === 'YY') {
+      result.push(year.slice(-2));
+    }
+  });
```
SanyaBratashchuk, line 88: Do not use map if you don't need array returned  by map

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #1027, opened 2023-01-13
Task function formatDate

ilona-romaniuk:


src/formatDate.js
```diff
@@ -50,7 +50,55 @@
...
+      day = i;
+    };
+
+    if (toFormat[i] === 'MM') {
+      month = i;
+    }
+
+    if (toFormat[i] === 'YY' || toFormat[i] === 'YYYY') {
+      year = i;
+    }
```
loralevitska, line 72: Use [`switch case`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) in cases where you constantly compare one value with others. Fix it everywhere

Anastasiia-Svintsova (approved):

PR #1034, opened 2023-01-20
Solution

pushkarskiyrodion:


src/formatDate.js
```diff
@@ -50,7 +50,38 @@
...
+      if (toFormat[k].includes(fromFormat[i][0])) {
+        formattedDate[toFormat[k]] = splitted[i];
+      }
+    }
+  }
+
+  if (formattedDate.YYYY && formattedDate[year].length === 2) {
+    formattedDate[year] = formattedDate[year] < 30
+      ? 20 + formattedDate[year]
+      : 19 + formattedDate[year];
```
Volodymyr-Melnyk-Fron10der, line 69: move formattedDate[year] into const variable and use it instead of formattedDate[year], because there are a lot of duplicates

src/formatDate.js
```diff
@@ -50,7 +50,38 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const splitted = date.split(fromFormat[3]);
+  const year = toFormat.includes('YY') ? 'YY' : 'YYYY';
```
Volodymyr-Melnyk-Fron10der, line 54: I would name `toYearKey ` instead of just `year `. It's not so understandable

src/formatDate.js
```diff
@@ -50,7 +50,38 @@
...
+    formattedDate[year] = formattedDate[year].slice(-2);
+  }
+
+  for (const prop in formattedDate) {
+    for (let i = 0; toFormat.length - 1 > i; i++) {
+      if (prop === toFormat[i]) {
+        result[i] = formattedDate[prop];
+      }
+    }
+  }
```
Volodymyr-Melnyk-Fron10der, line 82: It`s better to use  Object.entries + forEach

```
Object.entries(formattedDate).forEach(([key, value]) => {
  for (let i = 0; toFormat.length - 1 > i; i++) {
      if (key=== toFormat[i]) {
        result[i] = value;
      }
    }
})
```

src/formatDate.js
```diff
@@ -50,7 +50,38 @@
...
+  const formattedDate = {};
+  const result = [];
+
+  for (let i = 0; fromFormat.length - 1 > i; i++) {
+    for (let k = 0; toFormat.length - 1 > k; k++) {
+      if (toFormat[k].includes(fromFormat[i][0])) {
+        formattedDate[toFormat[k]] = splitted[i];
+      }
+    }
+  }
```
Volodymyr-Melnyk-Fron10der, line 64: Try to use `Array.forEach()` method

Volodymyr-Melnyk-Fron10der (changes requested):
Almost there! Just few fixes are left

Volodymyr-Melnyk-Fron10der (approved):
Much better, good job!

PR #1042, opened 2023-01-24
task done

tecHamster:


src/formatDate.js
```diff
@@ -50,7 +50,44 @@
...
+    return fromFormat.indexOf(value);
+  }
+
+  const newDate = [];
+  const oldDate = date.split(fromFormat[3]);
+
+  for (const item of toFormat) {
+    switch (item) {
+      case ('DD'):
+        newDate.push(oldDate[getIndex('DD')]);
```
Anastasiia-Svintsova, line 63: You make `newDate.push` each time. Maybe do it only once at the end of the function.

Anastasiia-Svintsova (changes requested):
Good! Check my comment

Anastasiia-Svintsova (approved):

PR #1045, opened 2023-01-25
Task done

andriy-litkovets:


src/formatDate.js
```diff
@@ -50,7 +50,47 @@
...
+      day = arrayOfDate[i];
+    }
+  }
+
+  for (const format of toFormat) {
+    if (format.includes('Y')) {
+      if (format.length < year.length) {
+        resultDate.push(year.slice(2));
+      } else {
+        resultDate.push(year.padStart(format.length, year >= 30 ? '19' : '20'));
```
agniya-i, line 80: you can use ternary operator here for value that you push to array

agniya-i (approved):

PR #1047, opened 2023-01-27
Solution

fisherok3452:


src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
+
+  if (fromFormat.includes('YY') && newDate[indexYearShort] >= 30) {
+    newDate[indexYearShort] = '19' + newDate[indexYearShort];
+  }
+
+  if (fromFormat.includes('YY') && newDate[indexYearShort] < 30) {
+    newDate[indexYearShort] = '20' + newDate[indexYearShort];
+  }
+
+  fromFormat[indexYearShort] = 'YYYY';
```
VitaliyBondarenko1982, line 65: Better to make constants for all hardcoded strings  and numbers outside of function.
Esprcially which uses more than one time.

fisherok3452: I didn't understand. Should I move all of variables outside of function? or whats wrong with code on lines 62-65?

VitaliyBondarenko1982 (approved):

PR #1048, opened 2023-01-27
task was solved

vanyachyzh:


src/formatDate.js
```diff
@@ -43,14 +43,53 @@
...
- * @param {string[]} fromFormat
+ * @param {string[]} string
  * @param {string[]} toFormat
  *
  * @returns {string}
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const string = fromFormat.join('.');
```
witflash, line 53: Why do you need to join the string? You can check by indexOf inside the original array.

src/formatDate.js
```diff
@@ -43,14 +43,53 @@
...
+  const separator = toFormat[3];
+  const newFormat = [];
+
+  if (fromFormat.includes('YYYY')) {
+    year = date.slice(string.indexOf('YYYY'), string.indexOf('YYYY') + 4);
+  } else {
+    year = date.slice(string.indexOf('YY'), string.indexOf('YY') + 2);
+  }
+
+  for (let i = 0; i < toFormat.length; i++) {
```
witflash, line 67: You don't need to use a cycle for this task.
You can check indexOf year, day and month inside `fromFormat`, get value by this index from `date` and place by index from `toFormat`.

witflash (changes requested):
Check my comments, please

src/formatDate.js
```diff
@@ -43,14 +43,52 @@
...
+
+    return nextDate.join(toFormat[3]);
+  }
+
+  if (!isDouble1) {
+    nextDate[indYear2] = year[2] + year[3];
+  }
+
+  if (!isDouble2) {
+    nextDate[indYear2] = year < 30 ? '20' + year : '19' + year;
```
VitaliyBondarenko1982, line 88: Good job!

Please create constants for all magic numbers and hardcoded strings outside of function in next tasks.

VitaliyBondarenko1982 (approved):

PR #1049, opened 2023-01-28
the function was created

YevheniiHura:


src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+          result.push(year);
+        } else if (toFormat[k] === 'YYYY' && year.length === 2 && year < 30) {
+          result.push(`20${year}`);
+        } else if (toFormat[k] === 'YYYY' && year.length === 2 && year >= 30) {
+          result.push(`19${year}`);
+        } else if (toFormat[k] === 'YY' && year.length === 4) {
+          result.push(year.slice(2));
+        } else if (toFormat[k] === 'YY' && year.length === 2) {
+          result.push(year);
+        }
```
Bohdan-Maksymiuk, line 96: avoid such nesting `else if` conditions
Such code is hard to read and maintain

Bohdan-Maksymiuk: also will be good to split conditions into separate variables

Bohdan-Maksymiuk (approved):

PR #1050, opened 2023-01-29
Solution

Oleg-Lukanov:


src/formatDate.js
```diff
@@ -50,7 +50,34 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateArr = date.split(fromFormat[3]);
```
YevhS, line 53: Use destructuring to extract values from dateArr for better readability

YevhS (approved):

PR #1052, opened 2023-01-30
The tests were completed successfully

read-my-code:


src/formatDate.js
```diff
@@ -50,7 +50,63 @@
...
+  return finallyDate.join(toFormat[3]);
+}
+
+function formatYear(year) {
+  if (+year < 30) {
+    return `20${year}`;
+  } else {
+    return `19${year}`;
+  }
 }
```
loralevitska, line 110:
```suggestion
function formatYear(year) {
  if (+year < 30) {
    return `20${year}`;
  }
  
  return `19${year}`;
}
```

src/formatDate.js
```diff
@@ -50,7 +50,63 @@
...
+        day = normalizeDate[i];
+        break;
+
+      case 'MM' :
+        month = normalizeDate[i];
+        break;
+    }
+  }
+
+  normalizeDate = [day, month, year];
```
loralevitska, line 79: Could you explain for what this rewriting if you don't use this variable after 79 line?

loralevitska (changes requested):
Good Job! Let's improve
1. Use `switch` with `default`
2. Pay attention to comments

If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

src/formatDate.js
read-my-code: А й правда юзлес. Не замітив.

SanyaBratashchuk (approved):

PR #1053, opened 2023-01-31
Remaked data done

oleksandr-plias:


src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparator = fromFormat[3];
+  const toSeparator = toFormat[3];
+  const originalDate = date.split(fromSeparator);
+
+  const newDate = [];
+  const dateObj = {
+
+  };
```
loralevitska, line 60:
```suggestion
  const dateObj = {};
```

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparator = fromFormat[3];
+  const toSeparator = toFormat[3];
+  const originalDate = date.split(fromSeparator);
+
+  const newDate = [];
```
loralevitska, line 57: Delete an extra empty line in blocks with the same logic
```suggestion
  const originalDate = date.split(fromSeparator);
  const newDate = [];
```

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+  }
+
+  for (let i = 0; i < 3; i++) {
+    newDate[i] = dateObj[toFormat[i]];
+
+    if (toFormat[i] === 'YY') {
+      newDate[i] = dateObj.YYYY.slice(2);
+    } else if (toFormat[i] === 'YYYY') {
+      newDate[i] = dateObj.YYYY;
+    }
```
loralevitska, line 81: You can use `toFormat[i].length` and simplify expresion to avoid code duplicate
```suggestion
    if (toFormat[i] === 'YY') {
      newDate[i] = dateObj.YYYY.slice(2);
    } else if (toFormat[i] === 'YYYY') {
      newDate[i] = dateObj.YYYY;
    }
```

loralevitska (changes requested):
Almost done) Let's improve

marina-tilniak (approved):

PR #1055, opened 2023-01-31
Solution

nadia-hope:


src/formatDate.js
```diff
@@ -50,7 +50,46 @@
...
+  const newDate = [];
+  const dateInfo = {};
+
+  for (let i = 0; i < fromFormat.length; i++) {
+    if (fromFormat[i] === 'YY') {
+      if (oldDate[i] < 30) {
+        dateInfo.YYYY = `20${oldDate[i]}`;
+      } else {
+        dateInfo.YYYY = `19${oldDate[i]}`;
+      }
```
polosanya, line 64: You can use ternary operator here instead

polosanya (approved):
Good job!

PR #1059, opened 2023-02-02
sollution

serbinm:
I was so angry with this task, so that I just wanted to finish it already. I think the code is quite messy, but this is by far the best working sollution I could present. Ready for any critics :D

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -50,7 +50,59 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const data = date.split(fromFormat[3]);
+  let fourNumbers = true;
+
+  let year;
+  let month;
+  let day;
+  let separator;
+
+  const answer = [];
+  let finalFormat = 0;
+
+  for (const time of fromFormat) {
+    if (time === 'YY') {
+      year = data[fromFormat.indexOf(time)];
+      fourNumbers = false;
+    } else if (time === 'YYYY') {
+      year = data[fromFormat.indexOf(time)];
+      fourNumbers = true;
+    } else if (time === 'MM') {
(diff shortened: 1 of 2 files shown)


src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+      year = data[fromFormat.indexOf(time)];
+      fourNumbers = false;
+    } else if (time === 'YYYY') {
+      year = data[fromFormat.indexOf(time)];
+      fourNumbers = true;
+    } else if (time === 'MM') {
+      month = data[fromFormat.indexOf(time)];
+    } else if (time === 'DD') {
+      day = data[fromFormat.indexOf(time)];
+    }
```
Anastasiia-Svintsova, line 75: Use switch/case, please

src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+      }
+
+      answer.push(year);
+    } else if (time === 'MM') {
+      answer.push(month);
+    } else if (time === 'DD') {
+      answer.push(day);
+    } else {
+      separator = time;
+    }
```
Anastasiia-Svintsova, line 100: Same here 
Try to avoid nesting

src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+  let month;
+  let day;
+  let separator;
+
+  const answer = [];
+  let finalFormat = 0;
+
+  for (const time of fromFormat) {
+    if (time === 'YY') {
+      year = data[fromFormat.indexOf(time)];
```
Anastasiia-Svintsova, line 66: Could you move `data[fromFormat.indexOf(time)]` to a variable? You use it no once

Anastasiia-Svintsova (changes requested):

PR #1060, opened 2023-02-03
Src was added

VadymTsyndra:


src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    dateObj[fromFormat[i]] = dateSplit[i];
+  }
+
+  if (dateObj.YYYY === '' && dateObj.YY >= 30) {
+    dateObj.YYYY = 19 + dateObj.YY;
+  }
+
+  if (dateObj.YYYY === '' && dateObj.YY < 30) {
```
loralevitska, line 72: Also you can combine your `if` in helper function. This expression `dateObj.YYYY === ''`  is repeated

loralevitska (changes requested):
Keep up the good work! Some comments to fix:

SanyaBratashchuk (changes requested):
There is no changes since last review. If you have any questions feel free to ask in Slack

src/formatDate.js
DarkMistyRoom: that means that you can create cutYear function and use it in this way:
`if (dateObj.YYY) {
  cutYear(dateObj);
}`

or something like it, to avoid repeating

DarkMistyRoom (approved):

PR #1061, opened 2023-02-03
Solution

AnastasiiaEm:


etojeDenys (approved):
LGTM! 👍🏻

also, you can practice and try to rewrite it using `Array.reduce` method

src/formatDate.js
```diff
@@ -50,7 +50,35 @@
...
+
+  for (let i = 0; i < fromFormat.length; i++) {
+    if (fromFormat[i] === 'YY') {
+      obj['YYYY'] = dateArr[i] < 30 ? `20${dateArr[i]}` : `19${dateArr[i]}`;
+    }
+    obj[fromFormat[i]] = dateArr[i];
+  }
+
+  for (const newFormat of toFormat) {
+    switch (newFormat) {
```
witflash, line 65: Why do you need switch/case here?
You can do this more simpler:
```
if (newFormat === 'YY') {
  // do your action
} else {
  // do action for all other cases
}
```

witflash (changes requested):
Good work!
But try to do small improvements to your work.

IvanFesenko (approved):
Good job

PR #1063, opened 2023-02-05
Solution

vlad-ptashnichenko:


src/formatDate.js
```diff
@@ -50,7 +50,63 @@
...
+  const interimObject = {};
+  const lastArray = [];
+  const dateArray = date.split(fromFormat[3]);
+
+  for (let iFrom = 0; iFrom < fromFormat.length - 1; iFrom++) {
+    for (let iTo = 0; iTo < toFormat.length - 1; iTo++) {
+      if (fromFormat[iFrom] === toFormat[iTo] && fromFormat[iFrom] === 'DD') {
+        DDfrom = iFrom;
+        DDto = iTo;
+      }
```
witflash, line 67: It's workable solution but too complex.
You don't need to use any cycles. Instead of them you can simple use `.indexOf()` method for fromFormatArray and toFormatArray.

src/formatDate.js
```diff
@@ -50,7 +50,63 @@
...
+    }
+
+    if (i === '1') {
+      lastArray.push(interimObject[i]);
+    }
+
+    if (i === '2') {
+      lastArray.push(interimObject[i]);
+    }
+  }
```
witflash, line 107: All this code just realization of `Object.values(interimObject)`.

But you don't really need even this. You can prepare date array directly instead of additional object usage.

witflash (changes requested):
Your solution is workable, but could you do some code improvements, please?

VitaliyBondarenko1982 (approved):
Approve.

Just one point.
Try to make constants for 'magic numbers' and 'hardcoded strings' in your code

PR #1067, opened 2023-02-06
FormatDataAdvanced

RomanHrytsiuk:


src/formatDate.js
```diff
@@ -50,7 +50,52 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromFormatSeparator = fromFormat[fromFormat.length - 1];
```
loralevitska, line 53: Use [at()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at)
```suggestion
  const fromFormatSeparator = fromFormat.at(-1);
```

src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
+    }
+  }
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    if (toFormat[i] === 'YY' && year.length === 4) {
+      result.push(year.slice(-2));
+    } else if (toFormat[i] === 'YYYY' && year < 30 && year.length === 2) {
+      result.push(20 + year);
+    } else if (toFormat[i] === 'YYYY' && year >= 30 && year.length === 2) {
+      result.push(19 + year);
```
loralevitska, line 84: In this condition many repeated expressions. Use helper function

loralevitska (changes requested):
Good job! Let's improve

src/formatDate.js
RomanHrytsiuk: at() doesn't work. It shows me an error: "TypeError: fromFormat.at is not a function"

slavanosatskyi (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,58 @@
...
+      day = fromFormat[i];
+    }
+  }
+
+  const defineYear = () => {
+    if (year < 30) {
+      return result.push(20 + year);
+    }
+
+    return result.push(19 + year);
```
DarkMistyRoom, line 83: no need to use return in this case

PR #1076, opened 2023-02-12
js_format_date_advanced

Mykola-Borodavka:


src/formatDate.js
```diff
@@ -50,7 +50,47 @@
...
+  let result = '';
+
+  for (let i = 0; i < 3; i++) {
+    let dateName = fromFormat[i];
+    let dateValue = dateValues[i];
+
+    if (dateName === 'YY') {
+      dateName = 'YYYY';
+
+      if (+dateValue < 30) {
```
Anastasiia-Svintsova, line 66: It's better not to use mysterious numbers. would be better to move them to variables (Like 30, 20, 19)

Anastasiia-Svintsova (approved):

PR #1080, opened 2023-02-14
solution

kuznetsova-anastasiia:


src/formatDate.js
```diff
@@ -50,7 +50,74 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let year;
+  let month;
+  let day;
+  let separator = fromFormat[3];
+  const fromArray = date.split(separator);
```
KuharchukDmytro, line 57: on my opninion, there is no need to declare separator
```suggestion
  const fromArray = date.split(fromFormat[3]);
```

src/formatDate.js
```diff
@@ -50,7 +50,74 @@
...
+        break;
+
+      default:
+        break;
+    }
+  }
+
+  separator = toFormat[3];
+
+  return resultArray.join(separator);
```
KuharchukDmytro, line 120: and here you can directly write separator in join
```suggestion
  return resultArray.join(toFormat[3]);
```

KuharchukDmytro (approved):
good job! only one moment with separator, but its only my opinion:)

src/formatDate.js
kuznetsova-anastasiia: yeah, it seems so. but it's more understandable for me using a separate variable for it)) thanks!

src/formatDate.js
```diff
@@ -50,7 +50,74 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let year;
+  let month;
+  let day;
+  let separator = fromFormat[3];
+  const fromArray = date.split(separator);
+  const resultArray = [];
+
+  for (let i = 0; i < 3; i++) {
```
illiarerun, line 60: what is `3`, is this some kind of magic number?

src/formatDate.js
```diff
@@ -50,7 +50,74 @@
...
+      case 'YYYY':
+        year = current;
+        break;
+
+      default:
+        break;
+    }
+  }
+
+  for (let i = 0; i < 3; i++) {
```
illiarerun, line 85: same here

src/formatDate.js
```diff
@@ -50,7 +50,74 @@
...
+        if (year.length === 2) {
+          if (+year < 30) {
+            year = '20' + year;
+          } else {
+            year = '19' + year;
+          }
+        }
+        resultArray.push(year);
+        break;
+
```
illiarerun, line 112: you do the same push on each step, can you push it after switch/case ends?

src/formatDate.js
```diff
@@ -50,7 +50,74 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let year;
+  let month;
+  let day;
+  let separator = fromFormat[3];
```
illiarerun, line 56: I don't like usage of `let` for it. you can create separate `toSeparator` and `fromSeparator` variables

illiarerun (changes requested):

src/formatDate.js
kuznetsova-anastasiia: > you do the same push on each step, can you push it after switch/case ends?

I think i can't... i do push in each step for the right consistency of date elements. i iterate toFormat. if on zero index in toFormat it's YY, then case 'YY' will work and will push the year on the right place in result array. if i push after all cases i don't know were have to be year, month and day. i need to add more if conditions or something like that. But i can replace push method in each case with `resultArray[i] = day / month/ year`, but it will be the same logic. Maybe there is another solution for this but now i can't remember anything, i`m sorry. maybe you can suggest something 😸 @illiarerun

illiarerun (approved):

PR #1081, opened 2023-02-14
Solution

stanislavcodes:


timoxa2t (approved):
Good work. Code is a little hard to read, but there is no guideline, and it works well)

src/formatDate.js
```diff
@@ -50,7 +50,30 @@
...
+
+      if (fromFormat[i] === 'YY') {
+        dateObject.YYYY = Number(yearSliced) < 30
+          ? `20${yearSliced}`
+          : `19${yearSliced}`;
+      } else {
+        dateObject.YY = yearSliced;
+      }
+    }
+    dateObject[fromFormat[i]] = dateSplit[i];
```
illiarerun, line 69: seems like you'll have both `YY` and `YYYY` in resulting object

src/formatDate.js
```diff
@@ -50,7 +50,30 @@
...
+  }
+
+  for (const datePart of toFormat.slice(0, 3)) {
+    dateInNewFormatArray.push(dateObject[datePart]);
+  }
+
+  return dateInNewFormatArray.join(toFormat[3]);
```
illiarerun, line 76: you can improve your solution a bit if you use `pop` on `toFormat` copy

illiarerun (approved):

PR #1082, opened 2023-02-14
tests done

Kocha1991:


DarkMistyRoom (changes requested):
it is good if it passes all tests, but I have to break the head to figure out this solution) 
let's try another approach: use the object where the keys are format("DD", "YY", ...), and values are appropriate values of a given date.

DarkMistyRoom (changes requested):
It is much better. But you can use only one object which can have a "YY" field as well as "YYYY". You will need just to compare them. It makes the code shorter and simpler. Let's improve

src/formatDate.js
```diff
@@ -42,6 +42,50 @@
...
+  //     objNewFormat.YYYY = `19${objOlldFormat.YY}`;
+  //   }
+  // }
+
+  // for (let i = 0; i < toFormat.length - 1; i++) {
+  //   result.push(objNewFormat[toFormat[i]]);
+  // }
+
+  // console.log(objOlldFormat);
+  // console.log(objNewFormat);
```
nazarmatsevych, line 87: Remove comments

nazarmatsevych (changes requested):
Looks good,  just remove comments

src/index.html
```diff
@@ -0,0 +1,13 @@
...
+  <meta charset="UTF-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Document</title>
+  <script src="formatDate.js" defer></script>
+</head>
+<body>
+  !!!!
+</body>
+</html>
```
loralevitska, line 13: Always add [empty line in the end of file](https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file)

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
+  if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
+    tempDate['YY'] = tempDate['YYYY'].slice(-2);
+  }
+
+  if (fromFormat.includes('YY') && toFormat.includes('YYYY')) {
+    if (tempDate.YY < 30) {
+      tempDate.YYYY = `20${tempDate.YY}`;
+    } else {
+      tempDate.YYYY = `19${tempDate.YY}`;
+    }
```
loralevitska, line 73: Use ternary operator here

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  // SECOND METOD
```
loralevitska, line 53: Remove comments

loralevitska (approved):

DarkMistyRoom (approved):
Great! Now I like it)

PR #1085, opened 2023-02-15
Solution

Vasyl-Pauk:


slavanosatskyi (approved):
Nice, just one suggestion for the future - don't forget about default cases in switch case statements. You never know which data you will start to receive in future

PR #1088, opened 2023-02-16
Format Date Advanced v.0.1 - Solution

NedobezhkinOleksandr:


src/formatDate.js
```diff
@@ -50,7 +50,58 @@
...
+  const toFormatSeparator = toFormat[toFormat.length - 1];
+  const dateArray = date.split(fmFormatSeparator);
+
+  const changedDate = [];
+
+  let year, month, day;
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    if (fromFormat[i] === 'YY' || fromFormat[i] === 'YYYY') {
+      fromFormat[i] = dateArray[i];
```
Volodymyr-Melnyk-Fron10der, line 63: 1. Sorry, don't understand the reason to do it. why?
2. Write pure function, don't mutate initial values comes as arguments

src/formatDate.js
```diff
@@ -50,7 +50,58 @@
...
+    }
+
+    return changedDate.push(19 + year);
+  };
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    if (toFormat[i] === 'YY' && year.length === 4) {
+      changedDate.push(year.slice(-2));
+    } else if (toFormat[i] === 'YYYY' && year.length === 2) {
+      makeYear(year);
```
Volodymyr-Melnyk-Fron10der, line 90: You use function only once, so maybe it's better just write a function body right here

Volodymyr-Melnyk-Fron10der (changes requested):
Keep it up👍 Check my comments please

src/formatDate.js
NedobezhkinOleksandr: I agree, it was useless, solved.

src/formatDate.js
NedobezhkinOleksandr: Done

vlad-demchuk (approved):

PR #1089, opened 2023-02-16
solution

KuharchukDmytro:


Yekatery (approved):
Good Job!

src/formatDate.js
```diff
@@ -50,7 +50,45 @@
...
+      result.push(dateParts['MM']);
+      continue;
+    }
+
+    if (!dateParts[toFormat[i]]) {
+      if (dateParts['YY']) {
+        result.push(
+          dateParts['YY'] < 30
+            ? '20' + dateParts['YY']
+            : '19' + dateParts['YY']
```
sergii-nosachenko, line 79: Use ternary for the prefix only. Don't repeat `dateParts['YY']` twice.

src/formatDate.js
```diff
@@ -50,7 +50,45 @@
...
+        );
+        continue;
+      }
+
+      result.push(dateParts['YYYY'].toString().slice(2, 4));
```
sergii-nosachenko, line 84:
```suggestion
      result.push(dateParts['YYYY'].toString().slice(-2));
```

src/formatDate.js
```diff
@@ -50,7 +50,45 @@
...
+      result.push(dateParts['DD']);
+      continue;
+    }
+
+    if (toFormat[i] === 'MM') {
+      result.push(dateParts['MM']);
+      continue;
+    }
+
+    if (!dateParts[toFormat[i]]) {
```
sergii-nosachenko, line 74: It looks hard to maintain. Better use `switch case` instead of multiple ifs.

sergii-nosachenko (changes requested):

src/formatDate.js
```diff
@@ -61,28 +61,27 @@ function formatDate(date, fromFormat, toFormat) {
...
-      result.push(dateParts['DD']);
-      continue;
-    }
-
-    if (toFormat[i] === 'MM') {
-      result.push(dateParts['MM']);
-      continue;
-    }
+    switch (true) {
+      case toFormat[i] === 'DD':
```
sergii-nosachenko, line 65:
```suggestion
    switch (toFormat[i]) {
      case 'DD':
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

src/formatDate.js
```diff
@@ -61,28 +61,27 @@ function formatDate(date, fromFormat, toFormat) {
...
+        result.push(dateParts['DD']);
+        continue;
```
sergii-nosachenko, line 67:
```suggestion
        break;
```

src/formatDate.js
```diff
@@ -61,28 +61,27 @@ function formatDate(date, fromFormat, toFormat) {
...
+            (dateParts['YY'] < 30
+              ? '20'
+              : '19') + dateParts['YY']
+          );
+          continue;
+        }
+
+        result.push(dateParts['YYYY'].toString().slice(-2));
+        continue;
     }
```
sergii-nosachenko, line 85: Add default case

sergii-nosachenko (changes requested):

src/formatDate.js
KuharchukDmytro: unfortunally i cant change it, because then my third case brakes
maybe you can give me a hint how i can do this?

sergii-nosachenko: Add two more cases: 'YY' and 'YYYY':
```js
case 1:
case 2:
  // your code for both
```

src/formatDate.js
```diff
@@ -50,7 +50,45 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateParts = {};
+  const datePartArr = date.split(fromFormat[3]);
+  const result = [];
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    if (!dateParts[fromFormat[i]]) {
```
sergii-nosachenko, line 58: Why do you need this condition?

sergii-nosachenko (approved):
Good work 👍

PR #1092, opened 2023-02-17
Solution

morozit:


src/formatDate.js
```diff
@@ -50,7 +50,56 @@
...
+          pushToArr(oldDateObj['YY']);
+        }
+        break;
+
+      case 'MM':
+        pushToArr(oldDateObj['MM'].toString().padStart(2, '0'));
+        break;
+
+      case 'DD':
+        pushToArr(oldDateObj['DD'].toString().padStart(2, '0'));
```
agniya-i, line 94: function `pushToArr` looks redundant to me, you can also create a variable for temporary result and push value to the array after switch/case

agniya-i (approved):
Good job, I approved, but have small comment

PR #1096, opened 2023-02-17
:white_check_mark: implement solution for formatDate kata

Bogdan-Mykhailov:
implement solution for formatDate kata

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -50,7 +50,35 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const resultedDate = [];
+  const separatorFrom = fromFormat[3];
+  const separatorTo = toFormat[3];
+
+  const separatedDate = date.split((separatorFrom));
+  let currentYear = separatedDate[fromFormat.indexOf('YY')]
+    || separatedDate[fromFormat.indexOf('YYYY')];
+
+  resultedDate[toFormat.indexOf('MM')]
+    = separatedDate[fromFormat.indexOf('MM')];
+
+  resultedDate[toFormat.indexOf('DD')]
+    = separatedDate[fromFormat.indexOf('DD')];
+
+  if (toFormat.includes('YYYY')) {
+    if (fromFormat.includes('YY')) {
+      currentYear = currentYear < 30 ? `20${currentYear}` : `19${currentYear}`;
+    }
+    resultedDate[toFormat.indexOf('YYYY')] = currentYear;
+  }
+
+  if (toFormat.includes('YY')) {
+    if (fromFormat.includes('YYYY')) {
+      currentYear = currentYear.slice(-2);
+    }
+    resultedDate[toFormat.indexOf('YY')] = currentYear;
+  }
+
+  return resultedDate.join(separatorTo);
 }
 
 module.exports = formatDate;


Zelinskyi-Serhii (approved):
Виглядає дивно але цікаве рішення.

src/formatDate.js
```diff
@@ -50,7 +50,35 @@
...
+
+  resultedDate[toFormat.indexOf('MM')]
+    = separatedDate[fromFormat.indexOf('MM')];
+
+  resultedDate[toFormat.indexOf('DD')]
+    = separatedDate[fromFormat.indexOf('DD')];
+
+  if (toFormat.includes('YYYY')) {
+    if (fromFormat.includes('YY')) {
+      currentYear = currentYear < 30 ? `20${currentYear}` : `19${currentYear}`;
```
sergii-nosachenko, line 69: Use ternary for the prefix only. Don't repeat `currentYear` twice.

src/formatDate.js
```diff
@@ -50,7 +50,35 @@
...
+    }
+    resultedDate[toFormat.indexOf('YYYY')] = currentYear;
+  }
+
+  if (toFormat.includes('YY')) {
+    if (fromFormat.includes('YYYY')) {
+      currentYear = currentYear.slice(-2);
+    }
+    resultedDate[toFormat.indexOf('YY')] = currentYear;
+  }
```
sergii-nosachenko, line 79: Looks bad. It's hard to read and maintain such a code style. Try to simplify and refactor.

src/formatDate.js
```diff
@@ -50,7 +50,35 @@
...
+
+  const separatedDate = date.split((separatorFrom));
+  let currentYear = separatedDate[fromFormat.indexOf('YY')]
+    || separatedDate[fromFormat.indexOf('YYYY')];
+
+  resultedDate[toFormat.indexOf('MM')]
+    = separatedDate[fromFormat.indexOf('MM')];
+
+  resultedDate[toFormat.indexOf('DD')]
+    = separatedDate[fromFormat.indexOf('DD')];
```
sergii-nosachenko, line 65: Better do it in the right order from the beginning using loop.

src/formatDate.js
```diff
@@ -50,7 +50,35 @@
...
+
+  if (toFormat.includes('YYYY')) {
```
sergii-nosachenko, line 67: `toFormat.includes('YYYY')` can be replaced by comparing the result of `toFormat.indexOf('YYYY')` with -1. So you willn't call an array loop twice (these methods are loops under the hood so are expensive operations). Same for other use cases.

sergii-nosachenko (changes requested):

src/formatDate.js
Bogdan-Mykhailov: 👍🏼

src/formatDate.js
Bogdan-Mykhailov: done

Ambulence (approved):

src/formatDate.js
```diff
@@ -50,7 +50,48 @@
...
+    const fromIndex = fromFormat.indexOf(component);
+    const toIndex = toFormat.indexOf(component);
+
+    if (fromIndex !== -1 && toIndex !== -1) {
+      resultedDate[toIndex] = separatedDate[fromIndex];
+    }
+  });
+
+  const indexOfYYYY = toFormat.indexOf('YYYY');
+  const indexOfYY = toFormat.indexOf('YY');
```
illiarerun, line 72: you can also replace your `includes` in bottom with it

src/formatDate.js
```diff
@@ -50,7 +50,48 @@
...
+
+  if (indexOfYYYY !== -1) {
+    const prefix = fromFormat.includes('YY')
+      ? (currentYear < 30 ? '20' : '19')
```
illiarerun, line 76: don't create nested ternaries

src/formatDate.js
```diff
@@ -50,7 +50,48 @@
...
+      : '';
+
+    resultedDate[indexOfYYYY] = prefix + currentYear;
+  }
+
+  const isFullYear = fromFormat.includes('YYYY');
+
+  if (indexOfYY !== -1) {
```
illiarerun, line 84: >= 0 maybe?

src/formatDate.js
```diff
@@ -50,7 +50,49 @@
...
+    || separatedDate[fromFormat.indexOf('YYYY')];
+  const dateComponents = ['MM', 'DD', 'YY', 'YYYY'];
+
+  dateComponents.forEach((component, index) => {
+    const fromIndex = fromFormat.indexOf(component);
+    const toIndex = toFormat.indexOf(component);
+
+    if (fromIndex !== -1 && toIndex !== -1) {
+      resultedDate[toIndex] = separatedDate[fromIndex];
+    }
```
sergii-nosachenko, line 68: You can do all necessary date logic inside this loop. F.e.:
```js
const fromIndex = ...;
const toIndex = ...;

switch (component) {
  case 'DD':
    // logic for day
   break;
  case 'MM':
    // logic for month
   break;
  case 'YY':
  case 'YYYY':
    // logic to get a proper year
   break;
}
```

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):

PR #1097, opened 2023-02-18
Solution

nazarkharkevych:


src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+      year = splitedDate[i];
+    }
+
+    if (fromFormat[i] === 'MM') {
+      month = splitedDate[i];
+    }
+
+    if (fromFormat[i] === 'DD') {
+      day = splitedDate[i];
+    }
```
Volodymyr-Melnyk-Fron10der, line 70: It would looks better with switch/case construction

src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+    return `20${year}`;
+  }
+
+  if (yearFormat === 'YYYY' && year >= 30 && year.length === 2) {
+    return `19${year}`;
+  }
+
+  if (yearFormat === 'YYYY' && year.length === 4) {
+    return year;
+  }
```
Volodymyr-Melnyk-Fron10der, line 105: It's better to move all same conditions into the constants and reuse them
```suggestion
  const isFourY = yearFormat === 'YYYY';
  const isLessThirty = year < 30;
  const isLengthTwo = year.length === 2
  
  if (yearFormat === 'YY') {
    return year.slice(-2);
  }

  if (isFourY && isLessThirty && isLengthTwo) {
    return `20${year}`;
  }

  if (isFourY && !isLessThirty && isLengthTwo) {
    return `19${year}`;
  }

  if (isFourY && year.length === 4) {
    return year;
  }
```

Volodymyr-Melnyk-Fron10der (changes requested):
Hi!  Great job, good solution with getFormatedYear, I like it. But there are a few changes I suggest to make your code perfect

src/formatDate.js
```diff
@@ -50,7 +50,68 @@
...
+        year = splitedDate[i];
+        break;
+
+      case fromFormat[i] === 'MM':
+        month = splitedDate[i];
+        break;
+
+      case fromFormat[i] === 'DD':
+        day = splitedDate[i];
+        break;
```
slavanosatskyi, line 71:
```suggestion
    switch (fromFormat[i]) {
      case 'YY': 
      case 'YYYY':
        year = splitedDate[i];
        break;

      case 'MM':
        month = splitedDate[i];
        break;

      case 'DD':
        day = splitedDate[i];
        break;
```

slavanosatskyi (changes requested):
nice, I see only one place to be improved

Volodymyr-Melnyk-Fron10der (approved):
Perfect now✅

PR #1098, opened 2023-02-18
Format Date Advanced - Solution

Yevhenii0536:


src/formatDate.js
```diff
@@ -50,7 +50,60 @@
...
+      case 'DD':
+        current = oldDay;
+        break;
+
+      case 'MM':
+        current = oldMonth;
+        break;
+
+      case 'YY':
+        if (String(oldYear).length > 2) {
```
sergey-lipikhin, line 78:
```suggestion
        if (oldYear.length > 2) {
```

src/formatDate.js
```diff
@@ -50,7 +50,60 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldSeparator = fromFormat.find((item) => item.length === 1);
```
sergey-lipikhin, line 53:
```suggestion
  const oldSeparator = fromFormat[3];
```

- Overcomplicated
- If separator = ', ' (its length not 1) this way will not work

src/formatDate.js
```diff
@@ -50,7 +50,60 @@
...
+            current = String(19 + oldYear);
+            break;
+          }
+
+          current = String(20 + oldYear);
+          break;
+        }
+
+        current = oldYear;
+        break;
```
sergey-lipikhin, line 98: I'd better use `else`s and only 1 `break` statement.

sergey-lipikhin (approved):
The logic of program is better than mine:)
Some remarks are not critical but I should mentioned them.

src/formatDate.js
Yevhenii0536: actually, when separator on different index it's doesn't work too :)

sergii-nosachenko: What if the separator will be `--`? Better use the index cause the array length is strictly defined by the task and you can rely on it.

src/formatDate.js
```diff
@@ -50,7 +50,60 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldSeparator = fromFormat.find((item) => item.length === 1);
+  const newSeparator = toFormat.find((item) => item.length === 1);
+
+  const fromArr = date.split(oldSeparator);
+  const index = toFormat.length - 1;
```
sergii-nosachenko, line 57:
```suggestion
  const lastIndex = toFormat.length - 1;
```

src/formatDate.js
```diff
@@ -50,7 +50,60 @@
...
+        current = oldMonth;
+        break;
+
+      case 'YY':
+        if (String(oldYear).length > 2) {
+          current = oldYear.slice(2);
+          break;
+        }
+
+        current = oldYear;
```
sergii-nosachenko, line 83: Can be replaced with ternary operator

src/formatDate.js
```diff
@@ -50,7 +50,60 @@
...
+        break;
+
+      case 'YYYY':
+        if (oldYear.length < 4) {
+          if (+oldYear > 23) {
```
sergii-nosachenko, line 88: Wrong condition

src/formatDate.js
```diff
@@ -50,7 +50,60 @@
...
+            current = String(19 + oldYear);
+            break;
+          }
+
+          current = String(20 + oldYear);
```
sergii-nosachenko, line 93: Can be optimized and refactored with ternary operator

sergii-nosachenko (changes requested):

src/formatDate.js
Yevhenii0536: Yes, but as for me separator can be on index[1] or anywhere. What we need to do on this situation? And for this issues i found element with length = 1. Or i didn't right?

sergii-nosachenko: You have task, task determines how data is prepared. Also, your solution doesn't work for separator longer than 1 char.

VadKol (approved):
gj

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+  const lastIndex = toFormat.length - 1;
+
+  const oldDay = fromArr[fromFormat.findIndex((item) => item.includes('D'))];
+  const oldMonth = fromArr[fromFormat.findIndex((item) => item.includes('M'))];
+  const oldYear = fromArr[fromFormat.findIndex((item) => item.includes('Y'))];
+
+  const result = [];
+
+  for (let i = 0; i < lastIndex; i++) {
+    let current = '';
```
sergii-nosachenko, line 66: Useless

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+
+    switch (toFormat[i]) {
+      case 'DD':
+        current = oldDay;
```
sergii-nosachenko, line 70:
```suggestion
        result.push(oldDay);
```
Can be pushed immediately

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+        break;
+
+      case 'MM':
+        current = oldMonth;
+        break;
+
+      case 'YY':
+        current = oldYear.length > 2
+          ? current = oldYear.slice(2)
+          : current = oldYear;
```
sergii-nosachenko, line 80: ??? Assign assigned value?

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+        break;
+
+      case 'YYYY':
+        if (oldYear.length < 4) {
+          current = Number(oldYear >= 30)
+            ? current = String(19 + oldYear)
+            : current = String(20 + oldYear);
```
sergii-nosachenko, line 87: Why do you repeat code? Use ternary only for finding 19 || 20.
Why do you use assignments inside the ternary?
Why do you need to convert a boolean to a number type? You only need to convert `oldYear` value.

sergii-nosachenko (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,50 @@
...
+          ? oldYear
+          : oldYear.toString().slice(2, 4);
+
+        result.push(yearWithTwoChars);
+        break;
+
+      case 'YYYY':
+        const sentury = oldYear.length === 4
+          ? oldYear
+          : +oldYear >= 30 ? 19 + oldYear : 20 + oldYear;
```
sergii-nosachenko, line 86:
```suggestion
        const year = oldYear.length === 4
          ? oldYear
          : (+oldYear >= 30 ? 19 : 20) + oldYear;
```

src/formatDate.js
```diff
@@ -50,7 +50,50 @@
...
+        break;
+
+      case 'MM':
+        result.push(oldMonth);
+        break;
+
+      case 'YY':
+        const yearWithTwoChars = oldYear.length === 2
+          ? oldYear
+          : oldYear.toString().slice(2, 4);
```
sergii-nosachenko, line 78:
```suggestion
          : oldYear.toString().slice(-2);
```

src/formatDate.js
```diff
@@ -50,7 +50,50 @@
...
-  // write code here
+  const oldSeparator = fromFormat[3];
+  const newSeparator = toFormat[3];
+
+  const fromArr = date.split(oldSeparator);
+  const lastIndex = toFormat.length - 1;
+
+  const oldDay = fromArr[fromFormat.findIndex((item) => item.includes('D'))];
+  const oldMonth = fromArr[fromFormat.findIndex((item) => item.includes('M'))];
+  const oldYear = fromArr[fromFormat.findIndex((item) => item.includes('Y'))];
```
sergii-nosachenko, line 61:
```suggestion
  const getItem = mask => fromArr[fromFormat.findIndex((item) => item.includes(mask))]
  const oldDay = getItem('D');
  const oldMonth = getItem('M');
  const oldYear = getItem('Y');
```
Don't repeat yourself.

sergii-nosachenko (changes requested):

Andr1yK (approved):

PR #1099, opened 2023-02-19
Solution

IvanVaverchak:


src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+        str = arrDate[j].slice(2);
+        result.push(str);
+      }
+
+      if (fromFormat[j].length < toFormat[i].length
+        && fromFormat[j].includes('Y') && toFormat[i].includes('Y')) {
+        if (arrDate[j] >= 30) {
+          str = 19 + arrDate[j];
+        } else {
+          str = 20 + arrDate[j];
```
polosanya, line 70: You can use a ternary operator here instead

polosanya (approved):
Good job!

PR #1100, opened 2023-02-19
commit message

Peti51:
Hi can you give me some advice how can I make it more readable and more simple?

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -50,7 +50,80 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const splitted1 = date.split(/[-:/:.]/).reverse();
+
+  const splitted2 = date.split(/[-:/:.]/);
+
+  const separator = toFormat.slice(toFormat.length - 1);
+
+  const wholeJoinedFrom = fromFormat.join('');
+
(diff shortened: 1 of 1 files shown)


qoa11a (changes requested):
let's see how it can be done more easier

first of all, a few comments:
1. don't make an empty line of code before every new variable
2. no need to create so many variables

now, let me help you with the algorithm behind this task. Please try to implement what I write with code. You should start from the beginning and don't use your current solution

1. you can get an incoming date by using the split method. The separator in this case will be fromFormat[3]
2. separator by which you'll join your result at the end can be accessed by toFormat[3]
3. then, you can create an object with keys as fromFormat and values from your date that you got in step 1 `const mapDate = { [fromFormat[0]]: dateArr[0] }` and so on
4. then you can create an array of newDate that you'll use join method upon at the end
5. then you can iterate over first 3 elements of toFormat
6. in iteration you can check if your mapDate has property from toFormat. If it has than you can just push value stored in this key in your mapeDate to your newDate array and continue to the next step of iteration
7. otherwise, if it has no such property it means it's either 'YY' or it was 'YY' and become 'YYYY'. You should handle these cases
8. at the end, you should just return newDate.join(toFormat[3])

qoa11a (approved):

PR #1103, opened 2023-02-19
Created solution for changing format of date

gretoq:
- Learned about padStart;
- Had practice in creating some functions;
- Had practice with array methods.

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -50,7 +50,89 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateViewInObject = getCurrentDayMonthYear(date, fromFormat);
+
+  const newSeparator = toFormat[toFormat.length - 1];
+  const newDateViewInArray = [];
+
+  for (let j = 0; j < toFormat.length - 1; j++) {
+    const format = toFormat[j];
+
+    if (format === 'DD') {
+      if (dateViewInObject.day.toString().length < 2) {
+        dateViewInObject.day = '0' + dateViewInObject.day.toString();
+      }
+
+      newDateViewInArray.push(dateViewInObject.day);
+    }
+
+    if (format === 'MM') {
+      if (dateViewInObject.month.toString().length < 2) {
+        dateViewInObject.month = '0' + dateViewInObject.month.toString();
+      }
+
+      newDateViewInArray.push(dateViewInObject.month);
+    }
+
+    if (format === 'YYYY') {
+      if (dateViewInObject.year.toString().length === 4) {
+        newDateViewInArray.push(dateViewInObject.year);
+        continue;
+      }
+
+      if (dateViewInObject.year < 30) {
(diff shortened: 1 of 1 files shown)


sergii-nosachenko (changes requested):
Very complicated and overloaded. You need to think about optimization.
Divide your logic into the next simple steps:
- split `from` and `to` arrays by separator;
- create an empty object to hold date parts;
- loop the `from` array and add props to that object. You need to have 'DD', 'MM', 'YY', and 'YYYY';
- create an empty formatted date array;
- loop the `to` array and push into it values from date object using `to[i]` values as keys;
- return joined array.

It will be much more clear and more understandable. And takes less than 50 rows.

src/formatDate.js
```diff
@@ -1,78 +1,27 @@
...
-
 function formatDate(date, fromFormat, toFormat) {
-  const dateViewInObject = getCurrentDayMonthYear(date, fromFormat);
-  const newSeparator = toFormat[toFormat.length - 1];
-  const newDateViewInArray = getDesireFormatDate(dateViewInObject, toFormat);
-
-  return newDateViewInArray.join(newSeparator);
-}
+  const fromSeparator = fromFormat.pop();
+  const toSeparator = toFormat.pop();
```
sergii-nosachenko, line 12: Don't mutate function args

src/formatDate.js
```diff
@@ -1,78 +1,27 @@
...
-
-  for (let i = 0; i < formats.length - 1; i++) {
-    const format = formats[i];
-    const amount = +dateViewInArray[i];
+  fromFormat.forEach((format, i) => {
+    const amount = +date.split(fromSeparator)[i];
 
     switch (format) {
       case 'YY':
+        dateViewInObject.year = amount < 30
```
sergii-nosachenko, line 20:
```suggestion
        dateViewInObject.YY = value;
        dateViewInObject.YYYY = value < 30
```

src/formatDate.js
```diff
@@ -1,78 +1,27 @@
...
+          ? 2000 + amount
+          : 1900 + amount;
+        break;
+
       case 'YYYY':
         dateViewInObject.year = amount;
```
sergii-nosachenko, line 26: Add YY prop too.

src/formatDate.js
```diff
@@ -88,75 +37,39 @@ function getCurrentDayMonthYear(date, formats) {
...
-        break;
+      case 'MM':
+        const month = dateViewInObject.month < 10
+          ? `0${dateViewInObject.month}`
+          : `${dateViewInObject.month}`;
 
+        return month;
       default:
         throw new Error(`Unknown format: ${format}`);
     }
```
sergii-nosachenko, line 69: Just return `dateViewInObject[format]` value. Instead of all of this.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 🔥

PR #1105, opened 2023-02-20
formatDate first try

RufatRM:


qoa11a (changes requested):
let's rename our variables using camelCase

PR #1106, opened 2023-02-20
solution

Zelinskyi-Serhii:


src/formatDate.js
```diff
@@ -50,7 +50,62 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newDate = date.split(fromFormat[3]);
+  const toSeparate = toFormat[3];
+  const resultDate = [];
+  let month = '';
+  let day = '';
+  let year = '';
```
sergii-nosachenko, line 58: Instead of having multiple variables, create one object to hold values.

src/formatDate.js
```diff
@@ -50,7 +50,62 @@
...
+
+  for (let i = 0; i < newDate.length; i++) {
+    switch (fromFormat[i]) {
+      case 'MM':
+        month = newDate[i];
```
sergii-nosachenko, line 63:
```suggestion
        dateObject.MM = newDate[i];
```
Use masks as keys. You need to have all possible values: 'DD', 'MM', 'YY' and 'YYYY'

src/formatDate.js
```diff
@@ -50,7 +50,62 @@
...
+        break;
+
+      case 'DD':
+        day = newDate[i];
+        break;
+
+      case 'YY':
+      case 'YYYY':
+        year = newDate[i];
+        break;
```
sergii-nosachenko, line 73: Prepare year value here. Make both 'YY' and 'YYYY' values.

src/formatDate.js
```diff
@@ -50,7 +50,62 @@
...
+          break;
+        }
+
+        if (year.length === 2) {
+          resultDate[i] = '19' + year;
+        } else {
+          resultDate[i] = year;
+        }
+        break;
+    }
```
sergii-nosachenko, line 105: Useless `switch...case`. You'll need only to push `dateObject[toFormat[i]]` value to the resultDate.

sergii-nosachenko (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,28 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newDate = date.split(fromFormat[3]);
+  const toSeparate = toFormat[3];
+  const resultDate = [];
+  const dateObject = {};
```
sergii-nosachenko, line 56:
```suggestion
  const fromDateParts = date.split(fromFormat[3]);
  const toSeparator = toFormat[3];
  const toDateParts = [];
  const dateObject = {};
```

sergii-nosachenko (approved):
Good work 👍
Just use proper naming.

PR #1107, opened 2023-02-20
Solution js_format_date_advanced

sergey-lipikhin:


src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+  const convertedDayIndex = toFormat.indexOf('DD');
+  const convertedMonthIndex = toFormat.indexOf('MM');
+  const convertedYearShort = toFormat.indexOf('YY') !== -1;
+  const convertedYearIndex = Math.max(
+    toFormat.indexOf('YY'),
+    toFormat.indexOf('YYYY')
+  );
+
+  convertedDate[convertedDayIndex] = initialDate[initialDayIndex];
+  convertedDate[convertedMonthIndex] = initialDate[initialMonthIndex];
```
sergii-nosachenko, line 73: Instead of creating constants for every index, better refactor it to follow next logic:
Divide your logic into the next simple steps:
- split `from` and `to` arrays by separator;
- create an empty object to hold date parts;
- loop the `from` array and add props to that object. You need to have 'DD', 'MM', 'YY', and 'YYYY';
- create an empty formatted date array;
- loop the `to` array and push into it values from date object using `to[i]` values as keys;
- return joined array.

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+
+  if (initialYearShort && !convertedYearShort) {
+    if (initialDate[initialYearIndex] < 30) {
+      convertedDate[convertedYearIndex] = '20' + initialDate[initialYearIndex];
+    } else {
+      convertedDate[convertedYearIndex] = '19' + initialDate[initialYearIndex];
+    }
```
sergii-nosachenko, line 80: It can be done with ternary operator

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+  } else if (!initialYearShort && convertedYearShort) {
+    convertedDate[convertedYearIndex] = initialDate[initialYearIndex].slice(2);
+  } else {
```
sergii-nosachenko, line 83: Bad code style: nested `ifs`, `else if` constructions.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1108, opened 2023-02-20
Solution

Ssviatt:


slavanosatskyi (approved):
cool 😎 though just for the future - it is always a good idea to write default cases for switch statements

PR #1109, opened 2023-02-20
Solution

Zavalishyn-Oleksandr:


slavanosatskyi (approved):
You solution is good, though there is one disadvantage - fromFormat.pop(); your changing input params inside a function which side effect. And side effects are not desired.

PR #1113, opened 2023-02-22
Solution

vitalii-heiko:


src/formatDate.js
```diff
@@ -51,6 +51,59 @@
...
+        break;
+
+      case 'MM':
+        month = formatingDate[i];
+        break;
+
+      case 'DD':
+        day = formatingDate[i];
+        break;
+    }
```
Anastasiia-Svintsova, line 83: Default cases are missed in both loops. Add them, please. You always should add a default case to the switch/case statement to prevent bags

Anastasiia-Svintsova (changes requested):

anstsot (approved):
Good job, approved

PR #1121, opened 2023-02-26
formatted date

JasterHim:


src/formatDate.js
```diff
@@ -50,7 +50,40 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const arrayDate = date.split(`${fromFormat[3]}`);
+  const result = [];
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    switch (fromFormat[i]) {
```
witflash, line 57: If you use `switch / case` it should have a `default` statement.
It may return nothing or handle some errors for instance.

witflash (approved):
Great!
I left a comment just one comment for the future.

witflash (approved):

PR #1122, opened 2023-02-26
solution

serhiilysenko:


src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromFormatSeparator = fromFormat[fromFormat.length - 1];
+  const toFormatSeparator = toFormat[toFormat.length - 1];
+  const newDate = date.split(fromFormatSeparator);
+  const result = [];
+  let year,
+    month,
+    day;
+
```
ivbutsykin, line 60: Better create obj for this where will store `year,` `month` and `day`

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+  for (let i = 0; i < fromFormat.length; i++) {
+    if (fromFormat[i] === 'YY' || fromFormat[i] === 'YYYY') {
+      fromFormat[i] = newDate[i];
```
ivbutsykin, line 63: Should we modify this value?  If it's really need it should be separate variable, but try realise it without modify

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+      year = fromFormat[i];
```
ivbutsykin, line 64:
```suggestion
      year = newDate[i];
```

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromFormatSeparator = fromFormat[fromFormat.length - 1];
+  const toFormatSeparator = toFormat[toFormat.length - 1];
```
ivbutsykin, line 54: As I understand, fromFormat and toFormat always have 4 elements, so you can simplify:
```suggestion
  const fromFormatSeparator = fromFormat[3];
  const toFormatSeparator = toFormat[3];
```

ivbutsykin (changes requested):

PR #1126, opened 2023-02-28
Solution

Nataliia95254:


src/formatDate.js
```diff
@@ -50,7 +50,45 @@
...
+    obj.YYYY = (obj.YYYY).substring(2);
+  }
+
+  if (fromFormat.includes('YY') && toFormat.includes('YYYY')) {
+    if (obj.YY < 30) {
+      obj.YYYY = `20${obj.YY}`;
+    } else {
+      obj.YYYY = `19${obj.YY}`;
+    }
+  }
```
sTorba24, line 76:
```suggestion
  if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
    obj.YYYY = (obj.YYYY).substring(2);
    
    if (obj.YY < 30) {
      obj.YYYY = `20${obj.YY}`;
    } else {
      obj.YYYY = `19${obj.YY}`;
    }
  }
```

sTorba24 (changes requested):
Good work! lets improve it

src/formatDate.js
Nataliia95254: With these changing tests are faild
      ✕ if we shorten YYYY to YY (4ms)
      ✕ if we extend YY < 30 to YYYY (31ms)
      ✕ if we extend YY = 00 to YYYY (2ms)
      ✕ if we extend YY = 30 to YYYY
      ✕ if we extend YY > 30 to YYYY (2ms)

src/formatDate.js
```diff
@@ -50,7 +50,45 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const [ a, b, c, separator ] = fromFormat;
+  const [ , , , newSeparator ] = toFormat;
```
etojeDenys, line 54:
```suggestion
  const newSeparator = toFormat[3];
```

src/formatDate.js
```diff
@@ -50,7 +50,45 @@
...
+    switch (toFor) {
+      case 'YYYY': newData.push(obj.YYYY);
+        break;
+      case 'MM': newData.push(obj.MM);
+        break;
+      case 'DD': newData.push(obj.DD);
+        break;
+      case 'YY': newData.push(obj.YYYY);
+        break;
+    }
```
etojeDenys, line 88: don't forget to add default case

etojeDenys (changes requested):

src/formatDate.js
Nataliia95254: Could you suggest which one default value is better. 
If I add default: console.log ' error massage', I have  lint chack error  Unexpected console statement. 
If I add default  throw new Error 'error massage' I have error ✕ encountered a declaration exception

src/formatDate.js
OleziO: Not resolved, just replace your code with `const newSeparator = toFormat[3];`

src/formatDate.js
OleziO, line 76: Try to combine these two conditions, cause they are the same.

src/formatDate.js
OleziO: In the default, case set the old value, or make an empty default case with a break (first variant is better)

OleziO (changes requested):

src/formatDate.js
Nataliia95254: Could you please show me how to combine this cases. I optimized them with the ternary operator, but still two checks remain: 
if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
    obj.YYYY = (obj.YYYY).substring(2);
  }

    obj.YYYY = obj.YY < 30 ? `20${obj.YY}` : `19${obj.YY}`;
  }

Nataliia95254: > Try to combine these two conditions, cause they are the same.
Could you please show me how to combine this cases. I optimized them with the ternary operator, but still two checks remain:
if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
obj.YYYY = (obj.YYYY).substring(2);
}

obj.YYYY = obj.YY < 30 ? 20${obj.YY} : 19${obj.YY};
}

Nataliia95254: > Try to combine these two conditions, cause they are the same.

Could you please show me how to combine this cases. I optimized them with the ternary operator, but still two checks remain:
if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
obj.YYYY = (obj.YYYY).substring(2);
}

obj.YYYY = obj.YY < 30 ? 20${obj.YY} : 19${obj.YY};
}

tetiana-ivanchenko (approved):

PR #1129, opened 2023-03-01
formatDate solution

RomanTsvindin:


src/formatDate.js
```diff
@@ -50,7 +50,29 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newFormat = [];
+  const splittedDate = date.split(fromFormat[3]);
```
Volodymyr-Melnyk-Fron10der, line 54: I would move  fromFormat[3] to a variable as it's using 2 times

Volodymyr-Melnyk-Fron10der (approved):
FantasticSimonCowellGIF

PR #1135, opened 2023-03-08
formatDate was created

Pavlo-Petrashevskyi:


src/formatDate.js
```diff
@@ -48,9 +48,34 @@
...
+        if (dateCopy[i] < 30) {
+        newArray[y] = `20${dateCopy[i]}`;
+        }
+
+        if (dateCopy[i] >= 30) {
+          newArray[y] = `19${dateCopy[i]}`;
+        }
+      }
+    }
+  }
```
DarkMistyRoom, line 76: Let's improve performance. We can avoid using nested loops by adding splited date to object like here:
`{
  'DD': 19,
  'MM': 02
  ...
}`

then we shouldn't iterate by the array to get access to specific property

DarkMistyRoom (changes requested):

src/formatDate.js
Pavlo-Petrashevskyi: Done.

PR #1137, opened 2023-03-09
solution

valeriaoleynik:


src/formatDate.js
```diff
@@ -50,7 +50,37 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const separator = fromFormat.slice(-1);
+  const newSeparator = toFormat.slice(-1);
+  const copyDate = date.split(separator);
+  const copyFromFormat = fromFormat.slice(0, 3);
+  const copyToFormat = toFormat.slice(0, 3);
```
witflash, line 57: Why do you need these array copies?
You can use `fromFormat` and `toFormat` directly by using cycle like this:
```
for (let i = 0; i < toFormat.length - 1; i++)
```

witflash (approved):
Good work!

Check my comment for your future work.

PR #1148, opened 2023-03-17
Advanced date format

roshirin:


M1k1ta (approved):
I think you write too big code. But this code works and I didn't find mistakes. So good job!

roshirin:
Just following the recomendations we get in our checklists every time.

src/formatDate.js
```diff
@@ -50,7 +50,74 @@
...
+        this.year = '20' + this.year;
+
+        return;
+      }
+
+      this.year = '19' + this.year;
+    },
+  };
+
+  const originDate = date.split(dateObject.inSepar);
```
liliia-livandovskaya, line 74: it would be better to move all the variables declarations to the top

liliia-livandovskaya (approved):

PR #1153, opened 2023-03-18
create function format Date

VenomMB:


glots-a (approved):

glots-a:
well done!

src/formatDate.js
```diff
@@ -50,7 +50,46 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const separatorFromFormat = fromFormat.pop();
+  const separatorToFormat = toFormat.pop();
+  const numbersOfDate = date.split(separatorFromFormat);
+  const currentFormat = {};
+  const correctFormat = [];
+  let i = 0;
```
liliia-livandovskaya, line 58: put it to the `for`, there is no need to move this variable from the loop

liliia-livandovskaya, line 58:
```suggestion
```
explain why you need it or just remove

src/formatDate.js
```diff
@@ -50,7 +50,46 @@
...
+
+  for (const format of fromFormat) {
+    currentFormat[format] = numbersOfDate[i];
+    i++;
+  }
+
+  let yearsFormat = 0;
```
liliia-livandovskaya, line 65: move all the variables declarations to the top

src/formatDate.js
```diff
@@ -50,7 +50,46 @@
...
+
+  for (const newFormat of toFormat) {
+    if (!fromFormat.includes(newFormat) && newFormat === 'YY') {
+      yearsFormat = currentFormat['YYYY'] % 100;
+      correctFormat.push(yearsFormat);
+      continue;
```
liliia-livandovskaya, line 71:
```suggestion
```
you can remove `continue`

src/formatDate.js
VenomMB: okay) I removed it.

src/formatDate.js
VenomMB: I can't remove continue because my program must go to next iteration if fromFormat includes 'YY' or 'YYY', because if this conditions aren't done, i push number in array with correct format.

PR #1159, opened 2023-03-20
Solution

DanielSorokowski:


src/formatDate.js
```diff
@@ -50,7 +50,39 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const givenDate = date.split(fromFormat[3]);
+  let day;
+  let month;
+  let year;
```
solaryasha, line 56: Let's create an object with such keys

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+    } else if (toFormat[i] === 'MM') {
+      newDateFormat.push(month);
+    } else if (toFormat[i] === 'YYYY') {
+      newDateFormat.push(year);
+    } else {
+      newDateFormat.push(year.slice(-2));
+    }
+  }
+
+  return newDateFormat.join(toFormat[3]);
```
solaryasha, line 85: what is `toFormat[3]` ?
Let's store this value in the variable

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+    } else if (fromFormat[i] === 'MM') {
+      month = givenDate[i];
+    } else {
+      year = givenDate[i];
+    }
+  }
+
+  if (year.length < 4) {
+    year = year < 30 ? `20${year}` : `19${year}`;
+  }
```
solaryasha, line 71: Let's create a function for such formatting

solaryasha (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,47 @@
...
+    } else if (fromFormat[i] === 'MM') {
+      newDate.month = givenDate[i];
+    } else {
+      newDate.year = givenDate[i];
+    }
+  }
+
+  const makeFullYear = (year) => {
+    return year < 30 ? `20${year}` : `19${year}`;
+  };
```
solaryasha, line 75: Do we need to create a new function for each `formatDate` call ?

solaryasha (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,58 @@
...
+        break;
+      default:
+        return 'wrong data provided';
+    }
+  }
+
+  if (newDate.year.length < 4) {
+    const makeFullYear = (year) => {
+      return year < 30 ? `20${year}` : `19${year}`;
+    };
```
solaryasha, line 83: This function should be outside `formatDate` function body

solaryasha (changes requested):

solaryasha (approved):

PR #1160, opened 2023-03-20
Solution

Zibi95:


src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
+  toFormat.forEach(newElement => {
+    const index = fromFormat.findIndex(el => {
+      return el.includes(newElement.slice(0, 1));
+    });
+
+    switch (newElement) {
+      case 'YYYY':
+        if (newElement > year) {
+          const toPush = dateArray[index] === '00' || +dateArray[index] < 30
+            ? '20' + dateArray[index]
```
Radoslaw-Czerniawski, line 68: Try to avoid adding strings and use template literals, here it's fine since you are working only on strings however JS sometimes does some crazy things like
```
3 * 5 => 15
"3" * "5" => 15
2 + " fish" => "2 fish"
"3" + "5" => "35"  // hmmm...
```
Also it's more readable when using template literals

src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateArray = date.split(fromFormat[3]);
+  const newDateFormat = [];
+  const separator = toFormat[3];
+
+  const year = fromFormat.find(el => el.includes('Y'));
+
+  toFormat.forEach(newElement => {
```
Radoslaw-Czerniawski, line 59: You could potentially change the name of the newElements it's kinda confusing. Maybe something like `newFormatElement` also you could potentially use map and return values instead of pushing them to an array outside of the scope of this method. This however would force you to pop the last element, separator or undefined value depending on you implementation. Nothing wrong with forEach here, good implementation of the formatDate overall.

Radoslaw-Czerniawski (approved):
My brain hurts. Very good pr, approved 💪

PR #1162, opened 2023-03-20
solution

Nazarii98:


Zad0rnik (approved):
good solution)

src/formatDate.js
```diff
@@ -51,6 +51,68 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
```
liliia-livandovskaya, line 53: dont forget to remove comments

src/formatDate.js
```diff
@@ -51,6 +51,68 @@
...
+    }
+  }
+
+  if (YY === '') {
+    YY = YYYY.slice(2);
+  } else if (YYYY === '' && YY <= 23) {
+    YYYY = `20${YY}`;
+  } else if (YYYY === '' && YY > 23) {
+    YYYY = `19${YY}`;
+  }
```
liliia-livandovskaya, line 90: it is better to avoid nesting

liliia-livandovskaya (approved):
Great job 👍

PR #1173, opened 2023-03-22
Solution

yurii-shkrobut-m:


src/formatDate.js
```diff
@@ -50,7 +50,63 @@
...
+  const separator = fromFormat[fromFormat.length - 1];
+  const joiner = toFormat[toFormat.length - 1];
+  const dateArr = date.split(separator);
+  let year = 0;
+  let month = 0;
+  let day = 0;
+  const arr = [];
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    if (fromFormat[i] === 'YYYY' || fromFormat[i] === 'YY') {
```
TarasHoliuk, line 62: use `switch` in loops (in both loops) instead of multiple `if` to check it

TarasHoliuk (changes requested):

DarkMistyRoom (approved):
Good job, I approve of it. But also I want you think about using object to store year, month, and day instead of simple variables. It may make your code shorter. I you want to try, just make rerequest

PR #1176, opened 2023-03-23
Dev

popeye88:


src/formatDate.js
```diff
@@ -50,7 +50,55 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const initialDate = date.split(fromFormat[fromFormat.length - 1]);
+  const initialFormat = fromFormat.slice(0, 3);
+
+  const requiredFormat = toFormat.slice(0, 3);
+  const joiner = toFormat[toFormat.length - 1];
```
witflash, line 57: I think it's not "joiner", it's "separator" :) Never mind )

witflash (approved):
Hey!
The cool solution in a clear imperative way 👍🏼

If you want, you may try to think about a more declarative solution. You can do it with just one cycle. Try it when you'll have time.

PR #1181, opened 2023-03-28
Solution

NVlasenko:


src/formatDate.js
```diff
@@ -50,7 +50,32 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newDate = [];
+  const arr = date.replace(/\W|_/g, '.').split().join('.').split('.');
```
tetiana-ivanchenko, line 54:
```suggestion
  const arr = date.replace(/\W|_/g, '.').split('.');
```

src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    for (let k = 0; k < toFormat.length; k++) {
+      if (fromFormat[i] === toFormat[k]) {
```
tetiana-ivanchenko, line 58: can you try to avoid nested loops?

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #1191, opened 2023-03-31
solution

danielpopek94:


SuaBy514 (approved):

src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
+
+  function dateFactory() {
+    for (let i = 0; i < fromFormat.length - 1; i++) {
+      if (fromFormat[i] === 'DD') {
+        day = inputDateComponents[i];
+      } else if (fromFormat[i] === 'MM') {
+        month = inputDateComponents[i];
+      } else if (fromFormat[i] === 'YY' || fromFormat[i] === 'YYYY') {
+        year = inputDateComponents[i];
+        previousYearFormat = fromFormat[i];
```
d-wasilewski, line 77: you can go with the switch(case) here, since you know all of the cases

src/formatDate.js
```diff
@@ -50,7 +50,81 @@
...
+            const outputYear = this.convertYearFormat(
+              this.year,
+              this.yearFormat,
+              currentFormat
+            );
+
+            outputDateComponents.push(outputYear);
+            break;
+          default:
+            console.error(`Wrong format: ${currentFormat}`);
```
d-wasilewski, line 105: The use of console.error in the dateComponents object's methods is not ideal. It is better to throw an error so that it can be handled properly in the calling function.

src/formatDate.js
```diff
@@ -50,7 +50,81 @@
...
+      }
+
+      const prefix = year < 30 ? '20' : '19';
+
+      return prefix + year;
+    },
+  };
+
+  dateComponents.dateFactory();
+  dateComponents.formatFactory();
```
d-wasilewski, line 125: for simplicity it's may be better to define the functions as standalone functions outside of an object. In this case for dateFactory and formatFactory you need to reference the object dateComponents every time you want to call a function, inside of it. It wouldn't work if you moved it outside of the formatDate function. convertYearFormat is an example of a function with a good purpose that actually helps and might be reused. However, move it outside of the formatDate function. date and format factories functions are not bringing any value, you can remove them and just run the code inside of this function.

src/formatDate.js
```diff
@@ -50,7 +50,81 @@
...
+  const inputSeparator = fromFormat[fromFormat.length - 1];
+  const inputDateComponents = date.split(inputSeparator);
+
+  const outputSeparator = toFormat[toFormat.length - 1];
+  const outputDateComponents = [];
+
+  const dateComponents = {
+    day: '',
+    month: '',
+    year: '',
```
d-wasilewski, line 62: these 3 are enough in this date object

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1195, opened 2023-04-03
Solution

SuaBy514:


src/formatDate.js
```diff
@@ -51,6 +51,55 @@
...
+  }
+
+  const reformatted = [0, 0, 0];
+
+  reformatted[newDayIndex] = numbers[oldDayIndex];
+  reformatted[newMonthIndex] = numbers[oldMonthIndex];
+
+  if (oldYearFull === true && newYearFull === false) {
+    reformatted[newYearIndex]
+    = (numbers[oldYearIndex][2] + numbers[oldYearIndex][3]);
```
d-wasilewski, line 95: what is numbers[oldYearIndex][2]?

d-wasilewski: this part of the code is hard to read

d-wasilewski (changes requested):

src/formatDate.js
SuaBy514: It was a makeshift solution to chopping off the first 2 characters, I did not know about the cool functions that do that easily back then

d-wasilewski (approved):

PR #1197, opened 2023-04-03
format-date solution

StepuraSerhii:


SerhiiKirik (approved):
Well done 🚀 
You can make the better algorithm & solve it with 1 for loop 🧑‍💻

PR #1202, opened 2023-04-06
Formated date was added

hyrych-max:


src/formatDate.js
```diff
@@ -50,7 +50,46 @@
...
+
+      case 'YY':
+        if (arrFromFormat[i].slice(0, 2) >= 30) {
+          formatedDate[yearLong] = '19' + arrFromFormat[i];
+          break;
+        }
+
+        formatedDate[yearLong] = '20' + arrFromFormat[i];
+        break;
+    }
```
Anastasiia-Svintsova, line 89: You should always add a default case to the switch /case construction to avoid unpredictable code breaks

Volodymyr-Melnyk-Fron10der (approved):

PR #1203, opened 2023-04-06
Solution

Sergey8793:


src/formatDate.js
```diff
@@ -50,7 +50,28 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newDate = [];
+  const obj = {};
+  const splitDate = date.split(fromFormat[3]);
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    obj[fromFormat[i]] = splitDate[i];
+
+    if (fromFormat[i] !== 'YY') {
```
TarasHoliuk, line 60: Does it work properly? because this condition will be `true` for `'DD'` and `'MM'` as well (at least it will overwrite your `obj.YY` value every time)

write here `fromFormat[i] === 'YY'` and do things you are doing inside your `else` block
then add another one `if` with `fromFormat[i] === 'YYYY'` and write there this code: `obj.YY = obj[fromFormat[i]].slice(2, 4);`

TarasHoliuk (changes requested):
Good solution (if it works ;)), but I'd recommend to refactor if a bit

tetiana-ivanchenko (approved):

PR #1204, opened 2023-04-06
Solution

NikaNika12:


src/formatDate.js
```diff
@@ -51,6 +52,28 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
```
tetiana-ivanchenko, line 54: don't left comments

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (changes requested):
fix prev comments

src/formatDate.js
NikaNika12: ı delete all my comments

NikaNika12:
no my comments

anstsot (approved):
Good job, approved

PR #1206, opened 2023-04-09
Solution

AnastassiaH:
Looks horrible, I know)

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -51,6 +51,34 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
-}
+  const arrDate = date.split(fromFormat[fromFormat.length - 1]);
+  const res = [];
+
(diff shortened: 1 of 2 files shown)


src/formatDate.js
```diff
@@ -51,6 +51,34 @@
...
+            if (+arrDate[x] < 30) {
+              res[i] = `20${arrDate[x]}`;
+            };
+
+            if (+arrDate[x] >= 30) {
+              res[i] = `19${arrDate[x]}`;
+            };
+          };
+        };
+      };
```
nazarmatsevych, line 77: Yeah, looks not really cool, we can try to rewrite it using switch/case statement or use [guard clauses](https://medium.com/@timothydan/javascript-guard-clauses-64b999e3240) instead of nested ifs statements

But task done so it's up to you

nazarmatsevych (approved):

PR #1208, opened 2023-04-09
Solution

MaksymKos:


src/formatDate.js
```diff
@@ -50,7 +50,61 @@
...
+    if (formatPart === 'YYYY') {
+      year = value;
+    } else if (formatPart === 'YY') {
+      year = value < 30 ? 2000 + value : 1900 + value;
+    } else if (formatPart === 'MM') {
+      month = value;
+    } else if (formatPart === 'DD') {
+      day = value;
+    }
+  }
```
etojeDenys, line 69: better to use here switch\case it will improve readability

etojeDenys (approved):
LGTM

PR #1211, opened 2023-04-12
Solution

Ir-ra:


witflash (approved):
Good classic imperative solution.
If you want you can try to make an alternative (more declarative) solution.

PR #1213, opened 2023-04-12
solution

mariuszgit:


src/formatDate.js
```diff
@@ -50,7 +50,68 @@
...
+      return `20${el}`;
+    } else {
+      return `19${el}`;
+    }
+  };
+
+  return el;
+};
+
+function toShortYear(el) {
```
Radoslaw-Czerniawski, line 113: This may be a little too much to have a function only for slicing the string, but I get the idea of having a more descriptive code.

src/formatDate.js
```diff
@@ -50,7 +50,68 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const resultTab = [];
+  const dateObj = {};
+
+  // save date using fromFormat-arg
```
Radoslaw-Czerniawski, line 56: Something more like // transform and map date values to dateObj based on fromFormat

src/formatDate.js
```diff
@@ -50,7 +50,68 @@
...
+      case 'D':
+        dateObj.day = el;
+        break;
+
+      default:
+        throw new Error('something went wrong with fromFormat');
+    }
+  });
+
+  // pull new date from toFormat-arg
```
Radoslaw-Czerniawski, line 76: Something more like // transform and push in right order values of dateObj based on toFormat

Radoslaw-Czerniawski (approved):
Very nice implementation. :) Do you think you could refactor it to only one iteration instead of two? 🤔

PR #1214, opened 2023-04-12
Solution

PolinaKM:


src/formatDate.js
```diff
@@ -50,7 +50,40 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const parts = date.split(fromFormat[fromFormat.length - 1]);
```
TarasHoliuk, line 53: `fromFormat[fromFormat.length - 1]` looks a bit clumsy
you may get separator and other keys in this way: `const [...fromFormatKeys, separator] = fromFormat;`

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+        }
+        break;
+      default:
+        newDate += parts[fromFormat.indexOf(toFormat[i])];
+        break;
+    }
+
+    if (i < toFormat.length - 2) {
+      newDate += toFormat[toFormat.length - 1];
+    }
```
TarasHoliuk, line 83: instead of adding a separator to a string inside such `if` block you may create an empty array and get rid of this `if`
then, return the array joined by the `separator` below

TarasHoliuk (changes requested):

src/formatDate.js
PolinaKM: I get `A rest element must be last in a destructuring pattern.ts(2462) Parsing error: Rest element must be last element.` with your approach. I have another idea.

src/formatDate.js
```diff
@@ -50,7 +50,48 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldSeparator = fromFormat.pop();
+  const newSeparator = toFormat.pop();
+  const parts = date.split(oldSeparator);
+  const newDateArr = ['', '', ''];
```
tetiana-ivanchenko, line 56:
```suggestion
  const newDateArr = [];
```

src/formatDate.js
```diff
@@ -50,7 +50,48 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldSeparator = fromFormat.pop();
```
tetiana-ivanchenko, line 53: don't change function arguments

tetiana-ivanchenko (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,48 @@
...
+
+        newDateArr[toFormat.indexOf('YY')] = shortYear;
+        break;
+      case 'DD':
+        newDateArr[toFormat.indexOf('DD')] = parts[fromFormat.indexOf('DD')];
+        break;
+      case 'MM':
+        newDateArr[toFormat.indexOf('MM')] = parts[fromFormat.indexOf('MM')];
+        break;
+    }
```
Volodymyr-Melnyk-Fron10der, line 91: set default for each switch construction

Volodymyr-Melnyk-Fron10der (approved):
Good job!

PR #1215, opened 2023-04-14
Solution

andrijdudar:


src/formatDate.js
```diff
@@ -48,9 +48,74 @@
...
+        }
+        result.push(year);
+        break;
+
+      case 'YYYY':
+        newYerFormat = 4;
+
+        if (oldYerFormat < newYerFormat && year < 30) {
+          year = '20' + year;
+        } else if (oldYerFormat < newYerFormat && year >= 30) {
```
Volodymyr-Melnyk-Fron10der, line 99: i would move oldYerFormat < newYerFormat into a variable

Volodymyr-Melnyk-Fron10der (approved):
good👍

PR #1220, opened 2023-04-18
Solution

4ykis:


src/formatDate.js
```diff
@@ -50,7 +50,47 @@
...
+    ) {
+      return year;
+    }
+
+    if (length === 4 && year.length === 2) {
+      if (year >= 30) {
+        return '19' + year;
+      }
+
+      return '20' + year;
```
VitaliyBondarenko1982, line 83:
```suggestion
      return year >= 30 ? '19' + year : '20' + year;
```

src/formatDate.js
```diff
@@ -50,7 +50,47 @@
...
+        break;
+      }
+    }
+  }
+
+  function convertYear(year, length) {
+    if (
+      (length === 4 && year.length === 4)
+      || (length === 2 && year.length === 2)
+    ) {
```
VitaliyBondarenko1982, line 74:
```suggestion
    if (length === year.length) {
```
looks like it can be simplier

VitaliyBondarenko1982 (approved):
Looks good.
Just pay attention on suggestions

PR #1223, opened 2023-04-20
js_format_date

YelyzavetaMolochynska:


src/formatDate.js
```diff
@@ -50,7 +50,30 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newFormat = [];
+  const dateArr = date.split(fromFormat[3]);
+  const dateYear = dateArr[fromFormat.indexOf('YYYY')]
+    || dateArr[fromFormat.indexOf('YY')];
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
```
Viktor-Kost, line 58: Consider using a 'for...of' loop instead of a 'for' loop to iterate over the 'toFormat' array. This makes the code more readable and avoids the need to use an index. For example: `for (const format of toFormat) { ... }`

src/formatDate.js
```diff
@@ -50,7 +50,30 @@
...
+        break;
+
+      default:
+        if (toFormat[i].length < dateYear.length) {
+          newFormat.push(dateYear.slice(2, 4));
+        } else if (toFormat[i].length > dateYear.length) {
+          newFormat.push((dateYear < 30 ? 20 : 19) + dateYear);
+        } else {
+          newFormat.push(dateYear);
+        }
```
Viktor-Kost, line 72: these default behavior looks bulky. It is better to use `code decomposition` or `early` return instead of `else if`

Viktor-Kost (changes requested):

mykhalenych (approved):

PR #1227, opened 2023-04-21
done

maxonshev:


src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+  const dateObj = {};
+
+  for (let i = 0; i < 3; i++) {
+    if (fromFormat[i] === 'YY' || fromFormat[i] === 'YYYY') {
+      dateObj.year = dateParts[i];
+    } else if (fromFormat[i] === 'MM') {
+      dateObj.month = dateParts[i];
+    } else if (fromFormat[i] === 'DD') {
+      dateObj.day = dateParts[i];
+    }
```
SanyaBratashchuk, line 64:
```suggestion
    } 
      dateObj[fromFormat[i]] = dateParts[i];
```
You can just use MM and DD instead of month and day

SanyaBratashchuk (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+  const dateObj = {};
+
+  for (let i = 0; i < 3; i++) {
+    if (fromFormat[i] === 'YY' || fromFormat[i] === 'YYYY') {
+      dateObj.year = dateParts[i];
+    } else if (fromFormat[i] === 'MM') {
+      dateObj.MM = dateParts[i];
+    } else if (fromFormat[i] === 'DD') {
+      dateObj.DD = dateParts[i];
+    }
```
Anastasiia-Svintsova, line 64: Use switch/case to avoid nesting, please

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+  const newDateParts = [];
+
+  for (let i = 0; i < 3; i++) {
+    if (toFormat[i] === 'YY' || toFormat[i] === 'YYYY') {
+      newDateParts.push(dateObj.year);
+    } else if (toFormat[i] === 'DD') {
+      newDateParts.push(dateObj.DD);
+    } else if (toFormat[i] === 'MM') {
+      newDateParts.push(dateObj.MM);
+    }
```
Anastasiia-Svintsova, line 84: Same here

Anastasiia-Svintsova (changes requested):

anstsot (approved):
Good job, approving. 
But the code can be simplified and be written only in one `for` loop

maxonshev:
> Good job, approving. But the code can be simplified and be written only in one `for` loop

how to do this? i can't understand

PR #1231, opened 2023-04-25
Solution

Artem193:


src/formatDate.js
```diff
@@ -50,7 +50,75 @@
...
+  if (toFormat[toYearIndex] === 'YYYY' && fromFormat[fromYearIndex] === 'YY') {
+    if (+fromDateArr[fromYearIndex] > 29) {
+      toYear = `19${fromDateArr[fromYearIndex]}`;
+    } else {
+      toYear = `20${fromDateArr[fromYearIndex]}`;
+    }
+  }
+
+  switch (0) {
+    case toYearIndex:
```
OleziO, line 84: Move all cases to the object.

src/formatDate.js
```diff
@@ -50,7 +50,75 @@
...
+      toDateArr.push(toYear);
+      break;
+    case toMonthIndex:
+      toDateArr.push(fromDateArr[fromMonthIndex]);
+      break;
+    case toDayIndex:
+      toDateArr.push(fromDateArr[fromDayIndex]);
+      break;
+  }
```
OleziO, line 105: Its bad practice. You need to move some variables to switch and check it with cases. It helps you optimize code.

OleziO (changes requested):

PR #1236, opened 2023-04-26
Solution

anpiet:


AndrzejPec (approved):
Nice job! It's concise although I encourage you to ponder on the following:

Your code is nested like a Russian doll ;D it could lead to performance issues for larger inputs.

The var name 'tempString' is... just as temporary as my hope to understand the code at first glance ;)

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateArr = date.split(fromFormat.pop());
+  const correctDate = [];
+  let tempString = '';
+
+  for (let i = 0; i < 4; i++) {
+    for (let j = 0; j < 3; j++) {
+      if (fromFormat[j].slice(0, 2) === toFormat[i].slice(0, 2)) {
```
asakevych, line 59: can you think about simplier logic? I think it's possible not to use loop inside loop

- you can read about algorithm complexity and why it's better to avoid having loop inside loop 

i think i's a good task to use object properties as well 😉

asakevych (changes requested):
Great job!! 👏  but I think the solution can be simplified, try to think about other solution 😉

choeqq (approved):
Looks good to me ✅

PR #1237, opened 2023-04-27
solution

HannieA:


src/formatDate.js
```diff
@@ -50,7 +50,60 @@
...
+      case 'YY':
+      case 'YYYY':
+        year = dates[i];
+        break;
+      case 'MM':
+        month = dates[i];
+        break;
+      case 'DD':
+        day = dates[i];
+        break;
```
Anastasiia-Svintsova, line 73: The switch/case construction should always have a default case to avoid unpredictable issues.

src/formatDate.js
```diff
@@ -50,7 +50,60 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let year = '';
+  let month = '';
+  let day = '';
+  const [,,, separatorFF] = fromFormat;
```
Anastasiia-Svintsova, line 56: It's easier to read, in my opinion. But it's really good that you know how to use destructuring
```suggestion
  const separator = fromFormat[3];
```

src/formatDate.js
```diff
@@ -50,7 +50,60 @@
...
+      case 'YYYY':
+        if (toFormat[i].length > year.length
+          && year < 30) {
+          newFormat.push(`20${year}`);
+        } else if (toFormat[i].length > year.length
+          && year >= 30) {
+          newFormat.push(`19${year}`);
+        } else {
+          newFormat.push(year);
+        }
```
Anastasiia-Svintsova, line 95: You can use a ternary operator to simplify it. Smth like this
```
const centuryPrefix = year < 30 ? '20' : '19';
const yearToPush = toFormat[i].length > year.length ? centuryPrefix + year : year;

Anastasiia-Svintsova (approved):

PR #1238, opened 2023-04-29
Solution

generatedman:


src/formatDate.js
```diff
@@ -50,7 +50,49 @@
...
+  let day = '';
+
+  for (const i of fromFormat) {
+    if (i === 'YY' || i === 'YYYY') {
+      year += dateArr[fromFormat.indexOf(i)];
+    } else if (i === 'MM') {
+      month += dateArr[fromFormat.indexOf(i)];
+    } else if (i === 'DD') {
+      day += dateArr[fromFormat.indexOf(i)];
+    }
```
NickMitchell61, line 65: let's avoid the "else if" statement in all cases

src/formatDate.js
```diff
@@ -50,7 +50,49 @@
...
+    } else {
+      year = '19' + year;
+    }
+  }
+
+  if (toFormat.includes('YY') && year.length > 2) {
+    year = year.split('');
+    year.shift();
+    year.shift();
+    year = year.join('');
```
NickMitchell61, line 80: you can use `year.slice(2)` instead

NickMitchell61 (changes requested):
Good job! See some suggestions, that I propose to get your code more readability

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateArr = date.split(fromFormat[3]);
```
33Orange, line 53: Use destructuring to extract values from arrays. Example: const [day, month, year] = dateArr;

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+  let year = '';
+  let month = '';
+  let day = '';
+
+  for (const i of fromFormat) {
```
33Orange, line 58: Use a more descriptive variable name, instead of 'i' use 'element' or 'formatElement'.

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+    }
+  }
+
+  if (toFormat.includes('YY') && year.length > 2) {
+    year = year.split('').slice(2).join('');
+  }
+
+  const dateFormated = [];
+
+  for (const i of toFormat) {
```
33Orange, line 86: Use a more descriptive variable name, instead of 'i' use 'element' or 'formatElement'.

33Orange (changes requested):
Almost done!

A few moments with readability

src/formatDate.js
generatedman: the fact is that destructuring is not appropriate here, since we cannot say exactly in what order the hour, month, and day go

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+    if (element === 'DD') {
+      dateFormated.push(day);
+    }
+
+    if (element === 'MM') {
+      dateFormated.push(month);
+    }
+  }
+
+  return dateFormated.join(toFormat[3]);
```
33Orange, line 100: Use a default parameter for the separator instead of using 'toFormat[3]'.

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+    if (element === 'DD') {
+      day += dateArr[fromFormat.indexOf(element)];
+    }
+  }
+
+  if (toFormat.includes('YYYY') && year.length < 4) {
+    if (year < 30) {
+      year = '20' + year;
+    } else {
+      year = '19' + year;
```
33Orange, line 76: Use template strings instead of concatenation.

33Orange (approved):
Well done! 
Pay attention to the comments. You can improve the code a little, but it's not critical

PR #1241, opened 2023-04-30
Solution

vintwp:


src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+    separator: fromFormat[fromFormat.length - 1],
+  };
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    if (fromFormat[i] === 'YYYY' || fromFormat[i] === 'YY') {
+      dateInput.dataConverter(
+        fromFormat[i], date.split(dateInput.separator)[i]
+      );
+    } else {
+      dateInput[fromFormat[i]] = date.split(dateInput.separator)[i];
```
nazarmatsevych, line 76: Similar to line 25, store the split date in a variable and use it in the loop instead of splitting it again and again.

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const formattedArray = [];
+
+  const dateInput = {
+
+    dataConverter(format, value) {
```
nazarmatsevych, line 57: The 'dataConverter' function is part of the 'dateInput' object. Consider using a more descriptive name, such as 'convertDateComponent'.

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+      }
+    },
+
+    separator: fromFormat[fromFormat.length - 1],
+  };
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    if (fromFormat[i] === 'YYYY' || fromFormat[i] === 'YY') {
+      dateInput.dataConverter(
+        fromFormat[i], date.split(dateInput.separator)[i]
```
nazarmatsevych, line 73: You can optimize the code by storing the split date in a variable and use it in the loop instead of splitting it again and again.

nazarmatsevych (approved):

PR #1244, opened 2023-05-02
finished

EvhenIvchenko:


src/formatDate.js
```diff
@@ -50,7 +50,43 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateArray = date.split(fromFormat[fromFormat.length - 1]);
+  let year, month, day;
+
+  for (let i = 0; i < dateArray.length; i++) {
```
mykhalenych, line 56: Instead of using a loop to assign date components, you can use 'findIndex' method to find the index of the component in the 'fromFormat' array and use that index to get the value from 'dateArray'. This will simplify the code and improve readability.

EvhenIvchenko: Okay , thank you

src/formatDate.js
```diff
@@ -50,7 +50,44 @@
...
+
+  const monthIndex = fromFormat.findIndex((format) => format === 'MM');
+  const month = dateArray[monthIndex];
+
+  const dayIndex = fromFormat.findIndex((format) => format === 'DD');
+  const day = dateArray[dayIndex];
+
+  const replacedDate = toFormat.map((format) => {
+    switch (format) {
+      case 'YYYY':
```
OleziO, line 75: Move all cases to the object

OleziO (changes requested):

src/formatDate.js
EvhenIvchenko: thanks , fixed

VitaliyBondarenko1982 (approved):

PR #1245, opened 2023-05-02
Develop

vitalii-fedusov:


src/formatDate.js
```diff
@@ -51,6 +51,63 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
```
OleziO, line 53: Remove all comments from the code.

OleziO (approved):

PR #1252, opened 2023-05-06
add task solution

maksym-nemera:


33Orange (approved):
Well done.

1 small moment with century.
Would be better to concatenate 2 strings.

PR #1255, opened 2023-05-07
Solution

VolodymyrPst:


src/formatDate.js
```diff
@@ -51,6 +51,53 @@
...
+  const separator = toFormat[toFormat.length - 1];
+  const formatOfDate = [];
+  const obj = {};
+  let str = '';
+
+  for (let i = 0; i < fromFormat.length; i++) {
+    const element = fromFormat[i];
+
+    obj[element] = dateArr[i];
+    formatOfDate.push(obj);
```
vlad-demchuk, line 64: why do you need formatOfDate array? you don't use it

vlad-demchuk (changes requested):

PR #1260, opened 2023-05-08
create formatDate function

pos1bl:


src/formatDate.js
```diff
@@ -50,7 +50,47 @@
...
+  const newDate = date.split(`${fromFormat[3]}`);
+  const result = [];
+  
+  const currentData = (data) => {
+    return newDate[fromFormat.indexOf(data)];
+  };
+
+  for (const value of toFormat) {
+    switch (value) {
+      case 'DD':
```
OleziO, line 62: Move all cases to the object.

OleziO (approved):

PR #1266, opened 2023-05-11
Solution

NickBrodyuk:


src/formatDate.js
```diff
@@ -50,7 +50,36 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromFormatSymb = fromFormat.pop();
```
volodymyr-soltys97, line 53: Avoid modifying the input parameters, create a new variable to store the last element of 'fromFormat' and 'toFormat' arrays.

src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+  const toFormatSymb = toFormat.pop();
+
+  const dateSplit = date.split(fromFormatSymb);
+
+  for (let i = 0; i < fromFormat.length; i++) {
+    for (let k = 0; k < toFormat.length; k++) {
+      switch (true) {
```
volodymyr-soltys97, line 60: Using a switch statement with 'true' as a condition is not recommended. Use if, else if, and else blocks instead for better readability.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more readable

vlad-demchuk (approved):

PR #1273, opened 2023-05-14
[ym] format date advanced

YeMyr:


src/formatDate.js
```diff
@@ -50,7 +50,82 @@
...
+    get date() {
+      return fromFormat.indexOf('DD');
+    },
+
+    get month() {
+      return fromFormat.indexOf('MM');
+    },
+
+    get year() {
+      return 3 - this.date - this.month;
```
naziamoff, line 67: 3 looks like a magic number here. Let's move to constants

naziamoff (approved):
interesting approach 👍

src/formatDate.js
YeMyr: @naziamoff 👍 
Hi! 
I've added it to objects properties, please, check

PR #1274, opened 2023-05-15
Solution

amashutikov:


naziamoff (changes requested):
awesome 🔥 
But let's now try to avoid nested loops. Nested loops are evil 😈 🙅‍♀️

33Orange (approved):
Well done

PR #1278, opened 2023-05-15
format date

DenisGurskiy:


src/formatDate.js
```diff
@@ -51,6 +51,26 @@
...
+  const result = [];
+
+  result[toFormat.indexOf('DD')] = arrDate[fromFormat.indexOf('DD')];
+  result[toFormat.indexOf('MM')] = arrDate[fromFormat.indexOf('MM')];
+
+  let year = (fromFormat.indexOf('YYYY') === -1)
+    ? arrDate[fromFormat.indexOf('YY')]
+    : arrDate[fromFormat.indexOf('YYYY')];
+
+  if (toFormat.indexOf('YYYY') === -1) {
```
vlad-demchuk, line 65: To check if 'YYYY' is present in 'toFormat', use 'includes' method instead of checking for index.

volodymyr-soltys97 (approved):
Great 👍

PR #1280, opened 2023-05-16
Solution

myplekan:


src/formatDate.js
```diff
@@ -50,7 +50,35 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparator = fromFormat[fromFormat.length - 1];
```
volodymyr-soltys97, line 53: Use destructuring to extract values from `fromFormat` and `toFormat` arrays.

src/formatDate.js
```diff
@@ -50,7 +50,35 @@
...
+  let newDate = date.split(fromSeparator);
```
volodymyr-soltys97, line 54: Use `const` instead of `let` for `newDate` variable since it's not reassigned.

src/formatDate.js
```diff
@@ -50,7 +50,35 @@
...
+          arr[i] = `19${newDate[j]}`;
+        }
+      }
+
+      if (toFormat[i] === fromFormat[j]) {
+        arr[i] = newDate[j];
+      }
+    }
+  }
+  newDate = arr.join(toSeparator);
```
volodymyr-soltys97, line 79: Use a different variable name instead of reassigning `newDate`, you can directly return the value.

src/formatDate.js
```diff
@@ -50,7 +50,35 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparator = fromFormat[fromFormat.length - 1];
+  let newDate = date.split(fromSeparator);
+  const toSeparator = toFormat[toFormat.length - 1];
+  const arr = [];
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    for (let j = 0; j < fromFormat.length - 1; j++) {
+      if (toFormat[i].includes('Y') && fromFormat[i].includes('Y')) {
```
volodymyr-soltys97, line 60: You have a typo in your condition, it should be `fromFormat[j]` instead of `fromFormat[i]`.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more sophisticated

mykhalenych (approved):

PR #1286, opened 2023-05-18
Solution

Natalia-Tsiutsiura:


src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
-  // write code here
-}
+  const partToConvert = toFormat.slice(-1);
+  const partToFormat = toFormat.slice(0, 3);
+  const partfromFormat = fromFormat.slice(0, 3);
+  const splitDate = date.split(fromFormat[3]);
+  const arr = [];
+  const objDate = {};
+
+  for (const part of splitDate) {
```
IvanFesenko, line 60: This loop is unnecessary as you can directly assign splitDate to arr using Array.from.

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
+
+  if (objDate.hasOwnProperty('YYYY')) {
+    objDate.YY = objDate.YYYY.slice(2);
+  }
 
+  if (objDate.YY < 30) {
+    objDate.YYYY = '20' + objDate.YY;
+  } else {
+    objDate.YYYY = '19' + objDate.YY;
+  }
```
IvanFesenko, line 76: use trenary instead of else statement

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
+
+  const formattedDate = partToFormat.map((item) => objDate[item]);
+  const newDate = formattedDate.join(' ').trim().split(' ').join(partToConvert);
+
+  return newDate;
```
IvanFesenko, line 81: No need in extra variable
```suggestion
  return formattedDate.join(' ').trim().split(' ').join(partToConvert);
```

IvanFesenko (changes requested):
Check my comments

mykhalenych (approved):

PR #1290, opened 2023-05-20
add task solution

KatyaZaremska:


src/formatDate.js
```diff
@@ -50,7 +50,51 @@
...
+      case 'YYYY':
+        newDate.push(year);
+        break;
+      case 'YY':
+        newDate.push(year.slice(-2));
+        break;
+    }
+  }
+
+  return newDate.join(toFormat[3]).toString();
```
33Orange, line 97: The `toString()` method is not necessary as `join()` already returns a string.

33Orange (approved):
Well done ! 🔥

33Orange (approved):

PR #1293, opened 2023-05-20
Solution

maryanmats:


src/formatDate.js
```diff
@@ -50,7 +50,30 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newDate = [];
+  const dateInfo = {};
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    dateInfo[fromFormat[i]] = date.split(fromFormat[fromFormat.length - 1])[i];
```
silchyn, line 57: that is not readable

silchyn (changes requested):
Think about simpler and easier to read solution

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newDate = [];
+  const dateInfo = {};
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    const formatKey = fromFormat[i];
+
+    dateInfo[formatKey] = date.split(fromFormat[fromFormat.length - 1])[i];
```
33Orange, line 59: Check @silchyn comment about readability.

Split date above your for loop.

33Orange (changes requested):
Almost done.

src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+    dateInfo.YY = dateInfo.YYYY.slice(2);
+  }
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    const formatKey = toFormat[i];
+
+    newDate.push(dateInfo[formatKey]);
+  }
+
+  return newDate.join(toFormat[toFormat.length - 1]);
```
33Orange, line 82:
```suggestion
  return newDate.join(toSeparator);
```

33Orange (approved):
Good job.

Pay attention to the comments. This will increase the understanding of the code for a person who has not seen it.

PR #1294, opened 2023-05-20
Solution

MykolaLopushenko:


src/formatDate.js
```diff
@@ -50,7 +50,33 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const partOfDate = date.split(`${fromFormat[3]}`);
```
etojeDenys, line 53: Avoid using template literals for a single variable. Use 'fromFormat[3]' directly.

src/formatDate.js
```diff
@@ -50,7 +50,33 @@
...
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    for (const part in dateObj) {
+      if (toFormat[i] === part) {
+        result.push(dateObj[part]);
+      }
+    }
+  }
+
+  return result.join(`${toFormat[3]}`);
```
etojeDenys, line 79: Avoid using template literals for a single variable. Use 'toFormat[3]' directly.

etojeDenys (changes requested):

src/formatDate.js
MykolaLopushenko: done

mykhalenych (approved):

PR #1295, opened 2023-05-21
add solution

dimarogkov:


src/formatDate.js
```diff
@@ -51,6 +52,63 @@
...
   // write code here
+  const res = [];
+  const currentYear = new Date().getFullYear() % 100;
+  const dateArr = date.split(`${fromFormat[fromFormat.length - 1]}`);
+  const dateObg = {
+    year: null,
+    month: null,
+    day: null,
+    separator: null,
+  };
```
VitaliyBondarenko1982, line 63:
```suggestion
  const dateObg = {};
```
looks like redundant to make each key null. we can start with empty object

VitaliyBondarenko1982 (approved):

PR #1303, opened 2023-05-24
Solution

reseneweb:


qoa11a (changes requested):
formatDate("97/02/18", ["YY", "MM", "DD", "/"], ["DD", "MM", "YYYY", "."]); // '18.02.1997'
This one doesn't work in your case. Your code gives 18.02.2018 as an answer. Fix it, please.

agniya-i (approved):

PR #1315, opened 2023-05-30
solution

vlad-fil:


src/formatDate.js
```diff
@@ -50,7 +50,29 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateToArr = date.split(fromFormat[3]);
+
+  const obj = {};
+
+  const newFormat = [];
+
+  for (let i = 0; i < 3; i++) {
```
VitaliyBondarenko1982, line 59:
```suggestion
  for (let i = 0; i < fromFormat.length; i++) {
```
looks like it can be not magic number

VitaliyBondarenko1982 (approved):

PR #1317, opened 2023-05-31
SolutionFormatDate

olena-bielozorova:


src/formatDate.js
```diff
@@ -50,7 +50,25 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
```
Julia14012000, line 52: Use default parameters to set default values for function arguments

src/formatDate.js
```diff
@@ -50,7 +50,25 @@
...
-  // write code here
+  const parts = date.split(fromFormat[fromFormat.length - 1]);
```
Julia14012000, line 53: Use destructuring to extract values from objects and arrays

src/formatDate.js
```diff
@@ -50,7 +50,25 @@
...
+    if (format === 'YYYY' && year) {
+      formattedParts.push(yearPrefix + year);
+    } else if (format === 'YY' && parts[fromFormat.indexOf('YYYY')]) {
+      formattedParts.push(parts[fromFormat.indexOf('YYYY')].slice(2));
+    } else {
+      formattedParts.push(value);
+    }
+  }
+
+  return formattedParts.join(toFormat[toFormat.length - 1]);
```
Julia14012000, line 71: Use template strings instead of concatenation

Julia14012000 (approved):

PR #1319, opened 2023-06-01
Solution

andrii-rykal:
I understand that the code is too much, but I still don't understand how to reduce it


VitaliyBondarenko1982 (approved):
looks good.
of coourse you can reduce amount of code for example to make other approach:
from array fromFormat make object, formattedDate and work with them.
will be something like that
Screenshot from 2023-06-01 14-07-00

PR #1326, opened 2023-06-09
solution

hrynkevych:


src/formatDate.js
```diff
@@ -1,56 +1,30 @@
...
+
+  if (!dateInfo.YYYY && +dateInfo.YY >= 30) {
+    dateInfo.YYYY = `19${dateInfo.YY}`;
+  }
+
+  if (!dateInfo.YYYY && +dateInfo.YY < 30) {
+    dateInfo.YYYY = `20${dateInfo.YY}`;
+  }
+
+  if (!dateInfo.YY && +dateInfo.YYYY >= 30) {
```
mykhalenych, line 19: This check is unnecessary as the YYYY value will always be 4 digits. Remove this condition.

src/formatDate.js
```diff
@@ -1,56 +1,30 @@
...
- *
- * @returns {string}
- */
-
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newDate = [];
+  const dateInfo = {};
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
```
mykhalenych, line 7: Consider using destructuring to get the separator and date parts from 'fromFormat' and 'toFormat'.

src/formatDate.js
```diff
@@ -1,56 +1,30 @@
...
+    dateInfo[fromFormat[i]] = date.split(fromFormat[fromFormat.length - 1])[i];
```
mykhalenych, line 8: Avoid calling 'split' multiple times in the loop. Call it once before the loop and store the result.

mykhalenych (changes requested):

mykhalenych (approved):

PR #1327, opened 2023-06-09
Solution formatDate Adv

oleh-dehtiarenko:


src/formatDate.js
```diff
@@ -50,7 +50,62 @@
...
+      YY = oldDateArr[i];
+    }
+
+    if (fromFormat[i] === 'MM') {
+      MM = oldDateArr[i];
+    }
+
+    if (fromFormat[i] === 'DD') {
+      DD = oldDateArr[i];
+    }
```
OleziO, line 75: Replace with switch case construction to make your code cleaner.

src/formatDate.js
```diff
@@ -50,7 +50,62 @@
...
+
+    if (i === 'MM') {
+      newDate.push(MM);
+    }
+
+    if (i === 'DD') {
+      newDate.push(DD);
+    }
+  }
```
OleziO, line 106: Same as previous comment

OleziO (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,71 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldDateArr = date.split(fromFormat[3]);
+  let YYYY = 0;
+  let MM = 0;
+  let DD = 0;
+  let YY = 0;
```
ivbutsykin, line 57: Such values better to store in object

src/formatDate.js
```diff
@@ -50,7 +50,71 @@
...
+
+      case 'MM':
+        MM = oldDateArr[i];
+        break;
+
+      case 'DD':
+        DD = oldDateArr[i];
+        break;
+
+      default:
```
ivbutsykin, line 78: Imagine if you have object (for ex. `parsedDate`) that stores your `YY`, `MM`, `DD`. So you can avoid switch/case and simple write:
```
parsedDate[fromFormat[i] = oldDateArr[i]
```

src/formatDate.js
```diff
@@ -50,7 +50,71 @@
...
+        newDate.push(MM);
+        break;
+
+      case 'DD':
+        newDate.push(DD);
+        break;
+
+      default:
+        break;
+    }
```
ivbutsykin, line 114: This also can be avoided with object:
```
newData.push(parsedDate[toFormat[i]]);
```

ivbutsykin (approved):
Fine, but I recommend you to take into account my suggestions

src/formatDate.js
oleh-dehtiarenko: Ok, thank you, I'll keep that in mind for the future)

PR #1330, opened 2023-06-11
found solution

Bonum-taurus:


src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+  const newDate = [];
+
+  for (const param of toFormat) {
+    newDate.push(objDate[param]);
+  }
+
+  // delete last element(undefined).
+  newDate.pop();
+
+  const newSeparator = toFormat[3];
```
volodymyr-soltys97, line 83: Consider using a more descriptive variable name like 'formattedDate' instead of 'updateDate'.

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1334, opened 2023-06-12
add task solution

MaxSchmide:


src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
+        }
+
+        break;
+      case 'M':
+        newDateArr.push(currentMonth);
+        break;
+      case 'D':
+        newDateArr.push(currentDay);
+        break;
+    }
```
ilona-borkovska, line 95: image

ilona-borkovska (approved):

src/formatDate.js
```diff
@@ -50,7 +50,56 @@
...
-  // write code here
+  const oldSeparator = fromFormat.pop();
+  const newSeparator = toFormat.pop();
+  let currentYear = '';
+  let currentMonth = '';
+  let currentDay = '';
+  const oldDateArr = date.split(oldSeparator);
+  const newDateArr = [];
+
+  fromFormat.map((el, index) => {
```
anstsot, line 61: `map` method should always return value for each iteration. here would be better to use `forEach`

src/formatDate.js
```diff
@@ -50,7 +50,56 @@
...
+        break;
+      case 'D':
+        currentDay = oldDateArr[index];
+        break;
+      default:
+        throw new Error('Invalid format');
+    }
+  });
+
+  toFormat.map((el) => {
```
anstsot, line 77: same here. use `forEach` or consider using `for ... of`

src/formatDate.js
```diff
@@ -50,7 +50,56 @@
...
+      case ('Y'):
+        if (el.length > currentYear.length) {
+          const year = (+currentYear < 30 ? '20' : '19') + currentYear;
+
+          newDateArr.push(year);
+        } else if (el.length < currentYear.length) {
+          newDateArr.push(currentYear.slice(-2));
+        } else {
+          newDateArr.push(currentYear);
+        }
```
anstsot, line 88: you can modify `currentYear` variable in `if ... else` constructions. and after that push it to `newDateArr`

anstsot (changes requested):
Almost done :+1:
Let’s add a few fixes to improve your code.

anstsot (approved):
Good work, approving:sparkles:

PR #1342, opened 2023-06-13
Solution

yehor-kor:


src/formatDate.js
```diff
@@ -50,7 +50,48 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateList = date.split(fromFormat[3]);
```
polosanya, line 53: Use destructuring to get day, month, and year from dateList array.

src/formatDate.js
```diff
@@ -50,7 +50,48 @@
...
+
+  const newDateList = [];
+
+  for (const format of toFormat) {
+    if (format === 'DD') {
+      newDateList.push(day);
+    } else if (format === 'MM') {
+      newDateList.push(month);
+    } else if (format === 'YY' || format === 'YYYY') {
+      newDateList.push(String(year));
```
polosanya, line 90: It is better to use switch...case when you have a limited amount of possible values
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

polosanya (changes requested):
Good job! Let's improve your code a bit

src/formatDate.js
yehor-kor: I have no idea how correctly write this, because the order is not the same

src/formatDate.js
yehor-kor: Done, good remark

agniya-i (approved):

PR #1343, opened 2023-06-13
create

LeoSpartaDeV:


src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+
+  for (let i = 0; i < arrData.length; i++) {
+    objDate[oldFormat[i]] = arrData[i];
+  }
+
+  if ('YYYY' in objDate) {
+    objDate.YY = objDate.YYYY.slice(-2);
+  }
+
+  if ('YY' in objDate && newFormat.includes('YYYY')) {
```
DarkMistyRoom, line 69: there may be a case when `oldFormat` contains 'YYYY', you add 'YY' in the previous condition and when the program is on this line, we already have both properties in the objDate. In this case, converting 'YY' to 'YYYY' will be redundant as we already have 'YYYY'

DarkMistyRoom (changes requested):
Good job! Let's handle one more case

OleziO (approved):

PR #1346, opened 2023-06-14
Solution

irinabolbat:


IraShtynda (approved):
The code works fine, and I approve it. 🌺 However, it was challenging for me to understand some parts of it.

src/formatDate.js
```diff
@@ -50,7 +50,60 @@
...
+        month = (dateParts[i].length === 1) ? `0${+dateParts[i]}` : dateParts[i];
+        break;
+
+      case 'DD':
+        day = (dateParts[i].length === 1) ? `0${+dateParts[i]}` : dateParts[i];
+        break;
+    }
+  }
+
+  const currentDate = new Date(year, month - 1, day);
```
anstsot, line 79: is it necessary to create the real date from parts?

anstsot (changes requested):
try to reduce and simplify the code a bit by removing the `new Date` creation and working with `fromFormat` parts directly

tetiana-ivanchenko (approved):

PR #1348, opened 2023-06-14
Solution

MelnychukVol:


src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+
+    if (index !== toFormat.length - 2) {
+      result += newSeparator;
+    }
+  }
+
+  return result;
+}
+
+function convertingYear(newFormat, year) {
```
volodymyr-soltys97, line 87: Function names should be in camelCase. Rename 'convertingYear' to 'convertYear'.

vlad-demchuk (approved):

PR #1356, opened 2023-06-17
solution

zuzu6:


src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newArr = date.split(fromFormat[3]);
+  const resultDate = [];
+  let dateValue = 0;
+  let monthValue = 0;
+  let yearValue = 0;
+
+  for (let index = 0; index < fromFormat.length; index++) {
+    if (fromFormat[index] === 'YYYY' || fromFormat[index] === 'YY') {
```
Anastasiia-Svintsova, line 60: Use switch/case instead of a lot of `if` to improve the readability

src/formatDate.js
```diff
@@ -50,7 +50,53 @@
...
+
+    if (fromFormat[index] === 'DD') {
+      dateValue = newArr[index];
+    }
+  }
+
+  for (const element of toFormat) {
+    if (element === 'YYYY' || element === 'YY') {
+      if (element.length === 2) {
+        if (yearValue.length === 2) {
```
Anastasiia-Svintsova, line 76: A lot of nesting here. Think about how to refactor it to avoid it

Anastasiia-Svintsova (changes requested):

PR #1362, opened 2023-06-18
Solution

yurii-krymskiy:


src/formatDate.js
```diff
@@ -50,7 +50,33 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let result = '';
+  const dateParts = date.split(fromFormat[fromFormat.length - 1]);
```
volodymyr-soltys97, line 54: Consider using destructuring to assign date parts to the newDate object directly.

src/formatDate.js
```diff
@@ -50,7 +50,33 @@
...
+    newDate.YY = newDate.YYYY.slice(-2);
+  }
+
+  if (newDate.YY < 30) {
+    newDate.YYYY = `20${newDate.YY}`;
+  } else {
+    newDate.YYYY = `19${newDate.YY}`;
+  };
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
```
volodymyr-soltys97, line 71: Consider using 'for of' loop with entries() method instead of 'for' loop to iterate over toFormat array.

volodymyr-soltys97 (approved):
Well done 👍

PR #1363, opened 2023-06-19
Solution

dipinchuk:


src/formatDate.js
```diff
@@ -50,7 +50,44 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const result = date.split(/,|-|\.|\//);
```
anstsot, line 53: it will be better not to hardcode all possible separators for date but use the actual one from `fromFormat`: `fromFormat[fromFormat.length - 1]`

anstsot (approved):
Good work, approving:sparkles:

PR #1364, opened 2023-06-19
Function formatted Date was created

Yuliia-Hrinchenko1:


src/formatDate.js
```diff
@@ -50,7 +50,37 @@
...
+    const datePart = toFormat[i];
+
+    formattedDate.push(dateParts[datePart]);
+
+    if (i !== toFormat.length - 1) {
+      formattedDate.push(newSeparator);
+    }
+  }
+
+  return formattedDate.join('').slice(0, -1);
```
volodymyr-soltys97, line 84: It's better to use `formattedDate.join(newSeparator)` instead of slicing the last character.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

volodymyr-soltys97 (approved):
Well done 🔥

PR #1365, opened 2023-06-19
dateTask

zipic:


src/formatDate.js
```diff
@@ -50,7 +50,44 @@
...
+    } else {
+      result.push(object[format]);
+    }
+  }
+
+  return result.join(separator);
+}
+
+function convertedYear(year) {
+  if (year.length === 4) {
```
loralevitska, line 86: Move magic numbers to constant (about 4). Fix all cases if present

loralevitska (approved):
Good job!

PR #1367, opened 2023-06-20
Solution

Nikita211112:
"I learned that I need to use objects in Slack, and it was very difficult for me to complete this task. Is there an alternative solution to this task without using objects?"


sTorba24 (approved):
Good work! Your solution seems good

PR #1370, opened 2023-06-22
Solution

Luk2asz:


Radoslaw-Czerniawski (approved):
Very good implementation, left some comments with ideas how your code might be improved. Only thing that I see explicitly wrong is unnecessary declaration of `resultString` instead of just:
```js
return resultArray.join(toFormat[3]);
```

PR #1371, opened 2023-06-22
add task js format date advanced

Lazarend:


src/formatDate.js
```diff
@@ -50,7 +50,34 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateArray = date.split(fromFormat[fromFormat.length - 1]);
+  // eslint-disable-next-line max-len
```
loralevitska, line 54: Try to fix eslint error and remove comment

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
+  const object = Object.fromEntries(fromFormat.slice(0, -1).map((format, index) => [format, dateArray[index]]));
```
loralevitska, line 55: Something like that)
```suggestion
  const object = Object.fromEntries(fromFormat.slice(0, -1).map(
    (format, index) => [format, dateArray[index]])
  );
```

loralevitska (approved):
Good job!

PR #1372, opened 2023-06-24
Solution

Erchu420:


src/formatDate.js
```diff
@@ -50,7 +50,24 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const from = Array.from(fromFormat);
```
mvjl000, line 53: You are only using this in for declaration. I think you should add the pop method here and rename it ;) 
```suggestion
  const oldSeparator = Array.from(fromFormat).pop();
```

src/formatDate.js
```diff
@@ -50,7 +50,24 @@
...
+  const to = Array.from(toFormat);
+  const sep = to.pop();
```
mvjl000, line 55:
```suggestion
  const newSeparator = to.pop();
```

src/formatDate.js
```diff
@@ -50,7 +50,24 @@
...
+  const tempOBJ = {};
```
mvjl000, line 56: I would change the name of this constant as well so it is clear what is it for.

mvjl000 (changes requested):
I would like to rename your constants so it's more self-explanatory. Besides of that - really good job! 🚀

src/formatDate.js
Erchu420: done :)

src/formatDate.js
Erchu420: done :)

src/formatDate.js
Erchu420: done :)

PR #1378, opened 2023-06-30
Solution

Sirius9312:
Format Date Task

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -50,7 +50,32 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateArr = date.split(fromFormat[3]);
+  const dateObj = {};
+  const result = [];
+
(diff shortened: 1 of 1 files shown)


src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
+    if (fromFormat[i] === 'YY') {
+      if (dateArr[i] < '30') {
+        dateArr[i] = '20'.concat(dateArr[i]);
+      } else {
+        dateArr[i] = '19'.concat(dateArr[i]);
+      }
+      fromFormat[i] = 'YYYY';
+    }
+
+    Object.assign(dateObj, { [fromFormat[i]]: dateArr[i] });
```
volodymyr-soltys97, line 67: Instead of using Object.assign for a single new property, you can use simple assignment. For example: dateObj[fromFormat[i]] = dateArr[i].

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1381, opened 2023-07-01
first commit

Faridun92:


src/formatDate.js
```diff
@@ -50,7 +50,50 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newSeparator = toFormat[toFormat.length - 1];
+  const oldSeparator = fromFormat[fromFormat.length - 1];
+  const newFormat = toFormat;
```
etojeDenys, line 55: The 'newFormat' variable is redundant and unnecessary. You are creating a new reference but not changing the original array 'toFormat'. You can use 'toFormat' directly in your code.

src/formatDate.js
```diff
@@ -50,7 +50,50 @@
...
+  if (fromFormat.includes('YY')) {
+    const altYear = parseInt(workDate[fromFormat.indexOf('YY')], 10);
+
+    year = altYear < 30 ? 2000 + altYear : 1900 + altYear;
+  }
+
+  const month = parseInt(workDate[fromFormat.indexOf('MM')], 10) - 1;
+  const day = parseInt(workDate[fromFormat.indexOf('DD')], 10);
+
+  const parseDate = new Date(year, month, day);
```
etojeDenys, line 69: Here you create a Date object 'parseDate', but you only use it to get the last two digits of the year. You already have the full year in the 'year' variable. You can get the last two digits of the year from the 'year' variable without creating a new Date object.

etojeDenys (changes requested):

loralevitska (approved):
Well done)

PR #1383, opened 2023-07-02
add task solution

NikitaOrliak:


src/formatDate.js
```diff
@@ -50,7 +50,85 @@
...
+
+      case 'DD': {
+        day = fromArray[i];
+
+        break;
+      }
+    }
+  }
+
+  if (year) {
```
volodymyr-soltys97, line 89: In this condition, if year is an empty string (''), the condition will return false although it's a valid value. Consider checking against undefined or null.

src/formatDate.js
```diff
@@ -50,7 +50,85 @@
...
+    smallYear = year.slice(2);
+  }
+
+  if (smallYear) {
+    if (smallYear < 30) {
+      year = 20 + smallYear;
```
volodymyr-soltys97, line 95: Concatenation of numbers and strings can lead to unexpected results. '20' + smallYear will result in a string, not a number.

src/formatDate.js
```diff
@@ -50,7 +50,85 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const [ , , , fromSeparator ] = fromFormat;
```
volodymyr-soltys97, line 53: Possible destructuring of an array can be done here in a more clean way.

src/formatDate.js
```diff
@@ -50,7 +50,85 @@
...
+  const [ , , , toSeparator ] = toFormat;
```
volodymyr-soltys97, line 54: Possible destructuring of an array can be done here in a more clean way.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

src/formatDate.js
```diff
@@ -50,7 +50,85 @@
...
+  const toSeparator = toFormat[toFormat.length - 1];
+  let year, smallYear, month, day;
+  const resultArray = [];
+  let resultDate = '';
+
+  const fromArray = date.split(fromSeparator);
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    switch (fromFormat[i]) {
+      case 'YYYY': {
```
anastasiiavorobiova, line 63: Consider using constants instead of magic strings here

src/formatDate.js
```diff
@@ -50,7 +50,85 @@
...
+    }
+  }
+
+  if (year !== undefined && year !== null) {
+    smallYear = year.slice(2);
+  }
+
+  if (smallYear !== undefined && smallYear !== null) {
+    if (smallYear < 30) {
+      year = 2000 + (+smallYear);
```
anastasiiavorobiova, line 95:
```suggestion
      year = 2000 + Number(smallYear);
```
This suggestion is a bit more readable and convenient to maintain

src/formatDate.js
```diff
@@ -50,7 +50,85 @@
...
+    } else {
+      year = 1900 + (+smallYear);
```
anastasiiavorobiova, line 97: See my previous comment

anastasiiavorobiova (changes requested):
Great solution! Added tiny comments to fix

PR #1384, opened 2023-07-03
solution

tndeineka:


src/formatDate.js
```diff
@@ -50,7 +50,74 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldSeparator = fromFormat.pop();
+  const newSeparator = toFormat.pop();
```
DarkMistyRoom, line 54: don't mutate the input data

src/formatDate.js
```diff
@@ -50,7 +50,74 @@
...
+
+      default:
+        break;
+    }
+  }
+
+  for (let i = 0; i < toFormat.length; i++) {
+    switch (toFormat[i]) {
+      case ('YY'):
+        newYearFormat = 2;
```
DarkMistyRoom, line 88: why do you need `newYearFormat` variable?

src/formatDate.js
```diff
@@ -50,7 +50,74 @@
...
+
+        if (oldYearFormat > newYearFormat) {
```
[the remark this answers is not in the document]
DarkMistyRoom: it will give the same result

src/formatDate.js
```diff
@@ -50,7 +50,74 @@
...
+      case ('YYYY'):
+        newYearFormat = 4;
+
+        if (oldYearFormat < newYearFormat && year < 30) {
+          year = '20' + year;
+        } else if (oldYearFormat < newYearFormat && year >= 30) {
+          year = '19' + year;
+        }
+        result.push(year);
+        break;
```
DarkMistyRoom, line 105:
```suggestion
      case ('YYYY'):
        if (oldYearFormat === 2 && year < 30) {
          year = '20' + year;
        } else if (oldYearFormat === 2 && year >= 30) {
          year = '19' + year;
        }
        result.push(year);
        break;
```

DarkMistyRoom (changes requested):
Good job! Pay attention to the comments:

src/formatDate.js
tndeineka: In one of the videos, it was said to limit the use of numeric data without variables, so that later we can know what the number is and where it came from. That's why I thought it would be better)

src/formatDate.js
```diff
@@ -50,7 +50,70 @@
...
+      case ('DD'):
+        day = arrDate[i];
+        break;
+
+      default:
+        break;
+    }
+  }
+
+  for (let i = 0; i < toFormat.length; i++) {
```
Anastasiia-Svintsova, line 85: It's better to use the map() method to create the result array, which simplifies the code and makes it more readable.
```
const result = toFormat.map((format) => {...

File: src/formatDate.js [hunk 6]
```diff
@@ -50,7 +50,70 @@
...
+  const arrDate = date.split(oldSeparator);
+
+  const result = [];
+  let year, month, day, oldYearFormat;
+
+  for (let i = 0; i < fromFormat.length; i++) {
+    switch (fromFormat[i]) {
+      case ('YY'):
+        oldYearFormat = 2;
+        year = arrDate[i];
```
Anastasiia-Svintsova, line 64: Look, you repeat the `arrDate[i]` value not once. It's better to move it to a separate variable to improve code readability and reduces redundancy.

Anastasiia-Svintsova (approved):
Good job! Pay attention to my comments, please

PR #1385, opened 2023-07-04
Solution

misha-shevchuk:


src/formatDate.js
```diff
@@ -50,7 +50,46 @@
...
+    if (fromFormat[i] === 'MM') {
+      month = dateData[i];
+    }
+
+    if (fromFormat[i] === 'DD') {
+      day = dateData[i];
+    }
+  }
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
```
volodymyr-soltys97, line 74: Same as above, replacing 'if' statements with a 'switch' statement would improve readability and performance. Also consider using a map or object structure for 'year', 'month', and 'day' variables.

volodymyr-soltys97 (approved):
Nice work 👍

PR #1388, opened 2023-07-06
solution

NataliMax:


src/formatDate.js
```diff
@@ -50,7 +50,35 @@
...
+  || toFormat[toFormat.indexOf('YY')];
+
+  if (newYears.length < years.length) {
+    years = years.slice(-2);
+  } else if (newYears.length > years.length) {
+    if (+years < 30) {
+      century = '20';
+    } else if (+years >= 30) {
+      century = '19';
+    }
```
Anastasiia-Svintsova, line 71: Use ternary operator instead of nesting `if`
```suggestion
    century = +years < 30 ? '20' : '19;
```

src/formatDate.js
```diff
@@ -50,7 +50,35 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldForm = [...fromFormat];
```
Anastasiia-Svintsova, line 53: Use a more descriptive variable name
```suggestion
  const oldFormat = [...fromFormat];
```

Anastasiia-Svintsova (approved):

PR #1390, opened 2023-07-06
Solution

GeorgeSavostikov:


src/formatDate.js
```diff
@@ -50,7 +50,98 @@
...
-  // write code here
+  const splitedDate = date.split(fromFormat[3]);
+  let year = '';
+  let month = '';
+  let day = '';
+
+  for (let i = 0; i < toFormat.length; i++) {
+    switch (fromFormat[i]) {
+      case 'YYYY':
+        year = splitedDate[i];
```
TarasHoliuk, line 61: instead of saving it to separate variables, create an object and save values under the same keys, e.g.
`{ YYYY: 2022, MM: 10, DD: 12 }`
in this case, in the next steps, you will know which format you have saved

src/formatDate.js
```diff
@@ -50,7 +50,98 @@
...
+
+      case 'DD':
+        day = splitedDate[i];
+        break;
+    }
+  }
+
+  const result = [];
+
+  const correctYear = transformYear(year, fromFormat, toFormat);
```
TarasHoliuk, line 80: `transformYear` function is too complicated
you may check which year format you have at the moment and just add another one to your object

TarasHoliuk (changes requested):

SanyaBratashchuk (approved):

PR #1391, opened 2023-07-06
solution

SashaKozoriz:


src/formatDate.js
```diff
@@ -50,7 +50,67 @@
...
+    switch (true) {
+      case toFormat[i] === "DD":
+        reformattedDate.push(dateInfo.day);
+        break;
+
+      case toFormat[i] === "MM":
+        reformattedDate.push(dateInfo.month);
+        break;
+
+      case toFormat[i] === "YY" && dateInfo.shortYear < 30:
```
mykhalenych, line 103: The condition toFormat[i] === 'YY' && dateInfo.shortYear < 30 doesn't seem to be necessary. Only toFormat[i] === 'YY' condition should be enough as the year is already formatted in the previous loop.

mykhalenych (approved):

PR #1392, opened 2023-07-07
Formated Date

dimkamg21:


src/formatDate.js
```diff
@@ -50,7 +50,28 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const [, , , oldSeparator] = fromFormat;
+  const [, , , newSeparator] = toFormat;
+
```
IvanFesenko, line 55: for this case instead of using destruction - better get value by index
```suggestion
  const oldSeparator = fromFormat[3];
  const newSeparator = toFormat[3];

```

IvanFesenko (approved):

PR #1394, opened 2023-07-07
create a function formatDate

Solovei-Andrii:


src/formatDate.js
```diff
@@ -50,7 +50,29 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const objValueOfDate = {};
+  const arrDateData = date.split(fromFormat[3]);
```
volodymyr-soltys97, line 54: Consider destructuring arrDateData directly from date.split for cleaner code. For example: const [day, month, year] = date.split(fromFormat[3]);

src/formatDate.js
```diff
@@ -50,7 +50,29 @@
...
+  const newArrDate = [];
```
volodymyr-soltys97, line 55: The variable newArrDate is not descriptive. Consider renaming it to something more meaningful, like 'formattedDateParts'.

src/formatDate.js
```diff
@@ -50,7 +50,29 @@
...
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    objValueOfDate[fromFormat[i]] = arrDateData[i];
+  }
+
+  if (toFormat.includes('YYYY') && !('YYYY' in objValueOfDate)) {
+    objValueOfDate['YYYY'] = objValueOfDate['YY'] < 30
+      ? 2000 + Number(objValueOfDate['YY'])
```
volodymyr-soltys97, line 63: You are using the Number constructor to convert a string into a number. Consider using the unary plus operator (+) for this conversion as it is more concise and faster.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

src/formatDate.js
Solovei-Andrii: but I don't understand how I can use destructuring in this case

PR #1406, opened 2023-07-15
Solution

katiailchenko:


src/formatDate.js
```diff
@@ -50,7 +50,61 @@
...
+      newDateArray.push(`20${year}`);
+    };
+
+    if (element === 'YYYY' && year.length === 2 && year >= 30) {
+      newDateArray.push(`19${year}`);
+    };
+  };
+
+  // eslint-disable-next-line no-console
+  console.log(arrayFromDate, separator);
```
ivadimko, line 105: Avoid using console.log in production code. It is usually used for debugging and should be removed in the final version of the code.

src/formatDate.js
```diff
@@ -50,7 +50,61 @@
...
+  for (const element of toFormat) {
+    if (element === 'DD') {
+      newDateArray.push(day);
+    };
+
+    if (element === 'MM') {
+      newDateArray.push(month);
+    };
+
+    if (
```
ivadimko, line 84: The conditions here can be simplified and the use of multiple 'if' statements can be avoided. Consider using a switch statement or an object as a map for the format.

ivadimko (approved):

PR #1415, opened 2023-07-21
Solution

NadieinOleh:


src/formatDate.js
```diff
@@ -50,7 +50,27 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const objectDate = {};
+  const separatorToFormat = toFormat[toFormat.length - 1];
+  const separatorFromFormat = fromFormat[toFormat.length - 1];
+  const arrayDate = date.split(separatorFromFormat);
+  const result = [];
+
+  for (let i = 0; i < arrayDate.length; i++) {
```
Anastasiia-Svintsova, line 59: Consider using `arrayDate.forEach` instead of a for loop to fill `objectDate`. This is more readable and avoids potential off-by-one errors.

src/formatDate.js
```diff
@@ -50,7 +50,27 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const objectDate = {};
+  const separatorToFormat = toFormat[toFormat.length - 1];
+  const separatorFromFormat = fromFormat[toFormat.length - 1];
```
Anastasiia-Svintsova, line 55: You are using `toFormat.length` for `separatorFromFormat`. It should be `fromFormat.length` to correctly find the separator in the `fromFormat` string.

src/formatDate.js
```diff
@@ -50,7 +50,27 @@
...
+  const result = [];
+
+  for (let i = 0; i < arrayDate.length; i++) {
+    objectDate[fromFormat[i]] = arrayDate[i];
+  }
+
+  if ('YYYY' in objectDate) {
+    objectDate['YY'] = objectDate['YYYY'].slice(2);
+  } else {
+    objectDate['YYYY'] = objectDate['YY'] < 30 ? `20${objectDate['YY']}` : `19${objectDate['YY']}`;
```
Anastasiia-Svintsova, line 66: You can move the calculation to a separate variable 
```
const prefix = objectDate['YY'] < 30 ? 20 : 19;

Anastasiia-Svintsova (changes requested):

witflash (approved):
Well done 👍

PR #1417, opened 2023-07-23
Develop

Sancz0pansa:


src/formatDate.js
```diff
@@ -50,7 +50,55 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldSeparator = fromFormat[3];
+  const separator = toFormat[3];
+  const arrDate = date.split(oldSeparator);
+  let day;
+  let month;
+  let year;
```
choeqq, line 58: it's a good pracitce to initialize new values with falsy value - string: `""`, number: `0`, object: `{}` etc.

src/formatDate.js
```diff
@@ -50,7 +50,55 @@
...
+    if (toFormat[i] === 'MM') {
+      result[i] = month;
+    }
+
+    if (toFormat[i] === 'YY' || toFormat[i] === 'YYYY') {
+      result[i] = year;
+    }
+  }
+
+  return result.join(separator);
```
choeqq, line 101: I've came up with that solution, you can check if that also works fine 😉 
```suggestion
  const dateParts = date.split(fromFormat[fromFormat.length - 1]);
  const [fromYear, fromMonth, fromDay] = dateParts;

  let toYear = "";
  let toMonth = "";
  let toDay = "";

  for (let i = 0; i < toFormat.length; i++) {
    if (toFormat[i].includes("YY")) {
      // Convert year from 4 digits to 2 digits (YYYY to YY)
      toYear = fromYear.slice(-2);
    } else if (toFormat[i].includes("YYYY")) {
      // Convert year from 2 digits to 4 digits (YY to YYYY)
      const prefix = Number(fromYear.slice(0, 2)) < 30 ? "20" : "19";
      toYear = prefix + fromYear;
    } else if (toFormat[i].includes("MM")) {
      toMonth = fromMonth;
    } else if (toFormat[i].includes("DD")) {
      toDay = fromDay;
    }
  }

  const formattedDate = [toYear, toMonth, toDay].join(toFormat[toFormat.length - 1]);

  return formattedDate;

```

choeqq (changes requested):
Your solution looks really good!

Please take a look at my comments 🚀

src/formatDate.js
Sancz0pansa: Checked it. https://prnt.sc/8B9YxlskI2pz
Something is wrong with extending year length and day/year placement.

choeqq: sure, no problem 😄

choeqq (approved):
Well done ✅

PR #1425, opened 2023-07-25
formated date

ivan-baranovskyi:


src/formatDate.js
```diff
@@ -50,7 +50,42 @@
...
+
+  for (const index of toFormat) {
+    if (index === 'YYYY') {
+      const year = dateToArr[yearIndex];
+
+      partsDate.push(formatYear(year));
+    } else if (index === 'YY') {
+      const year = dateToArr[yearIndex];
+
+      partsDate.push(formatYear(year).slice(-2));
```
TarasHoliuk, line 74: you separated formatting logic into 2 parts, try to handle it in the one place

src/formatDate.js
```diff
@@ -50,7 +50,42 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const [, , , signFrom] = fromFormat;
```
TarasHoliuk, line 53:
```suggestion
  const signFrom = fromFormat[3];
```
not critical, but in this case, it looks better

src/formatDate.js
```diff
@@ -50,7 +50,42 @@
...
+  const dateToArr = date.split(signFrom);
+  const partsDate = [];
+
+  let yearIndex = -1;
+
+  if (fromFormat.includes('YYYY')) {
+    yearIndex = fromFormat.indexOf('YYYY');
+  } else if (fromFormat.includes('YY')) {
+    yearIndex = fromFormat.indexOf('YY');
+  }
```
TarasHoliuk, line 64: you may use ternary here

src/formatDate.js
```diff
@@ -50,7 +50,42 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const [, , , signFrom] = fromFormat;
+  const [, , , signTo] = toFormat;
+  const dateToArr = date.split(signFrom);
```
TarasHoliuk, line 55:
```suggestion
  const dateArray = date.split(signFrom);
```
it's not a function, it's already an array

TarasHoliuk (changes requested):

mykhalenych (approved):

PR #1433, opened 2023-07-30
the Format Date was complete

Viktoria-Moroz:


src/formatDate.js
```diff
@@ -50,7 +50,57 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let day = '';
```
SanyaBratashchuk, line 53: Instead of using three different variables (day, month, year), consider using an object to store these values. This would make your code cleaner and easier to manage.

SanyaBratashchuk (changes requested):

etojeDenys (approved):
lgtm

PR #1434, opened 2023-07-30
Solution

in7264:


src/formatDate.js
```diff
@@ -50,7 +51,44 @@
...
+          oldDate[fromFormat.indexOf('YY')] = 20 + '' + oldDate[fromFormat.indexOf('YY')];
+          break;
+
+        case oldDate[fromFormat.indexOf('YY')] === '30': // if we extend YY = 30 to YYYY
+        case oldDate[fromFormat.indexOf('YY')] > '30': // if we extend YY > 30 to YYYY
+          oldDate[fromFormat.indexOf('YY')] = 19 + '' + oldDate[fromFormat.indexOf('YY')];
+          break;
+      }
+    }
+  }
```
denys-danyliuk, line 80: I think you can do this with two cycles and two swtich statements, if you want you can improve your solution

denys-danyliuk (approved):
GJ!

PR #1437, opened 2023-07-31
FormatDate

Kacper-Leman:
Soultion

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -50,7 +50,47 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const getYearFormat = (year) => {
+    if (year.length === 2) {
+      const shortYear = parseInt(year, 10);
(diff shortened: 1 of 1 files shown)


src/formatDate.js
```diff
@@ -50,7 +50,47 @@
...
+
+    format.forEach((part, index) => {
+      dateObj[part] = parts[index];
+    });
+
+    return dateObj;
+  };
+
+  const formatDatePart = (dateObj, format) => {
+    return format.map((part) => dateObj[part]).join(format[format.length - 1]);
```
darokrk, line 75: you can remove curly braces and return word:

`const formatDatePart = (dateObj, format) => format.map((part) => dateObj[part]).join(format[format.length - 1]);`

darokrk (approved):
Good job 👏  left one mention, you can improve it but it's not necessary 😎

PR #1438, opened 2023-07-31
format date was made

EkaterinaBessonnaya:


src/index.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Ekaterina developer</title>
+  <script src="./formatDate.js"></script>
+</head>
+<body>
+
```
maxim2310, line 10: There is an unnecessary empty line before the closing body tag. It's a good practice to remove unnecessary empty lines to keep your code neat and readable.

src/index.html
```diff
@@ -0,0 +1,12 @@
...
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Ekaterina developer</title>
+  <script src="./formatDate.js"></script>
+</head>
+<body>
+  
+</body>
+</html>
\ No newline at end of file
```
maxim2310, line 12: There should be a newline at the end of file. It's a standard convention that every text file should end with a newline. This helps with version control systems and some tools to read and manipulate files.

src/formatDate.js
```diff
@@ -50,7 +50,45 @@
...
+      day = dateArr[i];
+    }
+  }
+
+  for (const format of toFormat) {
+    if (format.includes('Y')) {
+      if (format.length < year.length) {
+        newDate.push(year.slice(2));
+      } else {
+        newDate.push(year.padStart(format.length, year >= 30 ? '19' : '20'));
```
maxim2310, line 78: it's cool that you use new func, but this func is using where we need to duplicate something many times, in this case you just need to add 19 or 20 in start of string. And it may lead to difficulty understanding what is written.

maxim2310 (approved):
Nice work)

PR #1439, opened 2023-07-31
add form Date function

Meg-Sowka:


choeqq (approved):
Clever solution, nicely done 🚀

PR #1440, opened 2023-08-01
task has been done

vladkorobka:


.github/workflows/test.yml
loralevitska, line 1: For future don't add such files for review. Commit only solution

src/formatDate.js
```diff
@@ -50,7 +50,67 @@
...
+  const oldDateSeparator = fromFormat[3];
+  const newDateSeparator = toFormat[3];
+  const oldDate = date.split(oldDateSeparator);
+  let oldYear;
+  const month = oldDate[fromFormat.indexOf('MM')];
+  const day = oldDate[fromFormat.indexOf('DD')];
+  let newYear;
+  const lastElementInNewDate = toFormat[2];
+
+  // finding year in old format date and converting it to new format
```
loralevitska, line 64: Remove comment. Try it write obvious code that explain all we need without comments)

src/formatDate.js
```diff
@@ -50,7 +50,67 @@
...
+
+    if (toFormat.includes('YY')) {
+      newYear = oldYear.slice(2);
+    }
+  }
+
+  if (fromFormat.includes('YY')) {
+    oldYear = oldDate[fromFormat.indexOf('YY')];
+
+    newYear = oldYear;
```
loralevitska, line 78: Repeatable code move to helper function

loralevitska (changes requested):
Almost done) Let's improve

volodymyr-soltys97 (approved):

PR #1443, opened 2023-08-02
js_format_date_advanced

KarolinaWiktoria:


src/formatDate.js
```diff
@@ -50,7 +50,43 @@
...
+  );
+
+  const resultArray = [];
+
+  for (const key of toFormat) {
+    const keyChar = key.charAt(0);
+
+    switch (keyChar) {
+      case 'D':
+        resultArray.push(fromMp[keyChar]);
```
darokrk, line 67: seems case 'D' and 'M' doing the same operation, please join they into one to not duplicate code

darokrk (changes requested):
not bad, one mention need to be fixed 🛠️

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const splitDate = date.split(fromFormat[3]);
+  const fromMp = splitDate.reduce(
```
DorotaLeniecDev, line 54: Name variables more descriptively so your code is more readable
`dateMap`

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+        if (fromMp[keyChar].length === 4 && key.length === 2) {
+          fromMp[keyChar] = fromMp[keyChar].slice(2);
+        }
+
+        if (fromMp[keyChar].length === 2 && key.length === 4) {
+          if (Number(fromMp[keyChar]) >= 30) {
+            fromMp[keyChar] = '19' + fromMp[keyChar];
+          } else {
+            fromMp[keyChar] = '20' + fromMp[keyChar];
+          }
```
DorotaLeniecDev, line 80: I think this is a bit easier to read
```suggestion
          const year = Number(fromMp[keyChar]) >= 30 ? "19" : "20";
          fromMp[keyChar] = `${year}${fromMp[keyChar]}`;
```

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+        }
+        resultArray.push(fromMp[keyChar]);
+        break;
```
DorotaLeniecDev, line 83: Add default case to swich statement

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const splitDate = date.split(fromFormat[3]);
```
DorotaLeniecDev, line 53: Create variable as you are using it twice:
```js
  const dateSeparator = fromFormat[3]
```

DorotaLeniecDev (changes requested):
Good job, small fixes needed

darokrk (approved):
Well done  👏  ! Nice work 🥇

PR #1444, opened 2023-08-02
solution

DenisChakhalian:


Wesses (approved):
Good job.
Not bad idea to use object for storing data)

src/formatDate.js
```diff
@@ -50,7 +50,37 @@
...
+      continue;
+    }
+
+    if (datePart.length === 4) {
+      arrDateTo[toFormat.indexOf('YY')] = objDateFrom[datePart].slice(2);
+      continue;
+    }
+
+    if (datePart.length === 2) {
+      arrDateTo[toFormat.indexOf('YYYY')] = objDateFrom[datePart] < 30
```
BudnikOleksii, line 77: You use objDateFrom[datePart] so many times, let's create a variable for it.

BudnikOleksii (changes requested):
Looks good👍
Let's fix one thing

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const separatorFrom = fromFormat[3];
+  const separatorTo = toFormat[3];
+
+  const arrDateFrom = date.split(separatorFrom);
```
roman-mirzoian, line 56:
```suggestion
  const splittedDateFrom = date.split(separatorFrom);
```

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+  const objDateFrom = {};
+
+  const arrDateTo = [];
```
roman-mirzoian, line 59:
```suggestion
  const formattedDateTo = [];
```

src/formatDate.js
```diff
@@ -50,7 +50,39 @@
...
+  for (const datePart in objDateFrom) {
+    const dateValue = objDateFrom[datePart];
+
+    if (toFormat.includes(datePart)) {
+      arrDateTo[toFormat.indexOf(datePart)] = dateValue;
+      continue;
+    }
+
+    if (datePart.length === 4) {
+      arrDateTo[toFormat.indexOf('YY')] = dateValue.slice(2);
```
roman-mirzoian, line 74: It would be cool to put the "magic values" in separate constants.

roman-mirzoian (changes requested):

BudnikOleksii (approved):
Nicely done!

PR #1448, opened 2023-08-02
add task solution

Wesses:


src/formatDate.js
```diff
@@ -50,7 +50,41 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateArray = date.split(fromFormat[3]);
```
AndriiKuz, line 53: Will be better if you put fromFormat[3] in the variable. 
For example: const separator = fromFormat[3].

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+
+  if (fromFormat[fromYearIndex] === toFormat[toYearIndex]) {
+    return;
+  }
+
+  fromFormat[fromYearIndex] = toFormat[toYearIndex];
+
+  if (fromFormat[fromYearIndex] === 'YYYY') {
+    if (+dateArray[fromYearIndex] < 30) {
+      dateArray[fromYearIndex] = '20' + dateArray[fromYearIndex];
```
AndriiKuz, line 78: Don`t use magic values. Create variables for it.
For example: const CENTURY_21 = 20;

AndriiKuz (changes requested):

AndriiKuz:
Nice work, but some changes are needed.

src/formatDate.js
```diff
@@ -48,9 +48,51 @@
...
  */
+const CENTURY_INDICATOR = 30;
+const CENTURY_21 = '20';
+const CENTURY_20 = '19';
+const YEAR_FORMAT_OF_4 = 'YYYY';
+const YEAR_FORMAT_OF_2 = 'YY';
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparator = fromFormat[3];
```
roman-mirzoian, line 58:
```suggestion
  const currentSeparator = fromFormat[3];
```

src/formatDate.js
```diff
@@ -48,9 +48,51 @@
...
+  const toSeparator = toFormat[3];
```
roman-mirzoian, line 59:
```suggestion
  const newSeparator = toFormat[3];
```

src/formatDate.js
```diff
@@ -48,9 +48,51 @@
...
+  const dateArray = date.split(fromSeparator);
```
roman-mirzoian, line 60:
```suggestion
  const splitedDate = date.split(fromSeparator);
```

src/formatDate.js
```diff
@@ -48,9 +48,51 @@
...
+  const newDateArray = [];
```
roman-mirzoian, line 61:
```suggestion
  const formatedDate = [];
```

src/formatDate.js
```diff
@@ -48,9 +48,51 @@
...
+
+  formatYear(fromFormat, toFormat, dateArray);
+
+  for (let i = 0; i < dateArray.length; i++) {
+    newDateArray[i] = dateArray[fromFormat.indexOf(toFormat[i])];
```
roman-mirzoian, line 66: Now everything is written correctly, but it's a bit hard to read, you could move the search for the new index into a function:
```js
function findFormatIndex(oldFormat, newFormat, i) {
  return oldFormat.indexOf(newFormat[i]);
}
```

src/formatDate.js
```diff
@@ -48,9 +48,51 @@
...
+  if (fromFormat[fromYearIndex] === toFormat[toYearIndex]) {
+    return;
+  }
+
+  fromFormat[fromYearIndex] = toFormat[toYearIndex];
+
+  if (fromFormat[fromYearIndex] === YEAR_FORMAT_OF_4) {
+    if (+dateArray[fromYearIndex] < CENTURY_INDICATOR) {
+      dateArray[fromYearIndex] = CENTURY_21 + dateArray[fromYearIndex];
+
```
roman-mirzoian, line 87:
```suggestion
```

src/formatDate.js
```diff
@@ -48,9 +48,51 @@
...
+      return;
+    }
+    dateArray[fromYearIndex] = CENTURY_20 + dateArray[fromYearIndex];
```
roman-mirzoian, line 90:
```suggestion

    dateArray[fromYearIndex] = CENTURY_20 + dateArray[fromYearIndex];
```

src/formatDate.js
```diff
@@ -48,9 +48,51 @@
...
+
```
roman-mirzoian, line 91:
```suggestion
```

src/formatDate.js
```diff
@@ -48,9 +48,51 @@
...
+  if (fromFormat[fromYearIndex] === toFormat[toYearIndex]) {
+    return;
+  }
+
+  fromFormat[fromYearIndex] = toFormat[toYearIndex];
+
+  if (fromFormat[fromYearIndex] === YEAR_FORMAT_OF_4) {
+    if (+dateArray[fromYearIndex] < CENTURY_INDICATOR) {
+      dateArray[fromYearIndex] = CENTURY_21 + dateArray[fromYearIndex];
```
roman-mirzoian, line 86: `dateArray[fromYearIndex]` is repeated many times, this can be put into a variable and reused.

roman-mirzoian (changes requested):
Remember good naming and use of variables to improve code.

src/formatDate.js
Wesses: through this expression, I change the value inside of the array, if I create a variable, both the variable and the direct access to the array element through this entry will appear in my code. I think it will confuse more then now.

src/formatDate.js
Wesses: linter

PR #1451, opened 2023-08-03
Add task solution

davinnchii:


src/formatDate.js
```diff
@@ -51,6 +51,45 @@
...
+    if (index <= 2) {
+      oldIndexes[value] = index;
+    }
+  });
+
+  toFormat.forEach((value, index) => {
+    if (index <= 2) {
+      newIndexes[value] = index;
+    }
+  });
```
FrontKid, line 70: You can use separate function for this cases

src/formatDate.js
```diff
@@ -51,6 +51,45 @@
...
+
+  // In case of old format was with YY year, we changing our year
+  // and property name for our convenience
+  if (oldIndexes.hasOwnProperty('YY') && newIndexes.hasOwnProperty('YYYY')) {
+    splittedDate[oldIndexes.YY] = splittedDate[oldIndexes.YY] >= 24
```
FrontKid, line 75: if we have XX29 it should be 2029 not 1929
change 24 to 30

FrontKid (changes requested):
GOOD JOB MY DEAR FRIED!!!

davinnchii:
> GOOD JOB MY DEER FRIED!!!
🦌

davinnchii:
```
const CURRENT_DATE = new Date();
const CURRENT_YEAR = Number(CURRENT_DATE.getFullYear().toFixed().slice(-2));
```
I used this code to convert the year, because as for me it is more obvious that if the year in date is bigger than current year then it is previous century. But in task I saw 30, which seems for me as a random number. So I decided to improve my code a bit

src/formatDate.js
```diff
@@ -49,8 +49,51 @@
  * @returns {string}
  */
 
+function getFormatIndexes(array, object) {
+  array.forEach((value, index) => {
+    if (index <= 2) {
```
33Orange, line 54: The `index <= 2` condition is not clear from the code. It would be better to add a comment explaining why we are only interested in the first three elements of the array or use a meaningful constant instead of `2` to enhance code readability.

33Orange (approved):
Well done

PR #1453, opened 2023-08-04
Format date

vserhiichuk:


src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+
+    if (elementArray === 'YYYY') {
+      FORMAT_YEAR = dateFormatOld[i];
+    } else if (elementArray === 'YY') {
+      FORMAT_YEAR = dateFormatOld[i];
+    } else if (elementArray === 'MM') {
+      FORMAT_MONTH = dateFormatOld[i];
+    } else if (elementArray === 'DD') {
+      FORMAT_DAY = dateFormatOld[i];
+    }
```
desevoker, line 74:
```suggestion
    const part = fromFormat[i];
    
    switch (part) {
      case 'YY':
      case 'YYYY':
        year = dateValuesOld[i];
        break;
      case 'MM':
        month = dateValuesOld[i];
        break;
      case 'DD':
        day =  dateValuesOld[i];
        break;
    }
```
Consider using `switch` as a better option.

src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+      ;
+    } else if (elementArray === 'YY') {
+      const changeFormatYear = FORMAT_YEAR.split('').slice(2, 4).join('');
+
+      dateFormatNew.push(changeFormatYear);
+    } else if (elementArray === 'MM') {
+      dateFormatNew.push(FORMAT_MONTH);
+    } else if (elementArray === 'DD') {
+      dateFormatNew.push(FORMAT_DAY);
+    }
```
desevoker, line 102: Same as in the section above: consider using `switch`, choose better names for the variables, try to reduce nesting.

src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+        changeFormatYear = FORMAT_YEAR < 30
+          ? '20' + FORMAT_YEAR
+          : '19' + FORMAT_YEAR;
+
+        dateFormatNew.push(changeFormatYear);
+      } else {
+        dateFormatNew.push(changeFormatYear);
+      }
+
+      ;
```
desevoker, line 93:
```suggestion
```

src/formatDate.js
```diff
@@ -50,7 +50,59 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let FORMAT_YEAR = '';
+  let FORMAT_MONTH = '';
+  let FORMAT_DAY = '';
+  const dateFormatNew = [];
```
desevoker, line 56:
```suggestion
  const dateValuesNew = [];
```

src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+        dateFormatNew.push(changeFormatYear);
+      } else {
+        dateFormatNew.push(changeFormatYear);
+      }
+
+      ;
+    } else if (elementArray === 'YY') {
+      const changeFormatYear = FORMAT_YEAR.split('').slice(2, 4).join('');
+
+      dateFormatNew.push(changeFormatYear);
```
desevoker, line 97:
```suggestion
      dateValuesNew.push(year.substring(2));
```
That can be done in a much simpler way and without a variable.

desevoker (commented):
Good job! 🚀

But overall naming and readability is not great.
Check the comments, hope they can help.

Orpheus29 (approved):
Everything's gonna be OK as soon as you consider Dmytro's remarks.

src/formatDate.js
alexander-ignatow: Comment makes sense - please fix it

src/formatDate.js
alexander-ignatow: Please fix this comment

src/formatDate.js
```diff
@@ -50,7 +50,66 @@
...
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    const part = toFormat[i];
+    let changeYear = year;
+
+    switch (part) {
+      case 'YYYY':
+        if (year.length === 2) {
+          changeYear = year < 30
+            ? CENTURE_XXI + year
+            : CENTURE_XX + year;
```
alexander-ignatow, line 93: Don't use `+` for the string concatenation - use template literals instead

src/formatDate.js
```diff
@@ -50,7 +50,66 @@
...
+        }
+        dateFormatNew.push(changeYear);
+        break;
+      case 'YY':
+        changeYear = year.split('').slice(2, 4).join('');
```
alexander-ignatow, line 98: Consider using `String.subsctring`

alexander-ignatow (changes requested):
Overall - not bad 👍 

You have fixed many issues - still some are left

src/formatDate.js
```diff
@@ -50,7 +50,65 @@
...
+  const separatorOld = fromFormat[fromFormat.length - 1];
+  const separatorNew = toFormat[toFormat.length - 1];
+
+  const dateValuesOld = date.split(separatorOld);
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    const part = fromFormat[i];
+
+    switch (part) {
+      case 'YY':
```
roman-mirzoian, line 69: YY, YYYY, MM etc. is also "magic values".

roman-mirzoian (approved):
Well done, but pay attention for using "magic values".

PR #1455, opened 2023-08-04
initial commit

maksym-mishchanchuk:


src/formatDate.js
```diff
@@ -50,7 +53,65 @@
...
+      = newFormat.filter(partDate => partDate.includes('YY'))[0];
+
+    const previousCentury = '19';
+    const currentCentury = '20';
+
+    if (yearBefore.toString().length > yearUpdateFormat.length) {
+      return yearBefore.slice(-2);
+    }
+
+    if (yearBefore.length === yearUpdateFormat.length) {
```
davinnchii, line 67: Do you really need this ``if``? You return the same value on 68th line.

src/formatDate.js
```diff
@@ -50,7 +53,65 @@
...
+      : fromFormat.indexOf('YYYY');
+
+    const indexMM = fromFormat.indexOf('MM');
+    const indexDD = fromFormat.indexOf('DD');
+
+    const day = dateArr[indexDD];
+    const month = dateArr[indexMM];
+    const year = dateArr[indexYY];
+
+    const separator = toFormat[toFormat.length - 1];
```
davinnchii, line 97: > const dateArr = date.split(fromFormat[3]);

On 85th line you used last index, but here you are using length - 1, as for me it will be better to do it in one way(length - 1 in both lines or last index in both lines)

src/formatDate.js
```diff
@@ -50,7 +53,65 @@
...
+    }
+
+    if (toFormat[i] === 'YY' || toFormat[i] === 'YYYY') {
+      const yearUpdate = prefixYear(year, toFormat);
+
+      newFormatDate += yearUpdate + separator;
+    }
+  }
+
+  return newFormatDate.slice(0, newFormatDate.length - 1);
```
davinnchii, line 114: I think it will be better to set newFormatDate to array and just add values as string by index and then return this array.join(separator). You will get rid of repeating separator and your return statement will be easier to understand.

davinnchii (changes requested):
GJ ma friend, but still need to improve it a bit!

src/formatDate.js
maksym-mishchanchuk: Так, тут я перевіряю якщо формати дати співпадають -- відразу повертаю рік без змін

src/formatDate.js
maksym-mishchanchuk: agree

src/formatDate.js
maksym-mishchanchuk: it seems obvious to me that slice newFormatDate. length - 1 I just don't include the last separator.

So your suggestion will make the code shorter, but I can't see the difference for reading))

davinnchii (approved):

src/formatDate.js
```diff
@@ -50,7 +53,65 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const prefixYear = (yearBefore, newFormat) => {
+    const yearUpdateFormat
+      = newFormat.filter(partDate => partDate.includes('YY'))[0];
```
alexander-ignatow, line 58: Please checkout `Array.find`

src/formatDate.js
alexander-ignatow: Agree with this condition
Probably, it makes sense to perform this check once you calculate `yearUpdateFormat` - from perspective of very slight optimization

src/formatDate.js
```diff
@@ -50,7 +53,65 @@
...
+      return yearBefore;
+    }
+
+    if (yearUpdateFormat.length === 4 && yearBefore >= 30) {
+      return previousCentury + yearBefore;
+    }
+
+    if (yearUpdateFormat.length === 4 && yearBefore < 30) {
+      return currentCentury + yearBefore;
+    }
```
alexander-ignatow, line 77: Please, don't use magic numbers

src/formatDate.js
```diff
@@ -50,7 +53,65 @@
...
+
+    return yearBefore;
+  };
+
+  let newFormatDate = '';
+
+  for (let i = 0; i < toFormat.length; i++) {
```
alexander-ignatow, line 84: Please re-check this cycle's body
There are some values that don't need to be calculated each iteration - you can di it once before the loop.

src/formatDate.js
```diff
@@ -50,7 +53,65 @@
...
+    if (toFormat[i] === 'MM') {
+      newFormatDate += month + separator;
+    }
+
+    if (toFormat[i] === 'YY' || toFormat[i] === 'YYYY') {
+      const yearUpdate = prefixYear(year, toFormat);
+
+      newFormatDate += yearUpdate + separator;
+    }
+  }
```
alexander-ignatow, line 112: In JS strings are immutable, so each time you `+=` them you create new instances - which is an antipattern.

There are several ways to fix this - but try to "mutate" strings

As an alternative - you can use an array and then make `Array.join` or you can go with template literals

alexander-ignatow (changes requested):
Good start, but some improvements needed

src/formatDate.js
maksym-mishchanchuk: thanks, all comment very relevant

src/formatDate.js
```diff
@@ -50,7 +53,68 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const prefixYear = (yearBefore, newFormat) => {
+    const yearUpdateFormat
+      = newFormat.find(partDate => partDate.includes('YY'));
+
+    const previousCentury = '19';
+    const currentCentury = '20';
+    const toDetermineCentury = 30;
```
roman-mirzoian, line 62: Immutable constants should be named in uppercase.

src/formatDate.js
```diff
@@ -50,7 +53,68 @@
...
+
+    if (yearBefore.length === yearUpdateFormat.length) {
+      return yearBefore;
+    }
```
roman-mirzoian, line 66: Nice condition, but then it's better to call it before creating the constants so that the constants aren't created when they're not needed.

src/formatDate.js
```diff
@@ -50,7 +53,68 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const prefixYear = (yearBefore, newFormat) => {
```
roman-mirzoian, line 56: The function name must begin with a verb.

src/formatDate.js
```diff
@@ -50,7 +53,68 @@
...
+    const yearUpdateFormat
+      = newFormat.find(partDate => partDate.includes('YY'));
```
roman-mirzoian, line 58: It would be cool to put all the "magic values" in a constant.

src/formatDate.js
```diff
@@ -50,7 +53,68 @@
...
+      return yearBefore.slice(-2);
+    }
+
+    if (yearUpdateFormat.length === 4 && yearBefore >= toDetermineCentury) {
+      return previousCentury + yearBefore;
+    }
+
+    if (yearUpdateFormat.length === 4 && yearBefore < toDetermineCentury) {
+      return currentCentury + yearBefore;
+    }
```
roman-mirzoian, line 78: You have a partially repeated condition and the same meaning of the return value, so it could have been written a little more simply:
```suggestion
    if (yearUpdateFormat.length === 4) {
      const century = yearBefore < toDetermineCentury ? currentCentury : previousCentury;
      return century + yearBefore;
    }
```

src/formatDate.js
```diff
@@ -50,7 +53,68 @@
...
+
+    return yearBefore;
```
roman-mirzoian, line 80: You work out all the possible conditions for your values and this `return` will never work, are you sure you need it here?

src/formatDate.js
```diff
@@ -50,7 +53,68 @@
...
+  };
+
+  const newFormatDate = [];
+  const dateArr = date.split(fromFormat[fromFormat.length - 1]);
```
roman-mirzoian, line 84:
```suggestion
  const splittedOldDate = date.split(fromFormat[fromFormat.length - 1]);
```

src/formatDate.js
```diff
@@ -50,7 +53,68 @@
...
+  const indexYY = fromFormat
+    .findIndex((partDate) => partDate === 'YY' || partDate === 'YYYY');
+
+  const indexMM = fromFormat.indexOf('MM');
+  const indexDD = fromFormat.indexOf('DD');
```
roman-mirzoian, line 89: It is worth choosing more meaningful variable names.

src/formatDate.js
```diff
@@ -50,7 +53,68 @@
...
+      newFormatDate.push(separator);
+    }
+
+    if (toFormat[i] === 'YY' || toFormat[i] === 'YYYY') {
+      const yearUpdate = prefixYear(year, toFormat);
+
+      newFormatDate.push(yearUpdate);
+      newFormatDate.push(separator);
+    }
+  }
```
roman-mirzoian, line 115: Why do you add a `separator` each time separately? Wouldn't it be easier to collect all the values in an array and combine them using a `join` specifying a new `separator` as an argument?

roman-mirzoian (changes requested):

src/formatDate.js
```diff
@@ -50,7 +53,66 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const yearFormatYY = 'YY';
+  const yearFormatYYYY = 'YYYY';
+  const dayFormat = 'DD';
+  const monthFormat = 'MM';
```
alexander-ignatow, line 59: Those constants should be UPPER_CASE

alexander-ignatow (changes requested):
Almost done 👍 

Just make a final change and rename the constants, please

Common rule: use UPPER_CASE for constants that are inited with some values and pascalCase for variables and something you calculate.

src/formatDate.js
```diff
@@ -50,7 +53,66 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const YEAR_FORMAT_YY = 'YY';
+  const YEAR_FORMAT_YYYY = 'YYYY';
+  const DAY_FORMAT = 'DD';
+  const MONTH_FORMAT = 'MM';
+  const getPrefixYear = (yearBefore, newFormat) => {
+    const yearUpdateFormat
+      = newFormat.find(partDate => partDate.includes(YEAR_FORMAT_YY));
+
```
BudnikOleksii, line 63: Don't create functions inside other functions.
You can move it outside, also, constants too.

src/formatDate.js
```diff
@@ -50,7 +53,66 @@
...
+    .findIndex((partDate) =>
+      partDate === YEAR_FORMAT_YY || partDate === YEAR_FORMAT_YYYY);
+
+  const indexForMounth = fromFormat.indexOf(MONTH_FORMAT);
+  const indexForDay = fromFormat.indexOf(DAY_FORMAT);
+
+  const day = SPLITTED_OLD_DATE[indexForDay];
+  const month = SPLITTED_OLD_DATE[indexForMounth];
+  const year = SPLITTED_OLD_DATE[indexYY];
+  const SEPARATOR_INDEX = toFormat[toFormat.length - 1];
```
BudnikOleksii, line 97: SEPARATOR_INDEX and SPLITTED_OLD_DATE are not general constants as they depend on what you pass to function. So use camelCase for them instead.

src/formatDate.js
```diff
@@ -50,7 +53,66 @@
...
+  for (let i = 0; i < toFormat.length; i++) {
+    if (toFormat[i] === DAY_FORMAT) {
+      newFormatDate.push(day);
+    }
+
+    if (toFormat[i] === MONTH_FORMAT) {
+      newFormatDate.push(month);
+    }
+
+    if (toFormat[i] === YEAR_FORMAT_YY || toFormat[i] === YEAR_FORMAT_YYYY) {
```
BudnikOleksii, line 108: Prefer to use "for of" instead of "for" loop. Also, looks like the switch case ші a better fit in this case

BudnikOleksii (changes requested):
Good work!
But it could be better and cleaner, check the comments.

src/formatDate.js
```diff
@@ -42,15 +42,77 @@
...
+  }
+
+  if (yearUpdateFormat.length === 4) {
+    const century = yearBefore < TO_DETERMINE_CENTURY
+      ? CURRENT_CENTURY
+      : PREVIOUS_CENTURY;
+
+    return century + yearBefore;
+  }
+};
```
alexander-ignatow, line 82: One minor problem with this function - it looks that there is possible situation when it returns `undefined`

Please make changes to make sure it always returns some meaningful value

alexander-ignatow (changes requested):
Almost done!

Please make one final change

BudnikOleksii (approved):
Nicely done🔥

PR #1456, opened 2023-08-04
Solution

desevoker:


src/formatDate.js
```diff
@@ -1,56 +1,68 @@
...
+
+/**
+ * @param {Object} data
+ * @param {string[]} format
+ *
+ * @returns {string}
+ */
+function arrangeDate(data, format) {
+  const parts = format.slice(0, -1);
+  const separator = format.slice(-1);
```
kaavka, line 58: You are using -1 so often if you made it in one function it could be separate const.

kaavka (commented):
Wow. Thats a pretty complicated solution. I dont see a reason why would we separate it on three functions. Like arrangeDate is literrary doing the same job what we expect from formatDate to do. I could be wrong of course) so... might be better to leave it to the judgment of the mentors

src/formatDate.js
desevoker: Two times in one function is not so often. But it sure can be a constant, even if it is being used only once.

kaavka: I counted 4 times.
You also have same logic on 27, 28, lines.

desevoker:
> I dont see a reason why would we separate it on three functions. Like arrangeDate is literrary doing the same job what we expect from formatDate to do. I could be wrong of course.

I think that expectations have nothing to do with implementation details. I can make as many helper functions as I want and then just compose them in the main function.

It looks like it is doing the same job, but it also requires a slightly different type of input values.

src/formatDate.js
desevoker: I know, that's why I wrote "two times in one function", and there's two functions, obviously :)

kaavka: Misunderstanding)

kaavka:
> I think that expectations have nothing to do with implementation details. I can make as many helper functions as I want and then just compose them in the main function.
Shure you could) Like, you could also make separate function to get out parts and divider. Just thinking about how necessary it is. Your solution, while works perfectly, seems a bit more complicat than it could be, at least to my perspective. But it's probalby just diference between our approaches.

kaavka (approved):

roman-mirzoian (approved):
Cool solution!
The only thing I can add: if you move the creation of `SEPARATOR_OFFSET` outside the definition of all functions (that is, before creating the `formatDate` function), then this constant will not need to be created several times.

PR #1457, opened 2023-08-05
Solution

neovendano:


src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const CENTURY_EDGE = 30;
+  const XXI_CENTURY = 20;
+  const XX_CENTURY = 19;
+  const SHORT_YEAR_FORM = 'YY';
+  const LONG_YEAR_FORM = 'YYYY';
+
+  const dateParts = date.split(fromFormat[3]);
+  const dateObj = {};
```
kaavka, line 60: Almoust shure that you could give it better name)

kaavka (changes requested):
Looks great! Few changes and going to be close to perfect.
And know that rejecting because of names could be tiring but...)

AngerDESTROYS (approved):
seems to be okay

src/formatDate.js
roman-mirzoian: I agree, better naming should be chosen here.

roman-mirzoian (approved):
Well done, but pay more attention to better naming.

PR #1463, opened 2023-08-06
add task solution

voievodik:


KatOlista (changes requested):
good job. But I think dubble switch-case is redundant

voievodik:
> good job. But I think dubble switch-case is redundant

why?

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const CENTURY_21 = '20';
+  const CENTURY_20 = '19';
+  const dateArr = date.split(`${fromFormat[3]}`);
```
BudnikOleksii, line 55:
```suggestion
  const dateArr = date.split(fromFormat[3]);
```
You don't need interpolation in such cases.

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
+
+  if (shortYear === undefined) {
+    shortYear = longYear.slice(2, 4);
+  } else if (longYear === undefined) {
+    if (shortYear < 30) {
+      longYear = CENTURY_21 + shortYear;
+    } else {
+      longYear = CENTURY_20 + shortYear;
+    }
+  }
```
BudnikOleksii, line 84: If you need to check that value doesn't exist you can use just Logical NOT (!).
!shortYear instead of shortYear === undefined.
Also, it's a bad practice to use "else if" and nested "if" conditions.
Try to get rid of them.

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
+
+  for (let i = 0; i < toFormat.length; i++) {
+    switch (toFormat[i]) {
```
BudnikOleksii, line 87: Prefer "for of" on regular "for" loop

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
+      case 'MM':
```
BudnikOleksii, line 88: "MM" and other you use the second time, looks like you can move them to variables.

BudnikOleksii (changes requested):
Good job!
Left some comments, check them.

src/formatDate.js
```diff
@@ -48,9 +48,61 @@
...
+        break;
+    }
+  }
+
+  if (!shortYear) {
+    shortYear = longYear.slice(2, 4);
+  }
+
+  if (!longYear) {
+    longYear = shortYear < 30 ? CENTURY_21 + shortYear : CENTURY_20 + shortYear;
```
alexander-ignatow, line 85: Avoid using `+` for string concatenation - consider template literals instead

src/formatDate.js
```diff
@@ -48,9 +48,61 @@
...
+        break;
+      case LONG_YEAR_ACRONYM:
+        longYear = dateArr[i];
+        break;
+    }
+  }
+
+  if (!shortYear) {
+    shortYear = longYear.slice(2, 4);
+  }
```
alexander-ignatow, line 82: No need to change - but an even better idea will be combining conditions with switch block
```suggestion
  for (let i = 0; i < fromFormat.length; i++) {
    switch (fromFormat[i]) {
      case MONTH_ACRONYM:
        month = dateArr[i];
        break;
      case DAY_ACRONYM:
        day = dateArr[i];
        break;
      case SHORT_YEAR_ACRONYM:
        shortYear = dateArr[i];
        break;
      case LONG_YEAR_ACRONYM:
        longYear = dateArr[I];
        shortYear = longYear.slice(2, 4);
        break;
    }
  }

```
And same for the next condition with long year

alexander-ignatow (changes requested):
very good 👍 

Please fix minor remaining things

BudnikOleksii (approved):
Nicely done👍

PR #1464, opened 2023-08-06
add task solution

Vayts:


viktoria-roik (approved):
Good Job!!!

src/formatDate.js
```diff
@@ -49,8 +49,45 @@
...
+      const dayIndex = fromFormat.indexOf(DAY_FORMAT);
+
+      formattedDate.push(dateParts[dayIndex]);
+    }
+
+    if (item === MONTH_FORMAT) {
+      const monthIndex = fromFormat.indexOf(MONTH_FORMAT);
+
+      formattedDate.push(dateParts[monthIndex]);
+    }
```
BudnikOleksii, line 87: In all if statements you check the item, so in this case you use switch case instead, it would be more readable.

BudnikOleksii (changes requested):
Nicely done!
Just use the switch case to check the item value.

PR #1466, opened 2023-08-07
Solution

SergiyVS:


src/formatDate.js
```diff
@@ -45,12 +45,67 @@
...
+      if (elem.length === oldYear.length) {
+        newYear = oldYear;
+      }
+
+      if (elem.length < oldYear.length) {
+        newYear = oldYear.slice(2, 4);
+      }
+
+      if (elem.length > oldYear.length) {
+        if (oldYear < '30') {
```
Anastasiia-Svintsova, line 97: The comparison of 'oldYear' with '30' is incorrect. It should be compared as a number, not as a string. Convert 'oldYear' to a number before the comparison.

src/formatDate.js
```diff
@@ -45,12 +45,67 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let newFormatDate = '';
+  let newDay = '';
+  let newMonth = '';
+  let newYear = '';
+  const newSeparator = toFormat[3];
+
+  let oldDateSplit = [];
```
Anastasiia-Svintsova, line 59: The initialization of 'oldDateSplit' and 'oldYear' is unnecessary since they are immediately given a value in the next lines.

src/formatDate.js
```diff
@@ -45,12 +45,67 @@
...
+        } else {
+          newYear = '19' + oldYear;
+        }
+      }
+
+      newFormatDate += (newYear + newSeparator);
+    }
+  }
+
+  return newFormatDate.slice(0, (newFormatDate.length - 1));
```
Anastasiia-Svintsova, line 108: The 'slice' method is used to remove the last separator. Consider using the 'join' method to concatenate date parts, it will automatically add separators between elements and avoid an extra removal operation.

Anastasiia-Svintsova (changes requested):
Almost done!

PR #1467, opened 2023-08-07
task solution

viktoria-roik:


ihorid (approved):
Good job!

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newDate = date.split(fromFormat[3]);
+  const datePart = {};
+  const formattedDate = [];
+
+  const currentCentury = '20';
+  const previousCentury = '19';
+
+  const CENTURY_INDICATOR = 30;
```
BudnikOleksii, line 60: You can move current constants outside the function so we can reuse them if needed.
Also, why do they have different naming styles? If the currentCentury and the previousCentury are local variables it makes sense, otherwise lets do them in the same case as CENTURY_INDICATOR.

src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
+
+  if (datePart.YY >= CENTURY_INDICATOR) {
+    datePart.YYYY = previousCentury + datePart.YY;
+  } else {
+    datePart.YYYY = currentCentury + datePart.YY;
+  }
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    formattedDate.push(datePart[toFormat[i]]);
+  }
```
BudnikOleksii, line 78: It's better to use "for of" instead of "for" loop.

BudnikOleksii (changes requested):
Great work!
Just a few suggestions to improve your solution.

src/formatDate.js
```diff
@@ -49,8 +49,35 @@
...
  */
 
+const CURRENT_CENTURY = '20';
+const PREVIOUS_CENTURY = '19';
+const CENTURY_INDICATOR = 30;
+
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newDate = date.split(fromFormat[3]);
+  const datePart = {};
```
roman-mirzoian, line 58:
```suggestion
  const dateParts = {};
```

src/formatDate.js
```diff
@@ -49,8 +49,35 @@
  * @returns {string}
  */
 
+const CURRENT_CENTURY = '20';
+const PREVIOUS_CENTURY = '19';
+const CENTURY_INDICATOR = 30;
+
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newDate = date.split(fromFormat[3]);
```
roman-mirzoian, line 57:
```suggestion
  const oldSplittedDate = date.split(fromFormat[3]);
```

roman-mirzoian (approved):
Well done 👍

PR #1469, opened 2023-08-07
add solution

KatOlista:


src/formatDate.js
```diff
@@ -49,8 +49,68 @@
...
+  const dateInNewFormat = [];
+  const formatDates = date.split(fromFormat[3]);
+
+  const yearIndexFromFormat = fromFormat.indexOf('YY') >= 0
+    ? fromFormat.indexOf('YY')
+    : fromFormat.indexOf('YYYY');
+
+  const yearIndexToFormat = toFormat.indexOf('YY') >= 0
+    ? toFormat.indexOf('YY')
+    : toFormat.indexOf('YYYY');
```
roman-mirzoian, line 78: It would be cool to put the "magic values" in separate constants.

src/formatDate.js
```diff
@@ -49,8 +49,68 @@
...
+}
+
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateInNewFormat = [];
+  const formatDates = date.split(fromFormat[3]);
+
+  const yearIndexFromFormat = fromFormat.indexOf('YY') >= 0
+    ? fromFormat.indexOf('YY')
+    : fromFormat.indexOf('YYYY');
```
roman-mirzoian, line 74: You can make the search easier by using `Math`:
```suggestion
   const yearIndexFromFormat = Math.max(fromFormat.indexOf('YY'), fromFormat.indexOf('YYYY'));
```
The same for `yearIndexToFormat`.

src/formatDate.js
```diff
@@ -49,8 +49,68 @@
...
+        dateInNewFormat[i] = getNewFormatYear(yearFromFormat, yearToFormat,
+          formatDates[yearIndexFromFormat]);
+        break;
+      }
+
+      case 'YYYY': {
+        dateInNewFormat[i] = getNewFormatYear(yearFromFormat, yearToFormat,
+          formatDates[yearIndexFromFormat]);
+        break;
+      }
```
roman-mirzoian, line 105: Why are you repeating the same code twice?
This part can be moved to the `default` block so as not to repeat the condition twice and set the length of the `for` loop `toFormat.length - 1` to ignore the last element.

src/formatDate.js
```diff
@@ -49,8 +49,68 @@
...
+      ? 20 + '' + oldFormatYear
+      : 19 + '' + oldFormatYear;
+  }
+
+  return newFormatYear;
+}
+
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateInNewFormat = [];
```
roman-mirzoian, line 69:
```suggestion
  const formattedDate = [];
```

src/formatDate.js
```diff
@@ -49,8 +49,68 @@
...
+  const formatDates = date.split(fromFormat[3]);
```
roman-mirzoian, line 70:
```suggestion
  const splittedOldDate= date.split(fromFormat[3]);
```

src/formatDate.js
```diff
@@ -49,8 +49,68 @@
  * @returns {string}
  */
 
+function getNewFormatYear(fromFormYear, toFormYear, oldFormatYear) {
+  let newFormatYear = oldFormatYear;
```
roman-mirzoian, line 53: It is a little unclear why you assign a variable of the new format to a variable of the old format.
It will be enough to `return` the changed value for each `if`, then there will be no need to create an extra variable.

src/formatDate.js
```diff
@@ -49,8 +49,68 @@
...
+
+  if (fromFormYear.length > toFormYear.length) {
+    newFormatYear = oldFormatYear.slice(2);
+  }
+
+  if (fromFormYear.length < toFormYear.length) {
+    newFormatYear = oldFormatYear < 30
+      ? 20 + '' + oldFormatYear
+      : 19 + '' + oldFormatYear;
```
roman-mirzoian, line 62:
```suggestion
      ? '20' + oldFormatYear
      : '19' + oldFormatYear;
```

roman-mirzoian (changes requested):

BudnikOleksii (approved):
Good work🚀

PR #1470, opened 2023-08-07
solution format data

ihorid:


src/formatDate.js
```diff
@@ -50,7 +50,48 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  let dateToArr = [];
+  const dateObject = {};
+  const newFormatOfDate = [];
+  let newFormatOfDateToReturn = '';
+
+  for (let i = fromFormat.length - 1; i >= 0; i--) {
+    if (i === fromFormat.length - 1) {
```
rostyslav48, line 59: Isn't it better to move the code inside this if out of the for loop. The variables are assigned only once when the loop starts. So, putting if there makes your code harder to read

src/formatDate.js
```diff
@@ -50,7 +50,42 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateToArr = date.split(fromFormat[3]);
+  const dateObject = {};
```
Moroz-Dmytro, line 54: Try to make up better names

src/formatDate.js
```diff
@@ -50,7 +50,42 @@
...
+  const newFormatOfDate = [];
+  let newFormatOfDateToReturn = '';
+  const LONG_FORMAT = 'YYYY';
+  const SHORT_FORMAT = 'YY';
```
Moroz-Dmytro, line 58: Move such constants outside the function

src/formatDate.js
```diff
@@ -50,7 +50,42 @@
...
+
+    if (curentValue in dateObject) {
+      newFormatOfDate.push(dateObject[curentValue]);
+    } else {
+      if (toFormat[i] === SHORT_FORMAT) {
+        year = dateObject[LONG_FORMAT].slice(2);
+      } else {
+        year = (dateObject[SHORT_FORMAT] < 30)
+          ? '20' + dateObject[SHORT_FORMAT]
+          : '19' + dateObject[SHORT_FORMAT];
```
Moroz-Dmytro, line 80: Magic values, create constants

src/formatDate.js
```diff
@@ -50,7 +50,42 @@
...
+      }
+      newFormatOfDate.push(year);
+    }
+  }
+
+  newFormatOfDateToReturn = newFormatOfDate.join(toFormat[3]);
+
+  return newFormatOfDateToReturn;
```
Moroz-Dmytro, line 88:
```suggestion
  return newFormatOfDate.join(toFormat[3]);
```

src/formatDate.js
```diff
@@ -50,7 +50,42 @@
...
+    }
+  });
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    const curentValue = toFormat[i];
+
+    if (curentValue in dateObject) {
+      newFormatOfDate.push(dateObject[curentValue]);
+    } else {
+      if (toFormat[i] === SHORT_FORMAT) {
```
Moroz-Dmytro, line 75: Try to avoid nested `if`. You can use `continue`

Moroz-Dmytro (changes requested):
Good progress 👍 
Some comments below

PR #1471, opened 2023-08-07
add task solution

AngerDESTROYS:


yuliiaant (approved):
Good job!

src/formatDate.js
```diff
@@ -51,6 +51,37 @@
...
+  for (const datePart in objDateFrom) {
+    const dateValue = objDateFrom[datePart];
+
+    if (toFormat.includes(datePart)) {
+      formattedDateParts[toFormat.indexOf(datePart)] = dateValue;
+      continue;
+    }
+
+    if (datePart.length === 4) {
+      formattedDateParts[toFormat.indexOf('YY')] = dateValue.slice(2);
```
roman-mirzoian, line 73: It would be cool to put the "magic values" in separate constants.

src/formatDate.js
```diff
@@ -51,6 +51,37 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const fromSeparator = fromFormat[3];
+  const toSeparator = toFormat[3];
+  const dateParts = date.split(fromSeparator);
+  const formattedDateParts = [];
+  const objDateFrom = {};
```
roman-mirzoian, line 58: Here it is worth thinking about a better name.

roman-mirzoian (changes requested):

PR #1472, opened 2023-08-08
Develop

Yevheniia-Sid:


src/formatDate.js
```diff
@@ -50,7 +50,62 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const inputSeparator = fromFormat[3];
+  const outputSeparator = toFormat[3];
+  const formatDay = 'DD';
+  const formatMonth = 'MM';
+  const shortFormatYear = 'YY';
+  const maxYear = 30;
+  const nineteenthСentury = '19';
+  const twentiethCentury = '20';
```
vserhiichuk, line 60: I would change the names of these variables.If the year is 1950, it is not the nineteenth century

src/formatDate.js
```diff
@@ -50,7 +50,62 @@
...
+  let dayInput;
+
+  for (let i = 0; i < fromFormat.length; i++) {
+    if (fromFormat[i].includes(shortFormatYear)) {
+      if (inputDate[i].length === shortFormatYear.length) {
+        yearInput = (inputDate[i] < maxYear)
+          ? twentiethCentury + inputDate[i] : nineteenthСentury + inputDate[i];
+      } else {
+        yearInput = inputDate[i];
+      }
```
vserhiichuk, line 74: Looks very complicated. You use if inside an if + ternary operator
Consider using `switch` as a better option.

src/formatDate.js
```diff
@@ -50,7 +50,62 @@
...
+    }
+  }
+
+  const outputDateFormat = [];
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    if (toFormat[i].includes(shortFormatYear)) {
+      if (toFormat[i] === shortFormatYear) {
+        outputDateFormat[i] = yearInput.slice(-2);
+      } else {
```
vserhiichuk, line 94: Same situation, consider using `switch` for less nesting and better variable naming

vserhiichuk (changes requested):
Good job, but there are a few problems

PR #1473, opened 2023-08-08
make formatDate function

rostyslav48:


src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
+  const toSeparator = toFormat[3];
+  const dateArr = date.split(fromSeparator);
+  const dateObj = {};
+  const newDateArr = [];
+
+  fromFormat.pop();
+  toFormat.pop();
+
+  fromFormat.forEach((dateType, index) => {
+    if (dateType === 'YY') {
```
roman-mirzoian, line 63: It would be cool to put the "magic values" in separate constants.

src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparator = fromFormat[3];
+  const toSeparator = toFormat[3];
+  const dateArr = date.split(fromSeparator);
+  const dateObj = {};
+  const newDateArr = [];
+
+  fromFormat.pop();
+  toFormat.pop();
```
roman-mirzoian, line 60: The `pop()` method removes the last element from an array and returns that element, so that in this way the `fromSeparator` and `toSeparator` variables can be immediately defined.

src/formatDate.js
```diff
@@ -50,7 +50,32 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparator = fromFormat[3];
+  const toSeparator = toFormat[3];
+  const dateArr = date.split(fromSeparator);
+  const dateObj = {};
+  const newDateArr = [];
```
roman-mirzoian, line 57: It is worth considering a better naming for these variables.

roman-mirzoian (changes requested):

BudnikOleksii (approved):
Nicely done!

PR #1476, opened 2023-08-09
solution

mpytlyk:


src/formatDate.js
```diff
@@ -50,7 +50,46 @@
...
+
+  const formYear = (year >= 30 ? '19' : '20') + year;
+
+  for (const format of toFormat) {
+    if (format.includes('Y')) {
+      newDate.push(format.length < year.length
+        ? year.slice(2)
+        : format.length === year.length
+          ? year
+          : formYear);
```
maxim2310, line 80: its bad practice to use two combine ternary, better to use 'if' or something else

maxim2310 (changes requested):
Almost there) Just one thing needs to be fixed

etojeDenys (approved):
good job

PR #1478, opened 2023-08-11
solution

vladyslav-svorin:


src/formatDate.js
```diff
@@ -50,7 +50,33 @@
...
+
+  if (toFormat.includes('YYYY') && fromFormat.includes('YY')) {
+    if (eachDate[fromFormat.indexOf('YY')] < 30) {
+      correctDate[toFormat.indexOf('YYYY')]
+        = '20' + eachDate[fromFormat.indexOf('YY')];
+    }
+
+    if (eachDate[fromFormat.indexOf('YY')] >= 30) {
+      correctDate[toFormat.indexOf('YYYY')]
+        = '19' + eachDate[fromFormat.indexOf('YY')];
```
TarasHoliuk, line 75: you are doing almost the same twice

let's do smth like that:
`const century = eachDate[fromFormat.indexOf('YY')] < 30 ? '20' : '19'`
and then just set correct value with pre-defined century

Also, generally, you are using a lot of `toFormat.indexOf('YY')` such things. to make it clearer you may save it into variables (or smth like `fromFormat[fromFormat.length - 1]`) with meaningful names

TarasHoliuk (changes requested):

etojeDenys (approved):
lgtm

PR #1485, opened 2023-08-18
solution

Patryk-Buczkowski:


src/formatDate.js
```diff
@@ -49,8 +49,41 @@
  * @returns {string}
  */
 
+// console.log(formatDate(
+//   '20/02/18',
+//   ['YY', 'MM', 'DD', '/'],
+//   ['YYYY', 'MM', 'DD', '-'],
+// ));
```
OleziO, line 56: Remove all comments.

OleziO (changes requested):

src/formatDate.js
```diff
@@ -48,9 +48,37 @@
...
+        this['YY'] = this['YYYY'].slice(2, 4);
+        delete this['YYYY'];
+      } else if (this.hasOwnProperty('YY') && format === 'YYYY') {
+        this['YYYY'] = this['YY'] <= 20 ? '20' + this['YY'] : '19' + this['YY'];
+        delete this['YY'];
+      }
+    },
+  };
+
+  const toSeparator = toFormat[3];
```
darokrk, line 65: it should not be picked last item in array ? rather then 3 always ?

`const toSeparator = toFormat[toFormat.length -1];`

src/formatDate.js
```diff
@@ -48,9 +48,37 @@
...
+  const fromSeparator = fromFormat[3];
```
darokrk, line 66: same here ?

`const fromSeparator = fromFormat[fromFormat.length - 1];`

src/formatDate.js
```diff
@@ -48,9 +48,37 @@
...
-  // write code here
-}
 
+  const calendar = {
+    formatYear(format) {
+      if (this.hasOwnProperty('YYYY') && format === 'YY') {
+        this['YY'] = this['YYYY'].slice(2, 4);
+        delete this['YYYY'];
+      } else if (this.hasOwnProperty('YY') && format === 'YYYY') {
+        this['YYYY'] = this['YY'] <= 20 ? '20' + this['YY'] : '19' + this['YY'];
```
darokrk, line 59: you can simplify it by using template strings literals:

image

darokrk (changes requested):
looks good 👍🏽  left comments for improvements

darokrk (approved):
Good job 💪🏽  🔥

PR #1497, opened 2023-08-24
solution

AnastasiiaKriuchkova:


src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+    if (parseInt(mapping['YY']) < 30) {
+      mapping['YYYY'] = '20' + mapping['YY'];
+    } else {
+      mapping['YYYY'] = '19' + mapping['YY'];
+    }
+  }
+
+  if (mapping['YYYY'] && !mapping['YY']) {
+    mapping['YY'] = mapping['YYYY'].slice(-2);
+  }
```
VitaliyBondarenko1982, line 74:
```suggestion
if (mapping['YY']) {
    mapping['YYYY'] = parseInt(mapping['YY']) < 30
      ? '20' + mapping['YY']
      : '19' + mapping['YY'];
  }

  if (mapping['YYYY']) {
    mapping['YY'] = mapping['YYYY'].slice(-2);
  }
```
suggest to use ternary instead nested if/else
also looks like condition in if can be simplier

src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+
+  const mapping = {};
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    mapping[fromFormat[i]] = dateParts[i];
+  }
+
+  if (mapping['YY'] && !mapping['YYYY']) {
+    if (parseInt(mapping['YY']) < 30) {
+      mapping['YYYY'] = '20' + mapping['YY'];
```
VitaliyBondarenko1982, line 66: suggest to make constants outside of function instead hardcoded strings for `YY` and `YYYY` and use in all places

VitaliyBondarenko1982 (changes requested):
Good job.
Let's make a little refactoring.

etojeDenys (approved):
good job

PR #1499, opened 2023-08-26
task solution

BeginerAlex:


src/formatDate.js
```diff
@@ -50,7 +50,50 @@
...
+    }
+
+    if (fromFormat[i].includes('M')) {
+      month = dateArray[i];
+    }
+
+    if (fromFormat[i].includes('Y')) {
+      year = dateArray[i];
+    }
+  }
```
VitaliyBondarenko1982, line 72: istead of create three separate variable suggest to create one object, where key will be format, and value will bie dateArray value.
it will decrease if conditions in next loop

VitaliyBondarenko1982 (changes requested):
Good job.
Let's improve solution.

PR #1507, opened 2023-08-30
Solution

Cy90sure:


denys-danyliuk (changes requested):
try to use switch and improve your solution, now it's too complicated

src/formatDate.js
```diff
@@ -51,6 +51,53 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
```
OleziO, line 53: Remove all comments from the code.

OleziO (approved):

PR #1508, opened 2023-08-30
add task solution

mishakoshilka:


src/formatDate.js
```diff
@@ -50,7 +50,60 @@
...
+  let result = [];
+
+  const partsDate = date.split(fromFormat[fromFormat.length - 1]);
+
+  const info = {};
+
+  for (let i = 0; i < fromFormat.length; i++) {
+    switch (fromFormat[i]) {
+      case 'YY':
+      case 'YYYY':
```
volodymyr-soltys97, line 62: Consider will be better if you create a variable for all magic words and use them everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's impove your code

volodymyr-soltys97 (approved):
Well done 🔥

PR #1509, opened 2023-08-31
solution task

Butovych:


src/formatDate.js
```diff
@@ -50,7 +50,45 @@
...
+    fromDates[keyYearTo] = fromDates[keyYearFrom].toString().slice(2);
+  }
+
+  if (keyYearFrom.length < keyYearTo.length && fromDates[keyYearFrom] < 30) {
+    fromDates[keyYearTo] = `20${fromDates[keyYearFrom]}`;
+  }
+
+  if (keyYearFrom.length < keyYearTo.length && fromDates[keyYearFrom] >= 30) {
+    fromDates[keyYearTo] = `19${fromDates[keyYearFrom]}`;
+  }
```
etojeDenys, line 89: we can update our condition and use the ternary operator instead of the two `if`s

etojeDenys (changes requested):

src/formatDate.js
Butovych: ✅

src/formatDate.js
```diff
@@ -50,7 +50,43 @@
...
+
+  if (keyYearFrom.length > keyYearTo.length) {
+    fromDates[keyYearTo] = fromDates[keyYearFrom].toString().slice(2);
+  }
+
+  fromDates[keyYearTo] = keyYearFrom.length < keyYearTo.length && fromDates[keyYearFrom] < 30 ? `20${fromDates[keyYearFrom]}`
+    : fromDates[keyYearTo];
+
+  fromDates[keyYearTo] = keyYearFrom.length < keyYearTo.length && fromDates[keyYearFrom] >= 30 ? `19${fromDates[keyYearFrom]}`
+    : fromDates[keyYearTo];
```
etojeDenys, line 87: i mean we can make something like 
```
(fromDates[keyYearFrom] < 30 ? '20' : '19') + fromDates[keyYearFrom]
```

etojeDenys (changes requested):

src/formatDate.js
Butovych: When I use such a ternary operator, the tests fail because the second condition is not checked.

Снимок экрана 2023-08-31 в 21 01 09

src/formatDate.js
```diff
@@ -50,7 +50,43 @@
...
+
+  return `${fromDates[toFormat[0]]}${separator}` + `${fromDates[toFormat[1]]}${separator}` + `${fromDates[toFormat[2]]}`;
```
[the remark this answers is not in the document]
Butovych: But I can't use this method for objects.

etojeDenys (approved):
looks good

PR #1511, opened 2023-08-31
Solution

HachevskyyMarkiyan:


src/formatDate.js
```diff
@@ -50,7 +50,83 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const [ , , , fromSeparator ] = fromFormat;
```
etojeDenys, line 53:
```suggestion
  const fromSeparator = fromFormat[3];
```
looks much more readable

etojeDenys (approved):
lgtm

src/formatDate.js
HachevskyyMarkiyan: Done ✅

HachevskyyMarkiyan: Sorry. Incidentally requested review😅

PR #1512, opened 2023-09-01
done

MaxMaksym:


src/formatDate.js
```diff
@@ -50,7 +50,56 @@
...
+  let result = '';
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    switch (toFormat[i]) {
+      case SHORT_YEAR_ABBR:
+        result += year.length === 4 ? year.slice(-2) : year;
+        break;
+      case LONG_YEAR_ABBR:
+        result
+          += year.length === 4
```
polosanya, line 81: image

src/formatDate.js
```diff
@@ -50,7 +50,56 @@
...
+            ? year
+            : year >= MAX_YEAR
+              ? `19${year}`
+              : `20${year}`;
```
polosanya, line 85: image

polosanya (changes requested):
Great! Let's improve the formatting a bit)

src/formatDate.js
MaxMaksym: got it, this styles was applied by vs code, so I decided that it`s fine 😁

src/formatDate.js
MaxMaksym: got it, this styles was applied by vs code, so I decided that it`s fine 😁

src/formatDate.js
```diff
@@ -50,7 +50,58 @@
...
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    switch (toFormat[i]) {
+      case SHORT_YEAR_ABBR:
+        result += year.length === 4 ? year.slice(-2) : year;
+        break;
+      case LONG_YEAR_ABBR:
+        if (year.length === 4) {
+          result += year;
+          break;
+        } else if (year >= MAX_YEAR) {
```
polosanya, line 83: Try to avoid using `else...if` for better readability

polosanya (approved):
Nice!

PR #1519, opened 2023-09-02
solution

maximmorenko:


src/formatDate.js
```diff
@@ -50,7 +50,40 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  // console.log(date);
+  // console.log(fromFormat);
+  // console.log(toFormat);
```
Andriy-Shymkiv, line 55: please remove comments everywhere

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
+
+  const obj = {
```
Andriy-Shymkiv, line 57: please give meaningful naming to your variables, because `obj` is not

Andriy-Shymkiv (changes requested):
please add error handling. what if some of params will be not the format as we expecting? the function will work not as expected

maxim2310 (approved):
Good work!!!

PR #1520, opened 2023-09-04
Solution

denedweb:


lerastarynets (approved):
All requirements are met.
However pls consider ways to shorten your solution)

PR #1533, opened 2023-09-13
.add task solution

Chumakv:


src/formatDate.js
```diff
@@ -50,7 +50,74 @@
...
+        newFormatDate += month;
+        break;
+
+      case ('D'):
+        newFormatDate += day;
+        break;
+
+      default:
+        return;
+    }
```
sTorba24, line 95: looks like the same 2 loops, can't actions be combined?

sTorba24 (changes requested):
Well done! Lets improve it

nazarmatsevych (approved):
Well done!

PR #1537, opened 2023-09-15
solution

myuzkiv:


src/formatDate.js
```diff
@@ -50,7 +50,45 @@
...
+
+  const dateArray = date.split(dateSeparator);
+  const dateObject = {};
+  const result = [];
+
+  for (let i = 0; i < 3; i++) {
+    dateObject[fromFormat[i]] = dateArray[i];
+  }
+
+  if (fromFormat.includes('YY') && toFormat.includes('YYYY')) {
```
VitaliyBodnar, line 69: first condition is redundant, because if fromFormat.includes('YYYY') that it's also included 'YY'
```suggestion
  if (toFormat.includes('YYYY')) {
```

src/formatDate.js
```diff
@@ -50,7 +50,45 @@
...
+    }
+    delete dateObject.YY;
+  }
+
+  if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
+    dateObject.YY = dateObject.YYYY.slice(-2);
+    delete dateObject.YYYY;
+  }
+
+  for (let i = 0; i < 3; i++) {
```
VitaliyBodnar, line 83: 3 is a 'magic number', so please move it to the constant

VitaliyBodnar (changes requested):
Almost done!

src/formatDate.js
myuzkiv: it doesn't work like that

etojeDenys (approved):
good job

PR #1539, opened 2023-09-16
format date

iLavruk:


src/formatDate.js
```diff
@@ -50,7 +50,64 @@
...
+    if (fromFormatDateString === 'YYMMDD') {
+      const [year, month, day] = date.split(fromSeparator);
+
+      result = day + toSeparator + month + toSeparator + convertYear(year);
+    }
+  }
+
+  if (toFormatDateString === 'DDMMYY') {
+    if (fromFormatDateString === 'DDMMYYYY') {
+      const [day, month, year] = date.split(fromSeparator);
```
etojeDenys, line 96: try to write some generic solution(using the loop) instead of many conditions

etojeDenys (changes requested):

src/formatDate.js
iLavruk: done

src/formatDate.js
```diff
@@ -48,9 +48,30 @@
...
+  for (let i = 0; i < fromFormat.length; i++) {
+    fromatObj[fromFormat[i]] = dateToArr[i];
+  }
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    if (toFormat[i] in fromatObj) {
+      result.push(fromatObj[toFormat[i]]);
+    } else if (toFormat[i] === 'YY') {
+      result.push(fromatObj['YYYY'].slice(2));
+    } else if ((toFormat[i] === 'YYYY') & (fromatObj['YY'] < 30)) {
```
VitaliyBodnar, line 67: `&` ?
maybe do you mean `&&`

VitaliyBodnar (changes requested):

src/formatDate.js
iLavruk: fixed

VitaliyBodnar (approved):

PR #1541, opened 2023-09-16
Solution

mrogozhinskaya:


src/formatDate.js
```diff
@@ -50,7 +50,37 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const currentSeparator = fromFormat[3];
+  const desiredSeparator = toFormat[3];
+  const dateArray = date.split(currentSeparator);
+  const resultArray = [];
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    for (let j = 0; j < toFormat.length - 1; j++) {
```
TarasHoliuk, line 59: your solution is clear and understandable so approved
but generally better to avoid nested loops (it affects performance a lot) - you may solve this task w/o nested loops

TarasHoliuk (approved):

PR #1542, opened 2023-09-17
format date

whiovt:


src/formatDate.js
```diff
@@ -51,6 +51,46 @@
...
+    if (year < 30) {
+      year = '20' + year;
+    } else {
+      year = '19' + year;
+    }
+  }
+
+  if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
+    year = year.slice(2);
+  }
```
anastasiia-tilikina, line 77: let's create a separate function for getting year

src/formatDate.js
```diff
@@ -51,6 +51,46 @@
 
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const dateArr = date.split(fromFormat[3]);
```
anastasiia-tilikina, line 54: let's create a var for separator

anastasiia-tilikina (changes requested):
well done!

let's just improve your code a bit

PR #1547, opened 2023-09-20
Solution

GlibVK:


src/formatDate.js
```diff
@@ -50,7 +50,48 @@
...
+    switch (format) {
+      case 'YYYY':
+        if (fromFormat.indexOf('YYYY') !== -1) {
+          year = dateParts[fromFormat.indexOf('YYYY')];
+        } else {
+          year = dateParts[fromFormat.indexOf('YY')];
+        }
+
+        formattedDate += year.length === 2 ? (year < 30) ? '20' + year
+          : '19' + year : year;
```
VitaliyBodnar, line 67: 30, 19, 20 are a 'magic number', so make sense to move them to the variables
check other places

src/formatDate.js
```diff
@@ -50,7 +50,48 @@
...
+        break;
+      case 'YY':
+        if (fromFormat.indexOf('YYYY') !== -1) {
+          year = dateParts[fromFormat.indexOf('YYYY')];
+        } else {
+          year = dateParts[fromFormat.indexOf('YY')];
+        }
```
VitaliyBodnar, line 74: please use ternary operator here, as it will make a code shorter 
check other places

VitaliyBodnar (changes requested):

IvanFesenko (approved):

PR #1548, opened 2023-09-20
Solution

Anastasia-Bernada:


src/formatDate.js
```diff
@@ -50,7 +50,29 @@
...
+
+  if (dayPart.YYYY) {
+    dayPart.YY = dayPart.YYYY.slice(-2);
+  }
+
+  if (dayPart.YY >= 30) {
+    dayPart.YYYY = 19 + dayPart.YY;
+  } else {
+    dayPart.YYYY = 20 + dayPart.YY;
+  }
```
IvanFesenko, line 69: use trenary instead
```suggestion
  dayPart.YYYY = dayPart.YY >= 30 ? 19 + dayPart.YY : 20 + dayPart.YY
```

IvanFesenko (approved):
Looks good

src/formatDate.js
Anastasia-Bernada: Thanks for the advice

PR #1553, opened 2023-09-21
solution

Vitalikmy:


src/formatDate.js
```diff
@@ -50,7 +50,24 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparator = fromFormat.pop();
+  const toSeparator = toFormat.pop();
```
superpooperxxx, line 54: `pop()` is a mutating method and we should not mutate arguments of function

src/formatDate.js
```diff
@@ -50,7 +50,24 @@
...
+  const DateArray = date.split(fromSeparator);
```
superpooperxxx, line 55: We do not name regular variables with a first letter capital

superpooperxxx (changes requested):
Interesting solution🤗 But mutating is a bad practice

src/formatDate.js
polosanya: Not fixed

polosanya (changes requested):
Please make sure that you fixed all comments. 
Feel free to ask questions in qna\personal chat and tag mentors if you have any difficulties 🙂

polosanya (approved):
👍

PR #1557, opened 2023-09-21
add "format_date_advanced" task solution

mpohorenyi:


src/formatDate.js
```diff
@@ -50,7 +50,91 @@
...
+        month = oldFormat[i];
+        break;
+
+      case 'YY':
+        year = normalizeYear(oldFormat[i]);
+        break;
+
+      case 'YYYY':
+        year = normalizeYear(oldFormat[i]);
+        break;
```
polosanya, line 86: Switch case should always have default case for error handling

src/formatDate.js
```diff
@@ -50,7 +50,91 @@
...
+    }
+  }
+
+  for (let i = 0; i < SEPARATOR_INDEX; i++) {
+    switch (toFormat[i]) {
+      case 'DD':
```
polosanya, line 92: Would be great to use variables instead of hardcoded values to avoid duplicating

polosanya (changes requested):
Nice! Let's improve your code a bit)

polosanya (approved):
Nice!

PR #1559, opened 2023-09-22
add task solution

tilugdmytro:


src/formatDate.js
```diff
@@ -50,7 +50,36 @@
...
+    dateObject[fromFormat[i]] = dateArray[i];
+  }
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    const key = toFormat[i];
+    let value = dateObject[key];
+    const fullYearFormat = dateObject.YYYY;
+    const halfYearFormat = dateObject.YY;
+
+    if (value === undefined) {
```
polosanya, line 71:
```suggestion
    if (!value) {
```
You can simplify it

polosanya (approved):
Good job!

PR #1561, opened 2023-09-23
add task solution

ndovgyi:


src/formatDate.js
```diff
@@ -50,7 +50,37 @@
...
+      }
+
+      if (fromFormat[i] === 'YY' && toFormat[j] === 'YYYY') {
+        year = dateArray[i];
+
+        if (parseInt(year) < 30) {
+          resultArray[j] = `20${year}`;
+        } else {
+          resultArray[j] = `19${year}`;
+        }
```
Anastasiia-Svintsova, line 78: Use a ternary operator to avoid nested if statements

Anastasiia-Svintsova (approved):

PR #1563, opened 2023-09-24
1

viktor-kniahnytskyi:


src/formatDate.js
```diff
@@ -50,7 +50,46 @@
...
+      const SHORT_YEAR = 'YY';
+      const LONG_YEAR = 'YYYY';
+
+      if (toFormatItem === LONG_YEAR) {
+        toFormatItem = SHORT_YEAR;
+
+        const currentYear = stringItems[fromFormat.indexOf(SHORT_YEAR)];
+        const LAST_DIGITS = 30;
+        const CURRENT_CENTURY_YEAR = '20';
+        const PREVIOUS_CENTURY_YEAR = '19';
```
VitaliyBodnar, line 72: it's better practice to declare static variables in top of the file
so, please move these top

src/formatDate.js
```diff
@@ -50,7 +50,46 @@
...
+  for (let i = START_ITERATION; i < END_ITERATION; i++) {
+    let toFormatItem = toFormat[i];
+
+    if (!fromFormat.includes(toFormatItem)) {
+      const SHORT_YEAR = 'YY';
+      const LONG_YEAR = 'YYYY';
+
+      if (toFormatItem === LONG_YEAR) {
+        toFormatItem = SHORT_YEAR;
+
```
VitaliyBodnar, line 68: make sense to use switch case statement instead of multiple if else

VitaliyBodnar (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
+  const LAST_DIGITS = 30;
+  const CURRENT_CENTURY_YEAR = '20';
+  const PREVIOUS_CENTURY_YEAR = '19';
+  const stringItems = date.split(fromFormat[SEPARATOR_INDEX]);
+  const newDate = [];
+
+  for (let i = START_ITERATION; i < END_ITERATION; i++) {
+    let toFormatItem = toFormat[i];
+
+    switch (true) {
```
VitaliyBodnar, line 65: currently you have only one case to this switch, so don't make any sense to use switch case here
you can use switch case statement only when you have a multiple if else blocks

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
+      case !fromFormat.includes(toFormatItem):
+        const SHORT_YEAR = 'YY';
+        const LONG_YEAR = 'YYYY';
```
VitaliyBodnar, line 68: please also move SHORT_YEAR and LONG_YEAR to the bottom

VitaliyBodnar (changes requested):

src/formatDate.js
viktor-kniahnytskyi: I don’t understand where exactly I need to move it down

src/formatDate.js
```diff
@@ -50,7 +50,55 @@
...
+    const item = stringItems[fromFormatIndex];
+
+    newDate.push(item);
+  }
+
+  return newDate.join(toFormat[SEPARATOR_INDEX]);
 }
 
+const SHORT_YEAR = 'YY';
+const LONG_YEAR = 'YYYY';
```
VitaliyBodnar, line 102: sorry, my mistake
I mean to the top 
```suggestion
```

src/formatDate.js
```diff
@@ -50,7 +50,55 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const SEPARATOR_INDEX = 3;
+  const START_ITERATION = 0;
+  const END_ITERATION = 3;
+  const LAST_DIGITS = 30;
+  const CURRENT_CENTURY_YEAR = '20';
+  const PREVIOUS_CENTURY_YEAR = '19';
+  const stringItems = date.split(fromFormat[SEPARATOR_INDEX]);
```
VitaliyBodnar, line 59:
```suggestion
  const CURRENT_CENTURY_YEAR = '20';
  const PREVIOUS_CENTURY_YEAR = '19';
  const SHORT_YEAR = 'YY';
  const LONG_YEAR = 'YYYY';
  
  const stringItems = date.split(fromFormat[SEPARATOR_INDEX]);
```

VitaliyBodnar (approved):

PR #1568, opened 2023-09-27
Solution

avakiel:


src/formatDate.js
```diff
@@ -50,7 +50,43 @@
...
+    } else if (fromFormat[i][0] === 'D') {
+      day = dateArray[i];
+    }
+  }
+
+  for (let k = 0; k < toFormat.length; k++) {
+    if (toFormat[k][0] === 'D') {
+      result.push(day);
+    } else if (toFormat[k][0] === 'Y') {
+      if (toFormat[k] === 'YYYY' && year.length <= 2 && +year < 30) {
```
nabokok, line 75: It would be better not to use nesting 'if'

src/formatDate.js
```diff
@@ -50,7 +50,43 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const separatorFrom = fromFormat[fromFormat.length - 1];
+  const separatorTo = toFormat[toFormat.length - 1];
+  const dateArray = date.split(separatorFrom);
+  let day;
+  let month;
+  let year;
+  const result = [];
```
nabokok, line 59: you can group variables. first constи and then let

nabokok (commented):
Looks good! I've added a few comments

nabokok (changes requested):
Looks good! I've added a few comments

src/formatDate.js
```diff
@@ -50,7 +50,51 @@
...
+      case 'Y':
+        if (toFormat[k] === 'YYYY' && year.length <= 2 && +year < 30) {
+          result.push('20' + year);
+        } else if (toFormat[k] === 'YYYY' && year.length <= 2) {
+          result.push('19' + year);
+        } else if (toFormat[k] === 'YY' && year.length > 2) {
+          result.push(year.slice(2));
+        } else {
+          result.push(year);
+        }
```
DanilWeda, line 89: For all of these conditions, the best way use constants. Also, in this case, I see magic numbers. What does mean 2 in this case? What mean 30? Add some readability for this part)
Thanks!

src/formatDate.js
```diff
@@ -50,7 +50,51 @@
...
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const separatorFrom = fromFormat[fromFormat.length - 1];
+  const separatorTo = toFormat[toFormat.length - 1];
+  const dateArray = date.split(separatorFrom);
+  const result = [];
+  let day;
+  let month;
+  let year;
```
DanilWeda, line 59: If you want, try to rewrite without using let)

DanilWeda (changes requested):
All great! 
Check some improvement readability comment)

src/formatDate.js
```diff
@@ -50,7 +50,56 @@
...
+  const yearCentury = 30;
+  const fullYearFormat = 'YYYY';
+  const endYearFormat = 'YY';
+  const maxLength = 2;
+  let day;
+  let month;
+  let year;
+
+  for (let i = 0; i < fromFormat.length; i++) {
+    switch (fromFormat[i][0]) {
```
SerhiiKirik, line 66: always add `default` case in `switch / case` construction

SerhiiKirik (changes requested):
Well done 🔥 
Let's try to make a few improvements 🧑‍💻

DanilWeda (approved):
Greate job! 
Approved!

PR #1571, opened 2023-09-27
Develop

BetterSol:


src/formatDate.js
```diff
@@ -50,7 +50,63 @@
...
+  const formattedDate = [];
+
+  for (let i = 0; i < 3; i++) {
+    if (fromFormat[i] === 'YYYY') {
+      year = splittedDate[i];
+    }
+
+    if (fromFormat[i] === 'YY') {
+      year = splittedDate[i];
+    }
```
superpooperxxx, line 68: You can combine these conditions in one and also do the transformations from YYYY to YY or vice versa here

src/formatDate.js
```diff
@@ -50,7 +50,63 @@
...
+  const oldSeparator = fromFormat[3];
+  const newSeparator = toFormat[3];
+  const splittedDate = date.split(oldSeparator);
+  let year = 0;
+  let month = 0;
+  let day = 0;
+  const formattedDate = [];
+
+  for (let i = 0; i < 3; i++) {
+    if (fromFormat[i] === 'YYYY') {
```
superpooperxxx, line 62: Do not hardcode 3, it comes from date length without separator

superpooperxxx (changes requested):
Nice try🔥 I think we can slightly make your code cleaner

src/formatDate.js
BetterSol: 👍

src/formatDate.js
BetterSol: tried my best 💪

src/formatDate.js
```diff
@@ -50,7 +50,60 @@
...
+    if (fromFormat[i].includes('Y')) {
+      const yearFormat = fromFormat[i];
+
+      year = splittedDate[i];
+
+      if (yearFormat.length === YEAR_SHORT_LENGTH) {
+        if (year < CURRENT_YEAR_SHORT) {
+          year = `20${year}`;
+        } else {
+          year = `19${year}`;
```
polosanya, line 74: You can use ternary operator here

src/formatDate.js
```diff
@@ -50,7 +50,60 @@
...
+
+    if (toFormat[i] === 'YY') {
+      formattedDate[i] = year.slice(-2);
+    }
+
+    if (toFormat[i] === 'MM') {
+      formattedDate[i] = month;
+    }
+
+    if (toFormat[i] === 'DD') {
```
polosanya, line 101: Consider using switch...case instead, as you have a limit amount of values

polosanya (approved):
Well done!

PR #1572, opened 2023-09-28
formatDate

OlenaGor:


src/formatDate.js
```diff
@@ -50,7 +50,34 @@
...
+    let value = dateObject[key];
+    const fullYearFormat = dateObject.YYYY;
+    const halfYearFormat = dateObject.YY;
+
+    if (value === undefined) {
+      if (key === 'YY') {
+        value = fullYearFormat.slice(2);
+      } else {
+        value = +halfYearFormat < YEAR_PERIOD ? `20${halfYearFormat}` : `19${halfYearFormat}`;
+      }
```
VitaliyBodnar, line 74: make sense to use ternary operator here as it will make a code shorter

VitaliyBodnar (changes requested):

nazarmatsevych (approved):

PR #1575, opened 2023-09-29
Task solution

Sireus:


src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
+      year = element;
+    }
+
+    if (fromFormat[i].includes('M')) {
+      month = element;
+    }
+
+    if (fromFormat[i].includes('D')) {
+      day = element;
+    }
```
VitaliyBodnar, line 72: it's better to use switch case statement here instead of multiple if blocks

src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
+      if (format.length < year.length) {
+        newArr.push(year.slice(2));
+      } else if (format.length === year.length) {
+        newArr.push(year);
+      } else {
+        const prefix = year < 30 ? '20' : '19';
+        const fullYear = prefix + year;
+
+        newArr.push(fullYear);
+      }
```
VitaliyBodnar, line 86: add some spaces here, because code is no readable

VitaliyBodnar (changes requested):

src/formatDate.js
Sireus: > it's better to use switch case statement here instead of multiple if blocks

How can I do this without using 'includes'? Or how can I use this method in a switch case?

src/formatDate.js
Sireus: > add some spaces here, because code is no readable

Linter gets angry if you add space.

Viktor-Kost (approved):

PR #1576, opened 2023-09-29
task

MariiaSuper:


anastasiiavorobiova (changes requested):
Great work! Consider using constants instead of magic strings here, for example ```const FULL_YEAR = 'YYYY'; ```

anastasiiavorobiova (approved):
Great work!

PR #1578, opened 2023-10-02
Solution

YevgeniySavchuk:


src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+  const dateArray = date.split(separator);
+  const newDateArray = [];
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    if (fromFormat[i] === 'YY') {
+      if (+dateArray[i] < 30) {
+        dateArray[i] = dateArray[i].padStart(4, '20');
+      } else {
+        dateArray[i] = dateArray[i].padStart(4, '19');
+      }
```
VitaliyBodnar, line 65: make sense to use ternary operator here as it will make a code shorter 
check other places

VitaliyBodnar (changes requested):

src/formatDate.js
YevgeniySavchuk: code works with ternary operator, but linter does not let it to pass.
image

YevgeniySavchuk: image

maxim2310 (approved):
Good work!!!

PR #1582, opened 2023-10-04
'Solution'

TVChulkova:


src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    if (toFormat[i] === 'DD') {
+      dataNewFormat[i] = day;
+    }
+
+    if (toFormat[i] === 'MM') {
+      dataNewFormat[i] = month;
+    }
+
+    if (toFormat[i].includes('YY')) {
```
maxim2310, line 81: would be nice to use switch case here

maxim2310 (changes requested):
Good job! Let's improve your code a bit)

src/formatDate.js
TVChulkova: done

volodymyr-soltys97 (approved):
Great work 🔥

PR #1583, opened 2023-10-04
Solution

DariaSydorenko:


src/formatDate.js
```diff
@@ -50,7 +50,69 @@
...
+        newDateArray.push(day);
+
+        break;
+    }
+  }
+
+  const newSeparator = toFormat[3];
+  const newFormatDate = newDateArray.join(newSeparator);
+
+  return newFormatDate;
```
VitaliyBodnar, line 115: newFormatDate is redundant 
```suggestion
  return newDateArray.join(newSeparator);
```

VitaliyBodnar (changes requested):

etojeDenys (approved):
good job

PR #1584, opened 2023-10-04
Solution format date advanced

garmoni:


src/formatDate.js
```diff
@@ -50,7 +50,31 @@
...
+  const tempArr = [];
+
+  for (let i = 0; i < newArr.length; i++) {
+    if (i !== newArr.length) {
+      for (let j = 0; j < newArr.length; j++) {
+        if (fromFormat[i] === toFormat[j]) {
+          tempArr[j] = newArr[i];
+        } else if ((fromFormat[i] === 'YY' || fromFormat[i] === 'YYYY')
+        && (fromFormat[i] !== toFormat[j])) {
+          if (fromFormat[i] === 'YY' && toFormat[j] === 'YYYY') {
```
TarasHoliuk, line 63: this, and 69-71 rows
image

might be handled in a bit different way

try to overwrite it (you can use an object to store your values, etc)

TarasHoliuk (changes requested):

denys-danyliuk (approved):

PR #1587, opened 2023-10-05
Solution

kseniia-chepur:


denys-danyliuk (approved):
pretty clean solution, good job!

PR #1592, opened 2023-10-07
add task solution

natamrshn:


src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
+      year = newDate[i];
+    }
+
+    if (fromFormat[i].includes('M')) {
+      month = newDate[i];
+    }
+
+    if (fromFormat[i].includes('D')) {
+      day = newDate[i];
+    }
```
VitaliyBodnar, line 70: it's better to overwrite it with switch case statement as it will make a code cleaner

src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
+      if (format.length < year.length) {
+        newArr.push(year.slice(2));
+      } else if (format.length === year.length) {
+        newArr.push(year);
+      } else {
+        if (year >= 30) {
+          lengthYear = '19' + year;
+        } else {
+          lengthYear = '20' + year;
+        }
```
VitaliyBodnar, line 84: make sense to use ternary operator here as it will make a code shorter

VitaliyBodnar (changes requested):
Almost done!

loralevitska (approved):
Well done)

PR #1593, opened 2023-10-07
implementation

Daniilart01:


src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+
+  // if different year formats
+  const oldYear = dateToConvert[oldFormatIndexes[2]];
+
+  if (!oldFormatYearQuadro && newFormatYearQuadro) {
+    if (+oldYear < 30) {
+      dateToConvert[oldFormatIndexes[2]] = `20${oldYear}`;
+    } else {
+      dateToConvert[oldFormatIndexes[2]] = `19${oldYear}`;
+    }
```
VitaliyBodnar, line 69: make sense to use ternary operator here as it will make a code shorter

src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+  } else if (oldFormatYearQuadro && !newFormatYearQuadro) {
+    dateToConvert[oldFormatIndexes[2]] = oldYear.split('').slice(2).join('');
+  }
+
+  const newDate = [];
+
+  for (let i = 0; i <= 3; i++) {
```
VitaliyBodnar, line 76: 3 is a 'magic number', so make sense to move it to the constant

src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+  const oldYear = dateToConvert[oldFormatIndexes[2]];
+
+  if (!oldFormatYearQuadro && newFormatYearQuadro) {
+    if (+oldYear < 30) {
+      dateToConvert[oldFormatIndexes[2]] = `20${oldYear}`;
+    } else {
+      dateToConvert[oldFormatIndexes[2]] = `19${oldYear}`;
+    }
+  } else if (oldFormatYearQuadro && !newFormatYearQuadro) {
+    dateToConvert[oldFormatIndexes[2]] = oldYear.split('').slice(2).join('');
```
VitaliyBodnar, line 71: 2 is a 'magic number', so make sense to move it to the constant

src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
-  // write code here
+  const oldFormatIndexes = [];
+  const oldFormatYearQuadro
+    = getDMYIndexesSequence(oldFormatIndexes, fromFormat);
+  const newFormatIndexes = [];
+  const newFormatYearQuadro
+    = getDMYIndexesSequence(newFormatIndexes, toFormat);
+  const dateToConvert = date.split(fromFormat[fromFormat.length - 1]);
+
+  // if different year formats
```
VitaliyBodnar, line 61: remove redundant comments

src/formatDate.js
```diff
@@ -50,7 +50,59 @@
...
+ *
+ * @returns {boolean}
+*/
+function getDMYIndexesSequence(indexes, format) {
+  indexes.push(format.indexOf('DD'));
+  indexes.push(format.indexOf('MM'));
+
+  let yearIndex;
+
+  if ((yearIndex = format.indexOf('YY')) !== -1) {
```
VitaliyBodnar, line 97: it's not good idea to change variable in the condition for if
you can set it in if block or outside the if

VitaliyBodnar (changes requested):

etojeDenys (approved):
lgtm

PR #1596, opened 2023-10-09
Solution

ArtemSydun:


src/formatDate.js
```diff
@@ -50,7 +50,65 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const obj = {};
+  const splittedDate = date.split(fromFormat[fromFormat.length - 1]);
+  const newArr = [];
```
MrHunterBoi, line 55: Variable names should have a little bit more sense into them -_-

MrHunterBoi (changes requested):
Good job

src/formatDate.js
polosanya, line 55: ➕, try to provide more meaningful names than just `obj`, `newArr`

src/formatDate.js
```diff
@@ -50,7 +50,65 @@
...
+        year = obj[key];
+        break;
+      case 'YYYY':
+        year = obj[key];
+        break;
+      case 'MM':
+        month = obj[key];
+        break;
+      case 'DD':
+        day = obj[key];
```
polosanya, line 77: Switch...case should always have a default case for error handling

src/formatDate.js
```diff
@@ -50,7 +50,65 @@
...
+    }
+  }
+  newYear = year;
+
+  if (toFormat.includes('YYYY') && fromFormat.includes('YY')) {
+    if (year < 30) {
+      newYear = `20${year}`;
+    } else {
+      newYear = `19${year}`;
```
polosanya, line 86: You can use ternary operator here instead

src/formatDate.js
```diff
@@ -50,7 +50,65 @@
...
+    }
+  }
+
+  if (toFormat.includes('YY') && fromFormat.includes('YYYY')) {
+    newYear = `${year.toString().slice(-2)}`;
+  }
+
+  for (let i = 0; i <= 2; i++) {
```
polosanya, line 94: What is `2`? Avoid using "magic numbers"

polosanya (changes requested):
Nice! Let's improve your code a bit)

src/formatDate.js
```diff
@@ -50,7 +50,70 @@
...
+  }
+  newYear = year;
+
+  if (toFormat.includes('YYYY') && fromFormat.includes('YY')) {
+    newYear = year < 30 ? `20${year}` : `19${year}`;
+    // if (year < 30) {
+    //   newYear = `20${year}`;
+    // } else {
+    //   newYear = `19${year}`;
+    // }
```
BudnikOleksii, line 92: Don't leave comments

BudnikOleksii (approved):
Good job 👍 
Just don't leave comments

PR #1597, opened 2023-10-09
Solution

julmak:


src/formatDate.js
```diff
@@ -50,7 +50,31 @@
...
+  for (let i = 0; i < dateArr.length; i++) {
+    if (fromFormat[i] === 'YY') {
+      const front = dateArr[i] < 30 ? 20 : 19;
+
+      dateKeys['YYYY'] = `${front}${dateArr[i]}`;
+    }
+
+    if (fromFormat[i] === 'YYYY') {
+      dateKeys['YY'] = dateArr[i].slice(2);
+    }
```
maxim2310, line 68: now you are doing  dateKeys[fromFormat[i]] = dateArr[i]; this part in every cases and it can rewrite your previous actions
```suggestion
    if (fromFormat[i] === 'YY') {
      const front = dateArr[i] < 30 ? 20 : 19;

      dateKeys['YYYY'] = `${front}${dateArr[i]}`;
      continue;
    }

    if (fromFormat[i] === 'YYYY') {
      dateKeys['YY'] = dateArr[i].slice(2);
      continue;
    }
```

maxim2310 (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)

src/formatDate.js
julmak: @maxim2310, but when I put "continue" for the 2nd "if" 4 tests are failing, and as a result, there is no year received. I assume you meant to add "continue" only for the 1st "if". Correct me if I'm wrong.

src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
+    if (fromFormat[i] === 'YY') {
+      const front = dateArr[i] < 30 ? 20 : 19;
+
+      dateKeys['YYYY'] = `${front}${dateArr[i]}`;
+      continue;
+    }
+
+    if (fromFormat[i] === 'YYYY') {
+      dateKeys['YY'] = dateArr[i].slice(2);
+    }
```
VitaliyBodnar, line 69: it's better to use switch case statement here instead of multiple if blocks

src/formatDate.js
VitaliyBodnar: @julmak I suggest you to use switch case statement in this case

VitaliyBodnar (changes requested):
Almost done!

src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
+  for (let i = 0; i < dateArr.length; i++) {
+    switch (fromFormat[i]) {
+      case 'YY':
+        const front = dateArr[i] < 30 ? 20 : 19;
+
+        dateKeys['YYYY'] = `${front}${dateArr[i]}`;
+        break;
+
+      case 'YYYY':
+        dateKeys['YY'] = dateArr[i].slice(2);
```
VitaliyBodnar, line 68: missed break for case 'YYYY'

src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
+    }
```
VitaliyBodnar, line 69: it's also better to add default case 
```
default:
  break;
```

src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
+
+  for (let i = 0; i < dateArr.length; i++) {
+    switch (fromFormat[i]) {
+      case 'YY':
+        const front = dateArr[i] < 30 ? 20 : 19;
+
+        dateKeys['YYYY'] = `${front}${dateArr[i]}`;
+        break;
+
+      case 'YYYY':
```
VitaliyBodnar, line 67: it's better to move 'YYYY' and 'YY' to the constants as it will make a code cleaner

src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
+  const [, , , fromSeparator] = fromFormat;
+  const [, , , toSeparator] = toFormat;
+  const dateArr = date.split(fromSeparator);
+  const dateKeys = {};
+  const finalView = [];
+
+  for (let i = 0; i < dateArr.length; i++) {
+    switch (fromFormat[i]) {
+      case 'YY':
+        const front = dateArr[i] < 30 ? 20 : 19;
```
VitaliyBodnar, line 62: 30, 20 and 19 is a magic numbers, so make sense to move them to variables

VitaliyBodnar (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+  const [, , , fromSeparator] = fromFormat;
+  const [, , , toSeparator] = toFormat;
+  const dateArr = date.split(fromSeparator);
+  const dateKeys = {};
+  const finalView = [];
+  const fullYear = 'YYYY';
+  const halfYear = 'YY';
+  const defaultNum = 30;
+  const prevCentury = 19;
+  const curCentury = 20;
```
VitaliyBodnar, line 62: it's better to use UPPER_CASE for some static constants

VitaliyBodnar (approved):

PR #1600, opened 2023-10-11
Homework: JS Format Date Advanced

ViktorBykh:


src/formatDate.js
```diff
@@ -50,7 +50,291 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
```
sarakhmen, line 52: What do think about the size of this method? I'm assured it's feasible to move reusable logic into separate methods and also it's possible to decrease the number of lines drastically. Let's refactor it :)

sarakhmen (changes requested):
Let's improve your solution :)

sarakhmen (approved):
Good job :)

PR #1602, opened 2023-10-11
solution task

skaletskyi:


src/formatDate.js
```diff
@@ -48,9 +48,65 @@
...
+      if (year.length < toFormat[i].length) {
+        if (year < yearsLimitForCurrentCentury) {
+          newDate[i] = currentCentury + year;
+        }
+
+        if (year >= yearsLimitForCurrentCentury) {
+          newDate[i] = lastCentury + year;
+        }
+      }
+    }
```
denys-danyliuk, line 96: try to avoid using nested conditions

denys-danyliuk (approved):
GJ!

PR #1616, opened 2023-10-18
Solution

ErnestMk1:


src/formatDate.js
```diff
@@ -50,7 +50,38 @@
...
+    }
+  }
+
+  if (expandedDateFrom['YY'] && !(expandedDateTo['YY'])) {
+    if (expandedDateFrom['YY'] < 30) {
+      expandedDateTo['YYYY'] = Number(`${20}${expandedDateFrom['YY']}`);
+    } else {
+      expandedDateTo['YYYY'] = Number(`${19}${expandedDateFrom['YY']}`);
+    }
+  }
```
VitaliyBodnar, line 76: - 3, 19, 20, 30 are the magic numbers, so make sense to move them to the constants 
- YY and YYYY are the magic words, so make sense to move them to the constants 
- using nested ternary operator is a bad practice, it's better to move it to the variable

src/formatDate.js
```diff
@@ -50,7 +50,38 @@
...
+
+  if (expandedDateFrom['YYYY'] && !(expandedDateTo['YYYY'])) {
+    expandedDateTo['YY'] = Number(`${expandedDateFrom['YYYY'].slice(2)}`);
```
VitaliyBodnar, line 79:
```suggestion
    expandedDateTo['YY'] = expandedDateFrom['YYYY'].slice(2)
```

VitaliyBodnar (changes requested):

anastasiiavorobiova (approved):
Good work!

PR #1622, opened 2023-10-23
Solution

noraaudonnet:


nazarmatsevych (changes requested):
Good work, please check comments below

src/formatDate.js
```diff
@@ -50,7 +50,62 @@
...
+      if (year.length === 2) {
+        if (+year < 30) {
+          result[i] = +year + 2000;
+        }
+
+        if (+year >= 30) {
+          result[i] = +year + 1900;
+        }
+      }
+    }
```
sTorba24, line 105: here you also can use switch case

sTorba24 (changes requested):
Almost done!

sTorba24 (approved):
Well done!

PR #1626, opened 2023-10-25
Solution

Eater228:


witflash (approved):
Good imperative approach 👍🏼
You can also think out how to do it in a more declarative way (you can use `indexOf` method for example). But it's just an improvement.

Eater228:
> Good imperative approach 👍🏼 You can also think out how to do it in a more declarative way (you can use `indexOf` method for example). But it's just an improvement.

де я можу це використати?

PR #1636, opened 2023-10-30
solution

joanna-konopacka:


src/formatDate.js
```diff
@@ -42,15 +42,31 @@
...
+    dateCollector.YY = dateCollector.YYYY.slice(2);
+  } else {
+    const longYear =
+      dateCollector.YY < 30 ? 20 + dateCollector.YY : 19 + dateCollector.YY;
+
+    dateCollector.YYYY = longYear;
+  }
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    newDateFormat.push(dateCollector[toFormat[i]]);
```
vlasiuksasha, line 66: The second 'for' loop has the same iteration as the first 'for' loop. Consider if these can be combined to improve performance.

src/formatDate.js
```diff
@@ -42,15 +42,31 @@
...
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    dateCollector[fromFormat[i]] = dateParts[i];
+  }
+
+  if (!fromFormat.includes("YY")) {
+    dateCollector.YY = dateCollector.YYYY.slice(2);
+  } else {
+    const longYear =
+      dateCollector.YY < 30 ? 20 + dateCollector.YY : 19 + dateCollector.YY;
```
vlasiuksasha, line 60: Consider concatenating strings using template literals instead of the '+' operator. For example, `20${dateCollector.YY}` instead of `20 + dateCollector.YY`.

src/formatDate.js
```diff
@@ -42,15 +42,31 @@
...
+
+    dateCollector.YYYY = longYear;
```
vlasiuksasha, line 62: The variable 'longYear' is not being converted to a string before being assigned to 'dateCollector.YYYY'. Ensure that it is converted to a string, especially if 'dateCollector.YYYY' is expected to be a string.

src/formatDate.js
```diff
@@ -42,15 +42,31 @@
...
-  // write code here
+  const dateParts = date.split(fromFormat[3]);
+  const dateCollector = {};
+  const newDateFormat = [];
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    dateCollector[fromFormat[i]] = dateParts[i];
+  }
+
+  if (!fromFormat.includes("YY")) {
```
vlasiuksasha, line 56: The 'if' condition checks if 'fromFormat' does not include 'YY', but there is no equivalent check for 'YYYY'. This could lead to errors if 'fromFormat' includes 'YYYY' but not 'YY'.

vlasiuksasha (changes requested):
check the comments

Ivan-Shvorak (approved):
Good job

PR #1638, opened 2023-10-31
The solution for the first test

Patrykmclaren14:


src/formatDate.js
```diff
@@ -50,7 +50,44 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const modifyDate = date.split(fromFormat[fromFormat.length - 1]);
+  const newDate = [0, 0, 0];
+  const object = {};
+  let count = 0;
+
+  let year = 0;
+
+  for (const key of fromFormat) {
```
jacekkosiba, line 60: Name 'key' here may be a little misleading, since 'fromFormat' is an array, so it does not have keys

jacekkosiba: Maybe using an ordinary for loop would be better, then you could get rid of the 'count' variable, and just use 'i' instead

src/formatDate.js
```diff
@@ -50,7 +50,44 @@
...
+    if (count <= modifyDate.length - 1) {
+      object[key] = modifyDate[count];
+    }
+    count += 1;
+  }
+
+  for (let i = 0; i < modifyDate.length; i++) {
+    for (const key in object) {
+      if (key === 'YYYY' && toFormat[i] === 'YY' && object[key].length === 4) {
```
jacekkosiba, line 69: I may be wrong but it seem like the first condition (key === 'YYYY') is not needed if you als echeck object[key].length later

src/formatDate.js
```diff
@@ -50,7 +50,44 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const modifyDate = date.split(fromFormat[fromFormat.length - 1]);
+  const newDate = [0, 0, 0];
+  const object = {};
```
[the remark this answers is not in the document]
jacekkosiba: I meant the 'object' variable here

jacekkosiba (approved):

src/formatDate.js
JoyCoffeeAddict: Good spot! These checks actually check for the same thing

JoyCoffeeAddict (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,40 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateParts = date.split(fromFormat[fromFormat.length - 1]);
+  const newDate = [0, 0, 0];
+  const dateInfo = {};
+  let count = 0;
+
+  let year = 0;
+
+  for (const key of fromFormat) {
```
mvjl000, line 60: @Patrykmclaren14 please apply previous reviewers' suggestions here, and your task will be approved ;)

mvjl000 (changes requested):

PR #1643, opened 2023-11-02
Solution

zaura333:


Barosz30 (approved):
That's one Clean code! And REALY clever in some places, like checking only one number for less than 30. 

I'm jelous xD

JoyCoffeeAddict (approved):
Good job!

PR #1647, opened 2023-11-03
Develop

Barosz30:
So here we are. Any suggestions?

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -52,13 +52,6 @@
 function formatDate(date, fromFormat, toFormat) {
   // write code here
   const dataHolder = {
-    // get YYYY() {
-    //   if (this.YY < 30) {
-    //     return '20' + this.YY;
-    //   } else {
-    //     return '19' + this.YY;
-    //   }
-    // },
   };
 
   let newData = '';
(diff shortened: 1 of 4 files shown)


zaura333 (approved):
Looks good to me! Interesting idea with using an object

src/formatDate.js
```diff
@@ -51,6 +51,48 @@
...
+  const oldDateArray = date.split(oldSeparator);
+
+  // need to separate date to an array;
+  for (let i = 0; i < 3; i++) {
+    dataHolder[oldDateformats[i]] = oldDateArray[i];
+  }
+
+  // couldnt put getter in object, so i put it here;
+  // Oh well
+  if (!dataHolder.YY) {
```
JoyCoffeeAddict, line 73: It's good to compare your value to something, instead of casting it to boolean. This way you know what is checked here.
```suggestion
  if (dataHolder.YY === undefined) {
```

src/formatDate.js
```diff
@@ -51,6 +51,48 @@
...
+    const year = dataHolder.YYYY;
+
+    dataHolder.YY = year.slice(-2);
+  }
+
+  if (!dataHolder.YYYY) {
```
JoyCoffeeAddict, line 79: As mentioned above.

JoyCoffeeAddict (changes requested):
Almost perfect!

PR #1648, opened 2023-11-03
add solution

MKLNKs:


src/formatDate.js
```diff
@@ -49,8 +49,53 @@
...
+          = dateArray[fromFormat.indexOf('YY')];
+
+        if (dateArray[fromFormat.indexOf('YY')] < 30) {
+          toFormatDate[toFormat.indexOf('YYYY')]
+            = 20 + toFormatDate[toFormat.indexOf('YYYY')];
+        } else {
+          toFormatDate[toFormat.indexOf('YYYY')]
+            = 19 + toFormatDate[toFormat.indexOf('YYYY')];
+        }
+      }
```
VitaliyBodnar, line 92: - 19, 20 and 30 are a magic numbers, so make sense to move them to the constants 
- make sense to use ternary operator here as it will make a code shorter

src/formatDate.js
```diff
@@ -49,8 +49,53 @@
...
+    }
+  }
+
+  const convertedDate = toFormatDate.join(toFormat[toFormat.length - 1]);
+
+  return convertedDate;
```
VitaliyBodnar, line 98: redundant variable 

```suggestion
  return toFormatDate.join(toFormat[toFormat.length - 1]);
```

src/formatDate.js
```diff
@@ -49,8 +49,53 @@
...
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    for (let j = 0; j < toFormat.length - 1; j++) {
+      if (fromFormat[i] === toFormat[j]) {
+        toFormatDate[j] = dateArray[i];
+      }
+
+      if (
+        fromFormat.includes('YYYY')
+        && toFormat.includes('YY')
```
VitaliyBodnar, line 67: `fromFormat.includes('YYYY') && toFormat.includes('YY')` this condition is duplicated, so make sense to create separate if for that

VitaliyBodnar (changes requested):

src/formatDate.js
MKLNKs: npm test doesn't allow me to use ternary operator. I didn't change this

PR #1652, opened 2023-11-06
Solution

Miishca:


src/formatDate.js
```diff
@@ -50,7 +50,28 @@
...
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    dateCollector[fromFormat[i]] = dateParts[i];
+  }
+
+  if (!fromFormat.includes('YY')) {
+    dateCollector.YY = dateCollector.YYYY.slice(2);
+  } else {
+    const longYear
+      = dateCollector.YY < 30 ? 20 + dateCollector.YY : 19 + dateCollector.YY;
```
VitaliyBodnar, line 65: 20,19,30 are the magic numbers, so make sense to move them to the constants

src/formatDate.js
```diff
@@ -50,7 +50,28 @@
...
+  const newDateFormat = [];
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    dateCollector[fromFormat[i]] = dateParts[i];
+  }
+
+  if (!fromFormat.includes('YY')) {
+    dateCollector.YY = dateCollector.YYYY.slice(2);
+  } else {
+    const longYear
```
VitaliyBodnar, line 64: LongYear is a redundant variable, because you can this value directly to dataCollector.YYYY

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):
Good job!

PR #1654, opened 2023-11-09
done task

Anton-Kuchmasov:


src/formatDate.js
```diff
@@ -50,7 +50,38 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateInput = date;
+  const oldFormat = fromFormat;
+  const newFormat = toFormat;
```
polosanya, line 55: What's the purpose of creating these variables? You can just use parameters directly

src/formatDate.js
```diff
@@ -50,7 +50,38 @@
...
+  const oldDelimiter = oldFormat.splice(3, 1);
```
polosanya, line 56:
```suggestion
  const oldDelimiter = oldFormat[3];
```
Splice is a mutating method. You shouldn't mutate the origin array.
Just get access to the needed element using `[]` or `at` method

src/formatDate.js
```diff
@@ -50,7 +50,38 @@
...
+  let resultDate = '';
+  const resultDateArray = [];
+
+  for (let i = 0; i <= oldFormat.length; i++) {
+    dateObj[oldFormat[i]] = dateArray[i];
+  }
+
+  if (newFormat.includes('YYYY') && (oldFormat.includes('YY'))) {
+    dateObj.YYYY = (dateObj.YY > 29 ? '19' : '20') + dateObj.YY;
+    delete dateObj.YY;
```
polosanya, line 70: Do you really need `dateObj`? Consider adding elements to the array straight away

polosanya (changes requested):
Nice start! Let's try to simplify your code a bit)

YehorHeneralov (approved):
well Done!)

BudnikOleksii (approved):
Good work 👍

PR #1655, opened 2023-11-10
done task

olegnovikov10:


src/formatDate.js
```diff
@@ -49,8 +49,43 @@
  * @returns {string}
  */
 
+const YY = 'YY';
+const YYYY = 'YYYY';
+
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const objDate = {};
```
Anton-Kuchmasov, line 56: is it necessary to use Object? 


asking cause i had the same solution, but mentor didn't approve it until i remove an object
and use ONLY arrays to resolve a problem

src/formatDate.js
```diff
@@ -49,8 +49,43 @@
...
+  const delimiterToFormat = toFormat[3];
+  const arrDate = date.split(delimiterFromFormat);
+
+  arrDate.map((item, i) => {
+    objDate[fromFormat[i]] = item;
+  });
+
+  const keysDate = Object.keys(objDate);
+
+  toFormat.map((item, i) => {
```
Anton-Kuchmasov, line 68: rather use forEach 

map is returning something, but seems not in that case


upd.: you can remove map method as well as returning true for 2 times 

and use only 2 if's conditions to check, if needed to swap YY and YYYY as well

Anton-Kuchmasov (changes requested):
nice job, but some issues need to be resolved!

Anton-Kuchmasov (approved):

src/formatDate.js
```diff
@@ -49,8 +49,39 @@
  * @returns {string}
  */
 
+const YY = 'YY';
+const YYYY = 'YYYY';
+
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const objDate = {};
+  const arrDateResult = [];
```
roman-mirzoian, line 57:
```suggestion
  const formattedDate = [];
```

src/formatDate.js
```diff
@@ -49,8 +49,39 @@
...
+  const delimiterFromFormat = fromFormat[3];
+  const delimiterToFormat = toFormat[3];
+  const arrDate = date.split(delimiterFromFormat);
```
roman-mirzoian, line 60:
```suggestion
  const splittedDate = date.split(delimiterFromFormat);
```

src/formatDate.js
```diff
@@ -49,8 +49,39 @@
  * @returns {string}
  */
 
+const YY = 'YY';
+const YYYY = 'YYYY';
```
roman-mirzoian, line 53:
```suggestion
const MONTH_SCHEMA = 'YY';
const YEAR_SCHEMA = 'YYYY';
```

src/formatDate.js
```diff
@@ -49,8 +49,39 @@
...
+
+  const keysDate = Object.keys(objDate);
+
+  toFormat.forEach((item, i) => {
+    if (item === YY && !keysDate.includes(item)) {
+      arrDateResult.push((objDate[YYYY]).slice(2));
+    }
+
+    if (item === YYYY && !keysDate.includes(item)) {
+      const year = `${objDate[YY] < 30 ? 20 : 19}${objDate[YY]}`;
```
roman-mirzoian, line 74: Avoid using "magic values" (30, 20, 19?)

src/formatDate.js
roman-mirzoian: @Anton-Kuchmasov the option using an object is also suitable, it will look better simply through an array

src/formatDate.js
```diff
@@ -49,8 +49,39 @@
  * @returns {string}
  */
 
+const YY = 'YY';
+const YYYY = 'YYYY';
+
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const objDate = {};
```
roman-mirzoian, line 56: Bad naming

roman-mirzoian (approved):
Well done 👍

PR #1657, opened 2023-11-12
formatDate_first_edition

oleksandr-rudiuk:


src/formatDate.js
```diff
@@ -50,7 +50,70 @@
...
+    if (oldIndexDate === 'YY') {
+      newYear = date.slice(n, (n + 2));
+      shortOldFormat = 'yes';
+      n = n + 3;
+
+      if (+newYear < 30) {
+        longNewYear = '20' + newYear;
+      } else {
+        longNewYear = '19' + newYear;
+      }
```
VitaliyBodnar, line 75: - 20, 19, 30 are the magic numbers, so make sense to move them to constants 
- make sense to use ternary operator as it will make a code shorter 
Check other places

src/formatDate.js
```diff
@@ -50,7 +50,70 @@
...
+  let shortOldFormat = '';
+  let longNewYear = '';
+
+  for (const oldIndexDate of fromFormat) {
+    if (oldIndexDate === 'MM') {
+      newMonth = date.slice(n, (n + 2));
+      n = n + 3;
+    }
+
+    if (oldIndexDate === 'YY') {
```
VitaliyBodnar, line 66: Use switch case statement instead of multiple if blocks

src/formatDate.js
```diff
@@ -50,7 +50,70 @@
...
+    }
+
+    if (oldIndexDate === 'YYYY') {
+      newYear = date.slice(n, (n + 4));
+      shortNewYear = date.slice((n + 2), (n + 4));
+      n = n + 5;
+    }
+
+    if (oldIndexDate === 'DD') {
+      newDay = date.slice(n, (n + 2));
```
VitaliyBodnar, line 85: YY, YYYY, DD and other are a magic words, so pleas move them to the constants

VitaliyBodnar (changes requested):

VitaliyBondarenko1982 (approved):
Looks good.

oleksandr-rudiuk:
Thank you.

PR #1659, opened 2023-11-13
Solution

Devery11:
Hello, can u explain to me how to delete second commit: "solutionn" (w/ double 'n' :))? Thx for response


lerastarynets (changes requested):
Pls just make another commit with changes that u have in the first commit, so the second one will be overwritten)
Then ask for review again

PR #1662, opened 2023-11-13
add task solution

reznik-denis:


src/formatDate.js
```diff
@@ -50,7 +50,38 @@
...
+  const dateArray = date.split(fromFormat[`${fromFormat.length - 1}`]);
+  const dateObject = {};
+  const resultArray = [];
+
+  for (let i = 0; i < dateArray.length; i++) {
+    dateObject[fromFormat[i]] = dateArray[i];
+  }
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    if (toFormat[i] === 'YY') {
```
TarasHoliuk, line 62: such a big tree of `else` and `if else` blocks is difficult to read and understand

There are a few ways to simplify it
One of them is to cover all possible cases in the first loop (in other words add `YYYY` and `YY` values to the date object, no matter what is in `toFormat`). It requires much less things to check

TarasHoliuk (changes requested):

src/formatDate.js
reznik-denis: I rewrite code, make it simple. thanks

maxim2310 (approved):
Good work!!!

PR #1670, opened 2023-11-17
js

nazardudnyk:


src/formatDate.js
```diff
@@ -50,7 +50,49 @@
...
+      year = newDate[i];
+    }
+
+    if (fromFormat[i].includes('M')) {
+      month = newDate[i];
+    }
+
+    if (fromFormat[i].includes('D')) {
+      day = newDate[i];
+    }
```
VitaliyBodnar, line 70: It's better to use switch case instead multiple if blocks

src/formatDate.js
```diff
@@ -50,7 +50,49 @@
...
+  }
+
+  for (const format of toFormat) {
+    if (format.includes('Y')) {
+      if (format.length < year.length) {
+        newFormat.push(year.slice(2));
+      } else if (format.length === year.length) {
+        newFormat.push(year);
+      } else {
+        const fullYear = (year >= 30 ? '19' : '20') + year;
```
VitaliyBodnar, line 80: 19, 20 and 30 are "magic numbers",  so make sense to move these to the constants

src/formatDate.js
```diff
@@ -50,7 +50,49 @@
...
+        newFormat.push(fullYear);
+      }
+    }
+
+    if (format.includes('M')) {
+      newFormat.push(month);
+    }
+
+    if (format.includes('D')) {
+      newFormat.push(day);
```
VitaliyBodnar, line 91: M, D, Y and other are a "magic words", so make sense to move it to the constants

VitaliyBodnar (changes requested):

src/formatDate.js
```diff
@@ -48,9 +48,66 @@
...
+  const separator = fromFormat[fromFormat.length - 1];
+  const newDate = date.split(separator);
+  const newSeparator = toFormat[toFormat.length - 1];
+  const newFormat = [];
+
+  let year, month, day;
+
+  for (let i = 0; i < fromFormat.length; i++) {
+    switch (fromFormat[i]) {
+      case 'YYYY':
```
anastasiiavorobiova, line 68: You also can use constants here

anastasiiavorobiova (approved):
Great work!

PR #1676, opened 2023-11-22
first solution

dima91020:


src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
+      } else {
+        toFormatObj[part] = '19' + fromFormatObj['YY'];
+      }
+    } else {
+      toFormatObj[part] = fromFormatObj[part];
+    }
+  }
+
+  const newDateParts = [toFormatObj[toFormat[0]],
+    toFormatObj[toFormat[1]], toFormatObj[toFormat[2]]];
```
denys-danyliuk, line 76: come up with a better solution

denys-danyliuk (approved):

PR #1678, opened 2023-11-25
Solution

KaterynaEV:


src/formatDate.js
```diff
@@ -50,7 +50,61 @@
...
+      year = splitDate[i];
+    }
+
+    if (fromFormat[i] === 'MM') {
+      month = splitDate[i];
+    }
+
+    if (fromFormat[i] === 'DD') {
+      day = splitDate[i];
+    }
```
VitaliyBodnar, line 70: - please use switch case statement instead of multiple if blocks 
- YYYY, MM, DD are magic words, so it's make sense to move them to constant

src/formatDate.js
```diff
@@ -50,7 +50,61 @@
...
+
+        break;
+
+      case ('YYYY'):
+        if (year.length === 4) {
+          formated.push(year);
+        } else if (year < 30) {
+          formated.push(20 + year);
+        } else {
+          formated.push(19 + year);
```
VitaliyBodnar, line 91: 4, 19, 20 and 30 are the magic numbers, so make sense to move the to the constants

src/formatDate.js
```diff
@@ -50,7 +50,61 @@
...
+        }
+
+        break;
+
+      case ('YY'):
+        if (year.length === 2) {
+          formated.push(year);
+        } else {
+          formated.push(year.slice(2));
+        }
```
VitaliyBodnar, line 101: make sense to use ternary operator here as it will make a code shorter

VitaliyBodnar (changes requested):

VitaliyBodnar (changes requested):
.

src/formatDate.js
```diff
@@ -50,7 +50,55 @@
...
+  let day;
+
+  const YEAR_THRESHOLD = 30;
+  const PREFIX_20 = 20;
+  const PREFIX_19 = 19;
+
+  for (let i = 0; i < fromFormat.length; i++) {
+    switch (fromFormat[i]) {
+      case 'YYYY':
+      case 'YY':
```
anastasiiavorobiova, line 66: Consider using variables with meaningful names instead of magic strings

src/formatDate.js
```diff
@@ -50,7 +50,55 @@
...
+        year = splitDate[i];
+        break;
+
+      case 'MM':
```
anastasiiavorobiova, line 70: The same issue here

src/formatDate.js
```diff
@@ -50,7 +50,55 @@
...
+        month = splitDate[i];
+        break;
+
+      case 'DD':
```
anastasiiavorobiova, line 74: Also, could you please fix this further in the code?

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+  const dateObject = {};
+
+  for (let i = 0; i < dateArrFromFormat.length; i++) {
+    dateObject[fromFormat[i]] = dateArrFromFormat[i];
+  }
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    if (!(toFormat[i] in dateObject)) {
+      switch (toFormat[i]) {
+        case 'YYYY':
```
Esceype, line 64: Move `YYYY` and `YY` to variable please, it still not fixed from previous comments

Esceype (changes requested):

PR #1679, opened 2023-11-25
format date solution

ly-manka:


src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+      const yearPrefix = +dateComponents[SHORT_YEAR_FROMAT] < 30 ? '20' : '19';
+
+      newFormatDate.push(yearPrefix + dateComponents[SHORT_YEAR_FROMAT]);
+
+      continue;
+    }
+
+    if (value === SHORT_YEAR_FROMAT) {
+      newFormatDate.push(dateComponents[LONG_YEAR_FROMAT].slice(-2));
+    }
```
VitaliyBodnar, line 82: it's better to use switch case statement instead of multiple if blocks

src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+  }
+
+  const formattedDate = newFormatDate.join(toFormat[toFormat.length - 1]);
+
+  return formattedDate;
```
VitaliyBodnar, line 87: redundant variable 
```suggestion
  return newFormatDate.join(toFormat[toFormat.length - 1]);;
```

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):
Good job!

PR #1683, opened 2023-11-26
FormateDates try#1

VolodymyrLiubyn:


src/formatDate.js
```diff
@@ -51,6 +51,32 @@
...
+    dateObject[fromFormat[i]] = dateArray[i];
+  };
+
+  if (!('YY' in dateObject)) {
+    dateObject['YY'] = dateObject['YYYY'].slice(2);
+  };
+
+  if (!('YYYY' in dateObject)) {
+    if (dateObject['YY'] < '30') {
+      dateObject['YYYY'] = '20' + [dateObject['YY']];
```
VitaliyBodnar, line 67: YYYY and YY are magic words, so make sense to move them to constants

src/formatDate.js
```diff
@@ -51,6 +51,32 @@
...
+    } else {
+      dateObject['YYYY'] = '19' + [dateObject['YY']];
```
VitaliyBodnar, line 69: The same with 19, 20 and 30

src/formatDate.js
```diff
@@ -51,6 +51,32 @@
...
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    dateObject[fromFormat[i]] = dateArray[i];
+  };
+
+  if (!('YY' in dateObject)) {
+    dateObject['YY'] = dateObject['YYYY'].slice(2);
+  };
+
+  if (!('YYYY' in dateObject)) {
+    if (dateObject['YY'] < '30') {
```
VitaliyBodnar, line 66: Make sense to use ternary operator instead of if else as it will make a code shorter

VitaliyBodnar (changes requested):

anastasiiavorobiova (approved):
Awesome work!

PR #1691, opened 2023-12-01
solution

oskushnir:


src/formatDate.js
```diff
@@ -50,7 +50,39 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const parts = date.split(fromFormat[fromFormat.length - 1]);
+  let newDate = '';
+
+  for (let i = 0; i < toFormat.length; i++) {
+    if (toFormat[i].includes('Y')) {
```
anastasiiavorobiova, line 57: Could you consider using a switch here?
Also, please create constants with meaningful names instead of magic strings and numbers.

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/formatDate.js
```diff
@@ -50,7 +50,52 @@
...
+
+  const YEAR_FULL = 'YYYY';
+  const YEAR_SHORT = 'YY';
+  const MONTH = 'MM';
+  const DAY = 'DD';
+
+  for (let i = 0; i < toFormat.length; i++) {
+    switch (true) {
+      case toFormat[i].includes(YEAR_SHORT):
+        if (fromFormat.includes(YEAR_FULL) && toFormat[i] === YEAR_SHORT) {
```
l4st1m0za, line 64: The 'includes' function is used multiple times for the same variables, which is not efficient. Consider assigning the result to a variable and reusing it instead.

l4st1m0za (changes requested):
Great job! You are preatty close!

l4st1m0za (approved):
Awesome job!

PR #1694, opened 2023-12-02
Solution

afesh4ak:


src/formatDate.js
```diff
@@ -50,7 +50,48 @@
...
+  const parts = date.split(fromFormat[fromFormat.length - 1]);
+  const resultMap = new Map();
+
+  fromFormat.forEach((part, index) => {
+    resultMap.set(part, parts[index]);
+  });
+
+  const result = toFormat
+    .map((part) => {
+      if (part.includes('YY')) {
```
anastasiiavorobiova, line 62: Consider using variables with meaningful names instead of magic strings

src/formatDate.js
```diff
@@ -50,7 +50,48 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const parts = date.split(fromFormat[fromFormat.length - 1]);
+  const resultMap = new Map();
+
```
anastasiiavorobiova, line 55:
```suggestion
const LONG_YEAR_FORMAT = 'YYYY';
```

src/formatDate.js
```diff
@@ -50,7 +50,48 @@
...
+  fromFormat.forEach((part, index) => {
+    resultMap.set(part, parts[index]);
+  });
+
+  const result = toFormat
+    .map((part) => {
+      if (part.includes('YY')) {
+        let year = resultMap.get('YYYY');
```
anastasiiavorobiova, line 63:
```suggestion
        let year = resultMap.get(LONG_YEAR_FORMAT);
```

src/formatDate.js
```diff
@@ -50,7 +50,48 @@
...
+        }
+
+        return year;
+      }
+
+      return resultMap.get(part);
+    })
+    .join(toFormat[toFormat.length - 1]);
+
+  return result.endsWith(toFormat[toFormat.length - 1])
```
anastasiiavorobiova, line 92: Consider simplifying this logic

anastasiiavorobiova (changes requested):
Great work! Added a few comments to fix

src/formatDate.js
afesh4ak: I use this logic to slice last separator, if date ends with separator

PR #1695, opened 2023-12-02
Solution

AndriyBabiy:


src/formatDate.js
```diff
@@ -50,7 +50,55 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const FIRST_SPLITTER = fromFormat[3];
+
```
anastasiiavorobiova, line 54:
```suggestion
const LONG_YEAR_FORMAT = 'YYYY';
```

src/formatDate.js
```diff
@@ -50,7 +50,55 @@
...
+  const dateSplit = date.split(FIRST_SPLITTER);
+
+  let year = '';
+
+  if (fromFormat.indexOf('YYYY') !== -1) {
```
anastasiiavorobiova, line 59:
```suggestion
  if (fromFormat.indexOf(LONG_YEAR_FORMAT) !== -1) {
```
Consider using variables with meaningful names instead of using magic strings

src/formatDate.js
```diff
@@ -50,7 +50,55 @@
...
+    year = dateSplit[fromFormat.indexOf('YYYY')];
```
anastasiiavorobiova, line 60: Also, consider fixing it further in the code

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/formatDate.js
AndriyBabiy: ✅

maxim2310 (approved):
Good work!!!

PR #1707, opened 2023-12-10
solution

pivkopa:


src/formatDate.js
```diff
@@ -50,7 +50,31 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateArr = date.split(fromFormat[3]);
+  let result = toFormat;
```
artem8746, line 54: better `toFormat.slice(0,3);` to not pop last element before return

src/formatDate.js
```diff
@@ -50,7 +50,46 @@
...
+    } else {
+      result[result.indexOf(YEAR_LONG_IDENTIFIER)]
+            = '20' + dateArr[fromFormat.indexOf(YEAR_SHORT_IDENTIFIER)];
+    }
+  } else if (fromFormat.includes(YEAR_LONG_IDENTIFIER)
+              && result.includes(YEAR_SHORT_IDENTIFIER)) {
+    result[result.indexOf(YEAR_SHORT_IDENTIFIER)]
+  = (dateArr[fromFormat.indexOf('YYYY')][2]
+  + dateArr[fromFormat.indexOf('YYYY')][3]);
+  }
```
dinaminahi, line 88: Instead of using multiple conditional blocks, consider a more streamlined approach. Use loops or simpler conditional statements to handle the transformation.

You can create an object that correlates the elements in the fromFormat and toFormat arrays. This object will help in mapping the elements directly without numerous conditional checks. For example 

```
 const keysDates = {};

 for (let i = 0; i < fromFormat.length - 1; i++) {
    keysDates[fromFormat[i]] = dateParts[i];
  }
```

dinaminahi (changes requested):
Great job 👍  Let's try to make you code simplier and more readable

PR #1711, opened 2023-12-12
Format Date Advanced

yurovych:


maxim2310 (changes requested):
its working but not a good way to use 3 nested loop for solving, would be nice to optimize it

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
+}
+
+function cycle(
+  fromFormatCopyCycle,
+  toFormatCycle,
+  resultArrayCycle,
+  factDateCycle
+) {
+  for (let i = 0; i < fromFormatCopyCycle.length; i++) {
+    for (let j = 0; j < toFormatCycle.length; j++) {
```
anastasiiavorobiova, line 98:
```suggestion
    
```

anastasiiavorobiova: Consider using just one cycle here

src/formatDate.js
```diff
@@ -50,7 +50,57 @@
...
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const factDate = date.split(fromFormat[fromFormat.length - 1]);
+  const fromFormatCopy = [...fromFormat];
+  const resultArray = [...toFormat];
+
+  fromFormatCopy.splice(-1, 1);
+  resultArray.splice(-1, 1);
+
+  if ((fromFormatCopy.includes('YYYY') && toFormat.includes('YYYY'))
```
anastasiiavorobiova, line 60:
```suggestion
const YEAR_LONG = 'YYYY';

  if ((fromFormatCopy.includes(YEAR_LONG) && toFormat.includes(YEAR_LONG))
```
Consider using variables with meaningful names instead of magic strings

anastasiiavorobiova (changes requested):
Great work! added a few comments to fix

SanyaBratashchuk (approved):

PR #1715, opened 2023-12-14
Solution

DenLys21:


src/formatDate.js
```diff
@@ -42,15 +42,50 @@
...
+  for (let i = 0; i < arrayDate.length; i++) {
+    objectData[fromFormat[i]] = arrayDate[i];
+  }
+
+  if (fromFormat.includes('YY') && toFormat.includes('YYYY')) {
+    let century = 0;
+
+    newKey = 'YYYY';
+
+    if (objectData['YY'] < 30) {
```
TarasHoliuk, line 67: good case to use ternary instead of `if else`

src/formatDate.js
```diff
@@ -42,15 +42,50 @@
...
+//  *
+//  * @returns {string}
+//  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const result = [];
+  const arrayDate = date.split(fromFormat[fromFormat.length - 1]);
+  const objectData = {};
+  let newKey = '';
```
TarasHoliuk, line 56: redundant variable

src/formatDate.js
```diff
@@ -42,15 +42,50 @@
...
+    newKey = 'YYYY';
+
+    if (objectData['YY'] < 30) {
+      century = 20;
+    } else {
+      century = 19;
+    }
+
+    objectData.YY = century + objectData['YY'];
+    objectData[newKey] = objectData.YY;
```
TarasHoliuk, line 74:
```suggestion
    objectData['YYYY'] = century + objectData['YY'];
```
no need to rewrite `YY` value, also `newKey` redundant

src/formatDate.js
```diff
@@ -42,15 +42,50 @@
...
+  }
+
+  if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
+    newKey = 'YY';
+
+    objectData.YYYY = objectData['YYYY'].slice(-2);
+    objectData[newKey] = objectData.YYYY;
```
TarasHoliuk, line 81: similar comment as above

src/formatDate.js
```diff
@@ -42,15 +42,50 @@
  *   ['DD', 'MM', 'YYYY', '.'],
  * ) // '18.02.1997'
  *
- * @param {string} date
- * @param {string[]} fromFormat
- * @param {string[]} toFormat
- *
- * @returns {string}
- */
+//  * @param {string} date
```
TarasHoliuk, line 45: no need to add `//` - it's already a comments

TarasHoliuk (changes requested):

maxim2310 (approved):
Good work!!!

PR #1716, opened 2023-12-14
solution

vSainiuk:


src/formatDate.js
```diff
@@ -50,7 +50,27 @@
...
+  }
+
+  if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
+    keysDates['YY'] = keysDates['YYYY'].slice(-2);
+  }
+
+  if (fromFormat.includes('YY') && toFormat.includes('YYYY')) {
+    keysDates['YYYY'] = keysDates['YY'] < 30
+      ? '20' + keysDates['YY'] : '19' + keysDates['YY'];
+  }
```
dinaminahi, line 69: Let's create some const variables for 30, 20 ... to avoid magic values

dinaminahi (approved):
Good job 👍

PR #1720, opened 2023-12-17
Add solution

OleksandrRezanov:


src/formatDate.js
```diff
@@ -50,7 +50,67 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dayDescription = 'DD';
+  const monthDescription = 'MM';
+  const yearDescription = 'YY';
+  const fullYearDescription = 'YYYY';
+
+  const indexOfDevider = 3;
```
olya-shyian, line 58: such fixed values are better to write in uppercase format like 'DAY_DESCRIPTION', 'INDEX_OF_DEVIDER'

When you will be working on project such constants will be kept outside the function (for example in separate file)

olya-shyian (changes requested):
Nicely done!😊
Just a minor update

AntonDuchenko (approved):
Perfect job! :grin:

YegorKorenkov (approved):
Good job!

PR #1729, opened 2023-12-20
first

p1zzat1me:


src/formatDate.js
```diff
@@ -51,6 +51,32 @@
...
+        finalArrToStr.push(`19${fromFormatArr[j]}`);
+       }
+      }
+      if (toFormat[i] === 'YY' && finalArrToStr.length !== 3) {
+        if (fromFormat[j] === 'YYYY') {
+         finalArrToStr.push(`${fromFormatArr[j].slice(2)}`);
+        }
+       }
+    }
+  }
```
SanyaBratashchuk, line 77: try to simplify code avoiding nested loops

SanyaBratashchuk (changes requested):

src/formatDate.js
```diff
@@ -48,9 +48,28 @@
  *
  * @returns {string}
  */
-
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromFormatArr = date.split(`${fromFormat[3]}`);
```
l4st1m0za, line 52: You're using template strings unnecessarily. There's no need to use them if you're not including any variables or expressions. Instead of `${fromFormat[3]}` you can just use fromFormat[3].

src/formatDate.js
```diff
@@ -48,9 +48,28 @@
...
+      const yearIndex = fromFormat.indexOf('YY');
+      const yearValue = +fromFormatArr[yearIndex];
+      finalArrToStr.push(yearValue < 30 ? `20${yearValue}` : `19${yearValue}`);
+    } else if (toFormatElement === 'YY' && finalArrToStr.length !== 3) {
+      const yearIndex = fromFormat.indexOf('YYYY');
+      finalArrToStr.push(`${fromFormatArr[yearIndex].slice(2)}`);
+    }
+  }
+
+  return finalArrToStr.join(`${separator}`);
```
l4st1m0za, line 72: Same as line 7, you're unnecessarily using template strings. Instead of `${separator}` you can just use separator.

l4st1m0za (approved):
Great job! Consider comments for the future tasks!

PR #1735, opened 2023-12-26
create solution

AnastasiiaHombalevska:


anastasiiavorobiova (changes requested):
Great work! Consider using variables with meaningful names instead of magic strings and numbers

VitaliyBondarenko1982 (approved):
GJ!

PR #1739, opened 2023-12-29
Solution

OksanaZderko:


src/formatDate.js
```diff
@@ -50,7 +50,35 @@
...
+    dateObject[fromFormat[i]] = dateArray[i];
+  }
+
+  for (const element of toFormat) {
+    for (const key in dateObject) {
+      if (element === key) {
+        result.push(dateObject[key]);
+      }
+
+      if (element === 'YY' && key === 'YYYY') {
```
anastasiiavorobiova, line 67: Consider using variables with meaningful names instead of magic strings
```suggestion
      if (element === SHORT_YEAR && key === LONG_YEAR) {
```

src/formatDate.js
```diff
@@ -50,7 +50,35 @@
...
+  const dateArray = date.split(fromFormat[3]);
+  const dateObject = {};
+  const result = [];
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    dateObject[fromFormat[i]] = dateArray[i];
+  }
+
+  for (const element of toFormat) {
+    for (const key in dateObject) {
```
anastasiiavorobiova, line 62:
```suggestion

```
It's enough to have one cycle here

anastasiiavorobiova (changes requested):
Great work! Almost done!

TarasHoliuk (approved):

PR #1744, opened 2024-01-03
format date

marichkamt:


Andriy-Shymkiv (approved):
the code looks good, but you don't have error handling. what if `fromFormat` or `toFormat` or `date` is not valid? the function will work incorrectly, so you can add some error handling for basic cases

PR #1749, opened 2024-01-07
finished code

kapesha:
its looks not very good, but its works

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -50,7 +50,50 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const OldDate = date.split(fromFormat[3]);
+  const newDate = [];
+  const oldFormat = {};
+  let year;
+
+  for (let i = 0; i < 3; i++) {
+    oldFormat[fromFormat[i]] = OldDate[i];
+  }
+
+  if (oldFormat.YYYY) {
+    year = oldFormat.YYYY;
+  } else {
+    year = oldFormat.YY;
+  }
+
+  if (fromFormat.includes('YYYY') & toFormat.includes('YY')) {
+    year = oldFormat.YYYY.slice(2);
+  }
+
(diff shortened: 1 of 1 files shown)


src/formatDate.js
```diff
@@ -50,7 +50,50 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const OldDate = date.split(fromFormat[3]);
+  const newDate = [];
+  const oldFormat = {};
+  let year;
+
+  for (let i = 0; i < 3; i++) {
```
SanyaBratashchuk, line 58:
```suggestion
  for (let i = 0; i < 3; i++) {
```
move 3 to the constant because it looks like a magic number

src/formatDate.js
```diff
@@ -50,7 +50,50 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const OldDate = date.split(fromFormat[3]);
```
SanyaBratashchuk, line 53:
```suggestion
  const oldDate = date.split(fromFormat[3]);
```

src/formatDate.js
```diff
@@ -50,7 +50,50 @@
...
+    year = oldFormat.YYYY;
+  } else {
+    year = oldFormat.YY;
+  }
+
+  if (fromFormat.includes('YYYY') & toFormat.includes('YY')) {
+    year = oldFormat.YYYY.slice(2);
+  }
+
+  if (fromFormat.includes('YY') & toFormat.includes('YYYY')) {
```
SanyaBratashchuk, line 72:
```suggestion
  if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
    year = oldFormat.YYYY.slice(2);
  }

  if (fromFormat.includes('YY') && toFormat.includes('YYYY')) {
```

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #1765, opened 2024-01-17
formatDate

vitalii-labunskyi:


src/formatDate.js
```diff
@@ -50,7 +50,26 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateSplit = date.split(fromFormat[fromFormat.length - 1]);
+  const dateArr = [];
+
+  fromFormat.forEach((i, j) => {
+    dateArr[i] = dateSplit[j];
```
TarasHoliuk, line 57: `i` is some string in this case, for example `'DD'`. Do you use it as a key of ... array? You should have `dateObject` in this case

TarasHoliuk (changes requested):

PR #1766, opened 2024-01-17
Solution

TMukha:


src/formatDate.js
```diff
@@ -50,7 +50,32 @@
...
+      if (toFormat[i] === fromFormat[y]) {
+        newDate[i] = dateToArray[y];
+      }
+
+      if (toFormat[i] === 'YY' && fromFormat[y] === 'YYYY') {
+        newDate[i] = dateToArray[y].slice(2);
+      }
+
+      if (toFormat[i] === 'YYYY' && fromFormat[y] === 'YY'
+      && dateToArray[y] < 30) {
```
TarasHoliuk, line 67: No need to create 2 almost the same `if`. Check `dateToArray[y] < 30` in ternary inside `if` block

src/formatDate.js
```diff
@@ -50,7 +50,32 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateToArray = date.split(fromFormat[3]);
+  const newDate = [];
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
+    for (let y = 0; y < fromFormat.length - 1; y++) {
+      if (toFormat[i] === fromFormat[y]) {
```
TarasHoliuk, line 58: You write `toFormat[i]`, `fromFormat[y]` very often. Let's create variables with meaningful names for them

TarasHoliuk (changes requested):

SanyaBratashchuk (approved):

PR #1784, opened 2024-01-30
add task solution

Xsaites:


etojeDenys (changes requested):
I cannot find the logic in your code that handles case like this:
fromFormat has "YY" and toFormat has "YY"

anastasiiavorobiova (changes requested):
Great work! Almost done!

etojeDenys (approved):
good job

PR #1786, opened 2024-01-31
Solution

ApheliosGM:


src/formatDate.js
```diff
@@ -50,7 +50,39 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const separator = toFormat[3];
+  const arrayDate = date.split(fromFormat[3]);
```
vadiimvooo, line 54: If you created sepatate value for toFormat separator, you can do the same for fromFormat as well

vadiimvooo (approved):
Good job!)

PR #1809, opened 2024-02-08
Solution

kovaltar:


src/formatDate.js
```diff
@@ -50,7 +50,56 @@
...
+      case el.includes('DD'):
+        objDate.day = oldDateArr[ind];
+        break;
+      default:
+        objDate.delimiter = oldDateArr[ind];
+    }
+  });
+
+  let { year, delimiter } = objDate;
+  const { month, day } = objDate;
```
TarasHoliuk, line 74:
```suggestion
  let { year } = objDate;
  const { month, day, delimiter } = objDate;
```
you don't change its value

src/formatDate.js
```diff
@@ -50,7 +50,56 @@
...
+    switch (true) {
+      case el.includes('YY'):
+        objDate.year = oldDateArr[ind];
+        break;
+      case el.includes('MM'):
+        objDate.month = oldDateArr[ind];
+        break;
+      case el.includes('DD'):
+        objDate.day = oldDateArr[ind];
+        break;
```
TarasHoliuk, line 67:
```suggestion
    switch (el) {
      case 'YY':
      case 'YYYY':
        objDate.year = oldDateArr[ind];
        break;
      case 'MM':
        objDate.month = oldDateArr[ind];
        break;
      case 'DD':
        objDate.day = oldDateArr[ind];
        break;
```
let's use `switch` in a more common way, now it's more like the set  of `if` blocks
same below

TarasHoliuk (changes requested):
Good solution :+1: 

2 minor comments about syntax

src/formatDate.js
kovaltar: image
But this value change here

src/formatDate.js
kovaltar: Change switch to if

etojeDenys (approved):
good job

PR #1815, opened 2024-02-10
Solution

mfcastilho:


src/formatDate.js
```diff
@@ -65,9 +65,13 @@ function formatDate(date, fromFormat, toFormat) {
...
       year = dateParts[index];
-    } else if (part === 'MM') {
+    }
+
+    if (part === 'MM') {
       month = dateParts[index];
-    } else if (part === 'DD') {
+    }
+
+    if (part === 'DD') {
```
jv-aquino, line 74: I was talking abou switch structure here - generally you should use if's (without else if) only if there's return statements

jv-aquino (changes requested):
Mario, sorry, I think I led you in the wrong direction in my first review - I was talking about switch and other structures to keep your code tight, and about the variables try to look again at naming conventions and also review your logic with the first commentary that I made. If somethings is not clear please tell me, sorry anyways.

[Google js naming convention](https://google.github.io/styleguide/jsguide.html#naming)

src/formatDate.js
```diff
@@ -50,7 +51,66 @@
...
+    }
+  }
+
+  const newDateFormatArr = [];
+
+  for (let i = 0; i < toFormatParts.length; i++) {
+    switch (toFormatParts[i]) {
+      case 'YYYY':
+        if (year.length === 2) {
+          const FORMATED_YEAR = year < 30 ? `20${year}` : `19${year}`;
```
EdPirro, line 86: Implicit type conversion in `year < 30` prefer `Number(year) < 30` this may cause unexpected behaviors

src/formatDate.js
```diff
@@ -50,7 +51,66 @@
...
+        break;
+      case 'DD':
+        day = dateParts[i];
+        break;
+    }
+  }
+
+  const newDateFormatArr = [];
+
+  for (let i = 0; i < toFormatParts.length; i++) {
```
EdPirro, line 82: Could use a `.map` instead of a `for` that always `.push`es into an external array; also since `i` is only ever used to index `toFormatParts` this could be a `for...of`

EdPirro (changes requested):
Great job, @mfcastilho! Your solution looks great! 👏

The only problem is that your PR is outdated, please sync your fork and make sure the automated tests run and succeed! If you're having issues with it, please let me know!

Also, I pointed out a few improvements for you to keep in mind for the next tasks.

PR #1822, opened 2024-02-15
Solution

ivan-gritsenko:


Rusik-best86 (approved):
Good job!

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const arrDate = date.split(fromFormat[3].toString());
+  const resultArray = [];
```
BudnikOleksii, line 54: result/array is not a descriptive name for the variable

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+  let day = 0;
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    if (fromFormat[i].includes('Y')) {
+      year = arrDate[i];
+    } else if (fromFormat[i].includes('M')) {
+      month = arrDate[i];
+    } else if (fromFormat[i].includes('D')) {
+      day = arrDate[i];
+    }
```
BudnikOleksii, line 67: - It's better to use `switch` instead of checking just one letter.
- And Y/M/D, etc. looks like magic values

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+  }
+
+  for (let i = 0; i < toFormat.length - 1; i++) {
```
BudnikOleksii, line 70: Use `for of` for better readability

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+    if (toFormat[i].includes('Y')) {
+      if (toFormat[i] === 'YY') {
+        if (year.toString().length === 2) {
+          resultArray.push(year);
```
BudnikOleksii, line 74: The same, and could be simplified

BudnikOleksii (changes requested):
Good job 👍 
Let's improve your solution, check the comments

src/formatDate.js
```diff
@@ -48,9 +48,64 @@
...
+  }
+
+  for (const formatPart of toFormat) {
+    if (formatPart.includes(YEAR)) {
+      if (formatPart === SHORT_YEAR_FORMAT) {
+        if (year.toString().length === 2) {
+          newFormatDate.push(year);
+        } else {
+          newFormatDate.push(year.toString().slice(2, 4));
+        }
```
BudnikOleksii, line 85: It's still too complicated, so many nested `if` statements, and as I mentioned it's better to use `switch`

BudnikOleksii (approved):
👍

PR #1826, opened 2024-02-17
sloved the task

ogerenko:
I guess my iterations and infinity `if` it's not the right solution for the task)

But, I was trying to solve the task with a more complicated solution and more simple for reading. And I was failing))

I'll be waiting for your suggestions and remarks! 
Thank you!

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -50,7 +50,62 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateValue = date.split(fromFormat[3]);
+  const newDate = [];
+
+  let day = '';
+  let month = '';
+  let year = '';
+
+  let lengthYearFrom = '';
+  let lengthYearTo = '';
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    if (fromFormat[i] === 'DD') {
+      day += dateValue[i];
+    }
+
+    if (fromFormat[i] === 'MM') {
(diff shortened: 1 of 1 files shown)


src/formatDate.js
```diff
@@ -50,7 +50,62 @@
...
+    }
+
+    if (fromFormat[i] === 'MM') {
+      month += dateValue[i];
+    }
+
+    if (fromFormat[i] === 'YY' || fromFormat[i] === 'YYYY') {
+      year += dateValue[i];
+      lengthYearFrom += fromFormat[i];
+    }
```
etojeDenys, line 75: better to use switch/case here instead of multiple if/else

src/formatDate.js
```diff
@@ -50,7 +50,62 @@
...
+      if (lengthYearFrom.length < lengthYearTo.length) {
+        if (year > 25) {
+          year = '19' + year;
+        }
+
+        if (year <= 25) {
+          year = '20' + year;
+        }
+      }
+
```
etojeDenys, line 103: same here

etojeDenys (changes requested):

ogerenko:
thank you for your remarks, I was thinking about switch case but didn't understand how to implement it.

maybe now you display me another solution like on the site 'solution of mentor'
because I believe my solution is to smile and long.

Thank you!

anastasiiavorobiova (approved):
Great work

PR #1829, opened 2024-02-18
Date format Dev

DenNice-cloud:


ViktoriiaRepo (approved):
Good job, but  avoid magic numbers as much as possible and use named constants and values to avoid confusion.

BudnikOleksii:
> Good job, but avoid magic numbers as much as possible and use named constants and values to avoid confusion.

Good point

src/formatDate.js
```diff
@@ -50,7 +50,35 @@
...
+    ? toFormat.indexOf('YYYY')
+    : toFormat.indexOf('YY');
+
+  if (fromFormat.includes('YYYY') && toFormat.includes('YYYY')) {
+    datePresForm[presYearPos] = datePrevForm[prevYearPos];
+  } else if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
+    datePresForm[prevYearPos] = datePrevForm[presYearPos].slice(-2);
+  } else if (fromFormat.includes('YY') && toFormat.includes('YYYY')) {
+    if (datePresForm[prevYearPos] < 30) {
+      datePresForm[prevYearPos] = '20' + datePrevForm[presYearPos];
```
BudnikOleksii, line 75: As the previous reviewer mentioned let's get rid of all magic values such as 'YYYY'/'YY'/30/'20' etc

BudnikOleksii (changes requested):
Good work 👍 
Let's get rid of magic values

Moroz-Dmytro (approved):

PR #1833, opened 2024-02-19
add solution

AntonSakh:


Nirvacsh (approved):
Don't know why i got review for formatt date from style_to_object task =/

PR #1834, opened 2024-02-19
complete

TideTwo:


choeqq (changes requested):
since you're using a lot of comments I think the variable names might be little confusing (you have to explain what they mean), could you please try to use better names? also please create constant variables for hardcoded numbers

TideTwo:
I did what was asked, but I think It's less readable now. And I only added the comments to show that I understand the code, so It does not look like I copied it from somewhere.

asakevych (approved):
Looks good to me :rocket:

PR #1838, opened 2024-02-19
add task solution

vladarskyi:


kshepetska (approved):
For me looks good, don't see any issues :)

src/formatDate.js
```diff
@@ -50,7 +50,54 @@
...
+    let current = currentFormat[value];
+
+    if (!currentFormat[value]) {
+      current = (value === 'YY')
+        ? currentFormat['YYYY']
+        : currentFormat['YY'];
+    }
+
+    switch (value) {
+      case 'DD':
```
stassokolov1, line 73: will be better define these types as constants

stassokolov1 (approved):
Overall looks good to me, I'm approving it🔥

PR #1842, opened 2024-02-22
add solution

BilokonYaroslav:


src/formatDate.js
```diff
@@ -51,6 +51,86 @@
...
+  };
+
+  if (yearLengthTo > yearLengthFrom) {
+    if (year >= 30) {
+      year = 1900 + +year;
+    };
+
+    if (year < 30) {
+      year = 2000 + +year;
+    };
```
etojeDenys, line 91: better to use ternary operator here

etojeDenys (approved):
good job

PR #1852, opened 2024-02-27
task solution

konradwasilewski444:
I got a lot of problem with tests. I couldnt do anything with error: TypeError: Error while loading rule 'prettier/prettier': context.getPhysicalFilename is not a function. So Im posting photo with passed test below: 
formatDate wyniki

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -1,4 +1,4 @@
-'use strict';
+"use strict";
 
 /**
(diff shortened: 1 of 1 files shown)


konradwasilewski444:
Im sorry, i forgot to delete them

choeqq (approved):
ok @konradwasilewski444 but why they were here in the first place?

PR #1861, opened 2024-03-01
ready

AmateZz:


src/formatDate.js
```diff
@@ -50,7 +50,57 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromArr = date.split(fromFormat[3]);
+  let fromMonth = 0;
+  let fromDay = 0;
+  let fromYear = 0;
```
SanyaBratashchuk, line 56: you can create an object instead of 3 variables it will look more compact

SanyaBratashchuk (changes requested):

src/formatDate.js
```diff
@@ -50,7 +50,59 @@
  */
 
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromArr = date.split(fromFormat[3]);
+  const from = {
+    Month: 0,
+    Day: 0,
+    Year: 0,
+  };
```
SanyaBratashchuk, line 58:
```suggestion
  const from = {
    month: 0,
    day: 0,
    year: 0,
  };
```

SanyaBratashchuk (approved):

PR #1867, opened 2024-03-03
Solution

andreinik1:


src/formatDate.js
```diff
@@ -50,7 +50,48 @@
...
+
+          if (year.length === 2 && year < 30) {
+            year = `20${year}`;
+          } else if (year.length === 2) {
+            year = `19${year}`;
+          }
+
+          newData[yearIndexToAdding] = year;
+        }
+      }
```
vadiimvooo, line 90: It's better to avoid nested loops. Please try to make it without this loop.

vadiimvooo (changes requested):
Good job. Just on thing to fix.

src/formatDate.js
```diff
@@ -50,7 +50,50 @@
...
+          }
+        } else if (yearFormatElement === 'YYYY') {
+          yearIndexToAdding = j;
+
+          if (year.length === 2 && year < 30) {
+            year = `20${year}`;
+          } else if (year.length === 2) {
+            year = `19${year}`;
+          }
+        }
```
sTorba24, line 90: you still have loop)

sTorba24 (approved):

PR #1879, opened 2024-03-12
Solution

nataliiailchuk:


src/formatDate.js
```diff
@@ -50,7 +50,41 @@
...
+  const newDate = [];
+
+  let day = '';
+  let year = '';
+  let month = '';
+
+  for (let i = 0; i < fromFormat.length; i++) {
+    if (fromFormat[i] === 'DD') {
+      day = dateValue[i];
+    } else if (fromFormat[i] === 'YYYY') {
```
Andriy-Shymkiv, line 63: no need to use `else if` blocks as it could be just `if/else`

Andriy-Shymkiv (approved):

PR #1882, opened 2024-03-13
js format date solution

petrushinsky-kirill:


lerastarynets (approved):
You're doing amazing, keep it up!

PR #1888, opened 2024-03-14
Solution

Aleks-ars:


src/formatDate.js
```diff
@@ -8,7 +8,53 @@
...
+
+          return newFormatDate;
+        }
+      }
+    }
+  }
+
+  let formatYear = '';
+
+  if (toFormat.indexOf('YY') === 2) {
```
etojeDenys, line 35: why do you need to check that YY index equals 2?

src/formatDate.js
```diff
@@ -8,7 +8,53 @@
...
+    dateIndex = fromFormat.indexOf('YYYY');
+    formatYear = oldDate.slice(dateIndex, dateIndex + 1);
+    formatYear = formatYear.join().slice(2);
+    newFormatDate.push(formatYear);
+  } else if (toFormat.indexOf('YYYY') === 0) {
```
etojeDenys, line 40: same here, why 0

src/formatDate.js
```diff
@@ -8,7 +8,53 @@
...
+    formatYear = oldDate.slice(dateIndex, dateIndex + 1);
+
+    if (formatYear < 30) {
+      formatYear.unshift('20');
+      formatYear = formatYear.join('');
+      newFormatDate.unshift(formatYear);
+    } else if (formatYear >= 30) {
+      formatYear.unshift('19');
+      formatYear = formatYear.join('');
+      newFormatDate.unshift(formatYear);
```
etojeDenys, line 51: you can just delete 2 first letter `formatYear.slice(2)`

src/formatDate.js
etojeDenys: move these magic numbers at least to some vars

anastasiiavorobiova (approved):
Great work!

PR #1891, opened 2024-03-18
Solution

valost:


src/formatDate.js
```diff
@@ -8,7 +8,35 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const SEPARATOR_OLD = fromFormat.pop();
+  const SEPARATOR_NEW = toFormat.pop();
```
SanyaBratashchuk, line 12: it is better to not mutate incoming arrays

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #1904, opened 2024-03-28
feat: add solution

mfcastilho:


src/formatDate.js
```diff
@@ -8,7 +9,66 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const OLD_SEPARATOR = fromFormat[3];
+  const NEW_SEPARATOR = toFormat[3];
```
jv-aquino, line 13: nice use of naming conventions!

jv-aquino (approved):
nice Mario! perfect solution! you used switch and conditionals in a very nice and clever manner! good job and keep on with these improvements!

PR #1907, opened 2024-03-31
Solution

ArtemZakhar:
Hi)
And have a good day!)

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -8,7 +8,67 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const separator = date.replace(/\d/gi, '')[0];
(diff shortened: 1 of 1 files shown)


src/formatDate.js
```diff
@@ -8,7 +8,67 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const separator = date.replace(/\d/gi, '')[0];
+  const newSeparator = toFormat[toFormat.length - 1];
```
alexander-ignatow, line 12: Minor, but I don't quite get, why you don't use just `toFormat[3]`

alexander-ignatow (approved):
👍 
Good job

src/formatDate.js
ArtemZakhar: I don't know) 
Very strange behavior)

PR #1908, opened 2024-04-01
Solution

lmmaarriiaa:


src/formatDate.js
```diff
@@ -8,7 +8,59 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldDate = date.split(fromFormat[3]);
```
iAlfreX, line 11: It's better to do it through fromFormat.length - 1

iAlfreX (approved):
Good job!

src/formatDate.js
lmmaarriiaa: Okay!

alexander-ignatow: In this very case, this is not critical, as what is exactly the aim of using `fromFormat.length - 1`?
Personally, I would use access by explicit index

src/formatDate.js
```diff
@@ -8,7 +8,59 @@
...
+  const newDate = [];
+
+  let year = '';
+  let month = '';
+  let day = '';
+  let oldYearFormat = '';
+  let newYearFormat = '';
+
+  for (let i = 0; i < fromFormat.length; i++) {
+    if (fromFormat[i].includes('D')) {
```
alexander-ignatow, line 21: Again, in this case, you can simply use `==DD`, because you don't have other formats with `D`

alexander-ignatow (approved):
Good job!
Please, check out some minor comments

src/formatDate.js
lmmaarriiaa: Got it, thanks!

PR #1915, opened 2024-04-03
Develop

denyssheremeta:


src/formatDate.js
```diff
@@ -1,14 +1,50 @@
...
+
+  const currentYear = new Date().getFullYear() % 100;
+
+  if (fromFormat[yearIndex] === 'YY') {
+    if (parseInt(year) <= currentYear) {
+      year = '20' + year.padStart(2, '0');
+    } else {
+      year = '19' + year.padStart(2, '0');
+    }
+  }
```
DanilWeda, line 26: Move all formats (YYYY,YY) to constant. 
Also move all number values (20, 19) to constant

DanilWeda (changes requested):
Good job. 
Check small comments)
Thanks

src/formatDate.js
```diff
@@ -1,14 +1,60 @@
...
+};
+
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldSeparator = fromFormat[3];
+  const newSeparator = toFormat[3];
+  const dateArr = date.split(oldSeparator);
+
+  const dayIndex = fromFormat.indexOf('DD');
+  const monthIndex = fromFormat.indexOf('MM');
```
alexander-ignatow, line 19: `DD` and `MM` also should be constants

alexander-ignatow (approved):
Good for now 👍 

However, please, avoid using string literals and magic numbers directly as a part of the logic. Such values should be constants (pls see my comment)

PR #1928, opened 2024-04-12
Solution

odunchyk:


alexander-ignatow (approved):
Very good and elegant solution!

PR #1940, opened 2024-04-25
Solution

danil-maksimenko:


src/formatDate.js
```diff
@@ -8,7 +8,65 @@
...
+    const component = fromFormat[i];
+    const value = dataParts[i];
+
+    if (component === 'YYYY') {
+      year = value;
+    }
+
+    if (component === 'YY') {
+      year = value;
+    }
```
GUSILLUS, line 24: Consider combining this in one if using `||` operator

src/formatDate.js
```diff
@@ -8,7 +8,65 @@
...
+  for (let i = 0; i < toFormat.length; i++) {
+    const component = toFormat[i];
+
+    if (component === 'YYYY') {
+      newDate.push(year);
+    }
+
+    if (component === 'YY') {
+      newDate.push(year);
+    }
```
GUSILLUS, line 58: Same

GUSILLUS (changes requested):
GJ! But:

src/formatDate.js
```diff
@@ -8,7 +8,57 @@
...
+
+    if (component === 'MM') {
+      newDate.push(month);
+    }
+
+    if (component === 'DD') {
+      newDate.push(day);
+    }
```
etojeDenys, line 58: better to simplify it using the switch/case

etojeDenys (approved):
lgtm

PR #1946, opened 2024-04-29
Done

IvanVaniuk:


src/formatDate.js
```diff
@@ -9,6 +9,65 @@
...
+    }
+
+    if (meaning === 'MM') {
+      month += dateParts[i];
+    }
+
+    if (meaning === 'DD') {
+      day += dateParts[i];
+    }
+  }
```
etojeDenys, line 46:
```suggestion
  const newDate = [];
  const objectDate = {};
  const splitDate = date.split(oldSeparator);

  for (let i = 0; i < splitDate.length; i++) {
    objectDate[fromFormat[i]] = splitDate[i];
  }
  
  if (fromFormat.includes('YY') && toFormat.includes('YYYY')) {
    if (objectDate['YY'] < 30) {
      objectDate['YYYY'] = '20' + objectDate['YY'];
    } else {
      objectDate['YYYY'] = '19' + objectDate['YY'];
    }
  }
```
we can simplify this logic, then just slice the year field if needed, and fill the newDate array with objectDate values

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #1947, opened 2024-05-06
Solution / Format Date Advanced

polinabs:
The function `formatDate` was created.

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -8,7 +8,53 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newFormatArr = [];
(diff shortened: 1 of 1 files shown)


src/formatDate.js
```diff
@@ -8,7 +8,53 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const newFormatArr = [];
+  const [, , , separatorTo] = toFormat;
+  const [, , , separatorFrom] = fromFormat;
```
etojeDenys, line 13:
```suggestion
  const separatorTo = toFormat[3];
  const separatorFrom = fromFormat[3];
```

etojeDenys (approved):
good job

PR #1970, opened 2024-05-26
Solution

MariCodec:


RuslanHnatyshyn (approved):
Good job.

src/formatDate.js
```diff
@@ -9,6 +9,34 @@
  */
 function formatDate(date, fromFormat, toFormat) {
   // write code here
```
roman-mirzoian, line 11: Do not leave unnecessary comments.
```suggestion
```

src/formatDate.js
```diff
@@ -9,6 +9,34 @@
...
+  const separator = date.split(fromFormat[3]);
```
roman-mirzoian, line 12: Using ``fromFormat[3]`` to define the delimiter is not flexible because it assumes that the delimiter is always in the 4th position, which may not always be correct. The delimiter should be defined as the last element of the format array.

Example:

```suggestion

  // Change: Define delimiter as the last fromFormat element
  const fromSeparator = fromFormat[fromFormat.length - 1];
  const toSeparator = toFormat[toFormat.length - 1];
  
  // Change: split date by separator
  const dateParts = date.split(fromSeparator);
```

Edit the code for these changes.

roman-mirzoian (changes requested):
Good job, but please pay attention to the all comments (AI comments too).

dinaminahi (changes requested):
Good job 👍 
Let's improve it a bit

dinaminahi (approved):
Well done ⭐

PR #1972, opened 2024-05-26
add solution

RuslanHnatyshyn:


MariCodec (approved):
well done!

src/formatDate.js
```diff
@@ -8,7 +8,61 @@
...
+  let month = '';
+  let day = '';
+  const result = [];
+
+  for (let i = 0; i < dateParts.length; i++) {
+    if (fromFormat[i] === 'YYYY') {
+      year = dateParts[i];
+    }
+
+    if (fromFormat[i] === 'YY') {
```
roman-mirzoian, line 24: Instead of using several if-statements, it is better to use the switch construct, which will make the code more readable.

roman-mirzoian (changes requested):
Good job, but please pay attention to the all comments (AI comments too).

dinaminahi (approved):
Well done ⭐

RuslanHnatyshyn:
Thank you for your comments.

PR #1975, opened 2024-05-29
solutionV1

kgGugala:


src/formatDate.js
```diff
@@ -1,14 +1,55 @@
...
+  const dateObj = {
+    day: '',
+    month: '',
+    year: '',
+  };
+
+  for (let i = 0; i < fromFormat.length; i++) {
+    if (fromFormat[i] === 'DD') {
+      dateObj.day = oldDate[i];
+    } else if (fromFormat[i] === 'MM') {
```
BogdanMaliuta, line 17: are you need this else if? 
you can only use if

BogdanMaliuta (approved):
Nice solution but don`t use else if if you no need this

PR #1979, opened 2024-06-03
add solution

FaiHamid:


src/formatDate.js
```diff
@@ -8,7 +8,32 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateArr = date.split(fromFormat[fromFormat.length - 1]);
```
roman-mirzoian, line 11: It is good practice to use separate variables to store delimiters. This makes the code more understandable and easier to debug.
Example: 
```suggestion

  const oldSeparator = fromFormat.pop();
  const newSeparator = toFormat.pop(); 

  const dateArr = date.split(fromFormat[fromFormat.length - 1]);
```

src/formatDate.js
```diff
@@ -8,7 +8,32 @@
...
+  const result = [];
```
roman-mirzoian, line 12: Use a more descriptive name for the variable, such as `formattedDate`, so that it is immediately clear what the variable is.

src/formatDate.js
```diff
@@ -8,7 +8,32 @@
...
+
+  result[toFormat.indexOf('DD')] = dateArr[fromFormat.indexOf('DD')];
+  result[toFormat.indexOf('MM')] = dateArr[fromFormat.indexOf('MM')];
```
roman-mirzoian, line 15: Check that `fromFormat` and `toFormat` actually contain the appropriate date parts. 
This will ensure that errors are avoided if one of the formats is missing.
I used the renamed variable ``result``
 Example:
```suggestion

  if (toFormat.includes('DD')) {
    formattedDate[toFormat.indexOf('DD')] = dateArr[fromFormat.indexOf('DD')];
  }
  if (toFormat.includes('MM')) {
    formattedDate[toFormat.indexOf('MM')] = dateArr[fromFormat.indexOf('MM')];
  }
```

roman-mirzoian (changes requested):
Good job but please pay attention to the comments.

roman-mirzoian (approved):
Well done!

PR #1980, opened 2024-06-04
Solution to the task

Karanelus:
My solution to the task. It's take big amount of time, and solution can be mutch easier, but my result have right to live xD

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -109,9 +109,9 @@
       }
     },
     "@babel/eslint-parser": {
-      "version": "7.23.10",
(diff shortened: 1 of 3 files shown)


natalia-klonowska (changes requested):
during tests on GitHub an internal error was encountered so could you make new commit to trigger tests again?

BogdanMaliuta (approved):
It`s works good.
But code is very hard to understand, mb you need more practice to write more readable code. Try to do extra tasks on leetcode, or codewars in js, and after that look on another users solution

PR #1982, opened 2024-06-07
format_date_done

Anastasiia-Tatarchenko:


src/formatDate.js
```diff
@@ -7,8 +7,56 @@
...
+      if (newFromDate.year < 30) {
+        newFromDate.year += 2000;
+      } else {
+        newFromDate.year += 1900;
+      }
+    } else if (format === 'MM') {
+      newFromDate.month = dateElements[i];
+    } else if (format === 'DD') {
+      newFromDate.day = dateElements[i];
+    }
```
etojeDenys, line 32: better to use switch case here instead of multiple if/else

etojeDenys (approved):
good job

PR #1989, opened 2024-06-20
Solution v 1.0

StanislavKapytsia:
Create a formatDate function that accepts the date string, the old fromFormat array and the new toFormat array. Function returns given date in new format.

The function can change a separator, reorder the date parts of convert a year from 4 digits to 2 digits and back.

When converting from YYYY to YY just use 2 last digit (1997 -> 97).
When converting from YY to YYYY use 20YY if YY < 30 and 19YY otherwise.

### DIFF
--- a/src/formatDate.js
+++ b/src/formatDate.js
@@ -8,7 +8,60 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const oldSeparator = fromFormat[3]; // For example "/"
+  const newSeparator = toFormat[3]; // For example "."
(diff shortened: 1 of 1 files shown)


etojeDenys (changes requested):
it looks too complex, try to simplify this logic and feel free to ask for some help in the chat or tag me there, I will try to help you with that

StanislavKapytsia:
I wrote in the chat and they rewrote the code a little, but Vadym Voiedilo did not see the full code

GUSILLUS (approved):
GJ

PR #1995, opened 2024-07-03
add solution

RimmaBuzuluk:


src/formatDate.js
```diff
@@ -8,7 +8,71 @@
...
+      case 'YY':
+        year = dateDay[i];
+        break;
+      case 'MM':
+        month = dateDay[i];
+        break;
+      case 'DD':
+        day = dateDay[i];
+        break;
+    }
```
iyehorova, line 36: `switch/case` should always have `default` case for error handling

We don't have a checklist in this task, but I think we need to add default case to both switches

iyehorova (changes requested):
Hi, Rimma. Good job! 
Please fix your code.

src/formatDate.js
```diff
@@ -8,7 +8,73 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
-}
+  // create property for symbols from and to
```
BudnikOleksii, line 11: Remove all comments

src/formatDate.js
```diff
@@ -8,7 +8,73 @@
...
+  const symbolFormat = fromFormat[fromFormat.length - 1];
+  const symbolTo = toFormat[toFormat.length - 1];
+  // create split array from date
+  const dateParts = date.split(symbolFormat);
+  // create property that will save new date
+  let year = Number();
+  let day = Number();
+  let month = Number();
```
BudnikOleksii, line 19: You don't need to create numbers - `Number()`

src/formatDate.js
BudnikOleksii: > `switch/case` should always have `default` case for error handling
> 
> We don't have a checklist in this task, but I think we need to add default case to both switches

Good point, you can throw an error in this case

src/formatDate.js
```diff
@@ -8,7 +8,75 @@
...
+    switch (toFormat[i]) {
+      case 'YYYY':
+        // if we need to create YYYY from 2 number,
+        // add "20" if year <30 or 19 if year >= 30
+        if (year.length === 2) {
+          if (year < 30) {
+            newDate.push('20' + year);
+          } else {
+            newDate.push('19' + year);
+          }
```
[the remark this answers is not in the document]
BudnikOleksii: > The condition 'year < 30' is doing a string comparison rather than a numerical one because 'year' is a string. This could lead to unexpected results. You should convert 'year' to a number before the comparison.

Good point

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution in accordance with suggestions

DanilWeda (approved):
Good job. 
Approved. 
Thanks)

PR #2003, opened 2024-07-07
Solution

NataliiaBoiko22:


src/formatDate.js
```diff
@@ -9,6 +9,57 @@
...
+  toFormat.forEach((format) => {
+    switch (format) {
+      case 'YYYY':
+        newDate += yearPart;
+        newDate += dividerNew;
+
+        break;
+      case 'YY':
+        newDate += yearPart.slice(-2);
+        newDate += dividerNew;
```
etojeDenys, line 45: you can add it only one time after the switch case

etojeDenys (changes requested):

volodymyr-soltys97 (approved):
Good job 👍

PR #2005, opened 2024-07-08
Solution

Kozubowicz:


src/formatDate.js
```diff
@@ -9,6 +9,58 @@
  */
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const dateSeparated = date.split(fromFormat[3]);
```
Zibi95, line 12: Instead of assuming the delimiter is at index 3, dynamically extract it.

src/formatDate.js
```diff
@@ -9,6 +9,58 @@
...
+  let day = '';
+  let month = '';
+  let year = '';
+
+  for (let i = 0; i < 3; i++) {
```
Zibi95, line 17: What if fromFormat or toFormat has more than 3 elements? To ensure flexible code, it's good to handle scenarios where fromFormat or toFormat might have more than 3 elements. This improves the function's usability and reliability. What I would use is `forEach` loop here.

src/formatDate.js
```diff
@@ -9,6 +9,58 @@
...
+      case 'YY':
+        year = dateSeparated[i];
+        break;
+      case 'MM':
+        month = dateSeparated[i];
+        break;
+      case 'DD':
+        day = dateSeparated[i];
+        break;
+    }
```
Zibi95, line 29: There is missing default case where you could throw an error or smth that the format is unprocessable.

Zibi95 (approved):
Well done! Read the comments for future reference. 👍

PR #2026, opened 2024-07-23
Solution

RostyslavSharuiev:


src/formatDate.js
```diff
@@ -8,7 +8,28 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dataCollector = {};
```
VenomMB, line 11:
```suggestion
  const dataMap = {};
```

src/formatDate.js
```diff
@@ -8,7 +8,28 @@
...
+  const newDateFormat = [];
```
VenomMB, line 12:
```suggestion
  const newDateFormats = [];
```

VenomMB (approved):
Great job, leave small comments

PR #2052, opened 2024-08-19
task solved

Artem-Brui:


src/formatDate.js
```diff
@@ -9,6 +9,58 @@
...
+      case 'd':
+        result.push(outputNeeded.day);
+        break;
+      case 'm':
+        result.push(outputNeeded.month);
+        break;
+      case 'y':
+        if (item.length === 4) {
+          result.push(outputNeeded.year);
+        } else if (item.length === 2) {
```
TarasHoliuk, line 56:
```suggestion
        } else {
```
there are no other possible cases

TarasHoliuk (approved):

PR #2072, opened 2024-08-30
done

MaxF1996:


MaxF1996:
We didn't learn map method yet, so I was guessing that it will be too soon use it here.

SanyaBratashchuk (approved):

PR #2088, opened 2024-09-07
add solution

Tulenty:


src/formatDate.js
```diff
@@ -8,7 +8,39 @@
...
+            result[i] = '20' + splitDate[year];
+          } else {
+            result[i] = '19' + splitDate[year];
+          }
+        } else {
+          result[i] = splitDate[year];
+        }
+      }
+    }
+  }
```
etojeDenys: it looks too complex, you can try to simplify it

at least you can use the switch case here instead of if statements

etojeDenys (changes requested):

anastasiiavorobiova (approved):
Awesome!

PR #2089, opened 2024-09-07
Kriuk Vlad solution

kriukvlad:


src/formatDate.js
```diff
@@ -8,7 +8,61 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const FROM_SEPARATOR_FORMAT = fromFormat[3];
```
TarasHoliuk: Do not use UPPER_CASE for common (dynamic - you've got its value based on function arguments) variables. We can use UPPER_CASE for really constant variables, that might be declared outside of the function and are always the same

Example: const DAYS_IN_WEEK = 7; 

We saved it in the variable so it will be clear what exactly it means in place we use it (instead of some magic 7 number).

But in the current task you don't need to use such variables

src/formatDate.js
```diff
@@ -8,7 +8,61 @@
...
+  const TO_SEPARATOR_FORMAT = toFormat[3];
+  let year, month, day;
```
TarasHoliuk: Let's create an object with keys that match format ('DD', 'MM').

Then in the second loop you will be able just take the correct value like this: `dateObject[part]`

src/formatDate.js
```diff
@@ -8,7 +8,61 @@
...
+    }
+
+    if (toFormat[i] === 'DD') {
+      RESULT.push(day);
+    }
+  }
+
+  const FINAL_RESULT = RESULT.join(TO_SEPARATOR_FORMAT);
+
+  return FINAL_RESULT;
```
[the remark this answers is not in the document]
TarasHoliuk: Good comment

TarasHoliuk (changes requested):

vadiimvooo (approved):
Well done

PR #2091, opened 2024-09-09
Solution

KrivorotkoYurii:


src/formatDate.js
```diff
@@ -1,14 +1,62 @@
...
+        day = dateToArrow[i];
+        break;
+
+      case 'MM':
+        month = dateToArrow[i];
+        break;
+
+      default:
+        year = dateToArrow[i];
+        break;
```
Valeraom, line 22: Maybe it's better to use separate cases for 'YY' and 'YYYY' and add default case for other wrong values

Valeraom (approved):
Looks good!

src/formatDate.js
```diff
@@ -1,14 +1,66 @@
...
+        month = dateComponents[i];
+        break;
+
+      case 'YY':
+        year = dateComponents[i];
+        break;
+
+      case 'YYYY':
+        year = dateComponents[i];
+        break;
```
BudnikOleksii, line 26:
```suggestion
      case 'YY':
      case 'YYYY':
        year = dateComponents[i];
        break;
```
- You can combine cases if they do the same
- And you still must have default case to handle unexpected input

BudnikOleksii (changes requested):
Good work 👍 
Just one moment should be fixed, in switch case you always need the default case to handle unexpected input(you can throw an error for example)

KosolapovAnatolii (approved):
Nice job! 🆒

PR #2120, opened 2024-10-05
Solution

Krykunov:


minimal005 (approved):
good job)

src/formatDate.js
```diff
@@ -8,7 +8,69 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  // Set the new divider type
+  const newDivider = toFormat[toFormat.length - 1];
```
BudnikOleksii: Remove all comments, your variable describes what it does.

If you need to write comments to provide more information think about how to improve your solution so every developer can understand what your code does instead

src/formatDate.js
```diff
@@ -8,7 +8,69 @@
...
+      ? toFormat.indexOf('YYYY')
+      : toFormat.indexOf('YY');
+  const toFormatMonthIndex = toFormat.indexOf('MM');
+  const toFormatDayIndex = toFormat.indexOf('DD');
+
+  // Figure out the desired format year literal length
+  const toFormatYearLength = toFormat[toFormatYearIndex].length;
+
+  // Transform the year to the desired literal length
+  const transformYear = (year, length) => {
```
BudnikOleksii: Move this function outside the function

BudnikOleksii (changes requested):
Good job 👍 
Left a few suggestions, check them

BudnikOleksii (approved):
Good job 👍

PR #2129, opened 2024-10-15
add solution

clavigo:


src/formatDate.js
```diff
@@ -9,6 +9,47 @@
...
+    const datePart = previousDate[i];
+
+    if (format === 'DD') {
+      day = datePart;
+    } else if (format === 'MM') {
+      month = datePart;
+    } else if (format === 'YYYY') {
+      year = datePart;
+    } else if (format === 'YY') {
+      year = datePart < 30 ? '20' + datePart : '19' + datePart;
```
etojeDenys: better to rewrite it using the switch case

etojeDenys (approved):

PR #2142, opened 2024-10-29
Develop

iirk1:


src/formatDate.js
```diff
@@ -8,7 +8,61 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparator = fromFormat.find(
```
anastasiiavorobiova: The separator is the last array element

src/formatDate.js
```diff
@@ -8,7 +8,61 @@
...
+    (item) => item === '/' || item === '-' || item === '.',
+  );
+
+  const separator = toFormat.find(
```
anastasiiavorobiova: The separator is the last array element

src/formatDate.js
```diff
@@ -8,7 +8,61 @@
...
+
+  const newArray = [];
+
+  const dateObject = {};
+
+  for (let i = 0; i < fromFormat.length - 1; i++) {
+    dateObject[fromFormat[i]] = newDate[i];
+  }
+
+  const indexOfYear = fromFormat.includes('YYYY')
```
anastasiiavorobiova:
```suggestion
  const indexOfYear = fromFormat.includes(LONG_YEAR_FORMAT)
```
Consider using variables instead of magic strings

anastasiiavorobiova (changes requested):
Great work! Almost done!

SanyaBratashchuk (approved):

PR #2145, opened 2024-10-30
Solution

natalia646:


src/formatDate.js
```diff
@@ -8,7 +8,54 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const [oldData, idFromYear, oldYear, fromSumbol] = getFormat(fromFormat);
```
diabetishche: change sumbol on symbol (correct form of the word)

diabetishche (approved):
Great job, just check if u are writing a word u wish to use correctly (like symbol)

PR #2155, opened 2024-11-07
Solution

OkMoroz:


src/formatDate.js
```diff
@@ -8,7 +8,35 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparator = fromFormat.pop();
```
Krykunov: Pop() method modifies the original array, this can cause an unexpected behavior, I would replace it with access by last index

src/formatDate.js
```diff
@@ -8,7 +8,35 @@
...
+  const toSeparator = toFormat.pop();
```
Krykunov: The same as on the previous line

Krykunov (changes requested):

Krykunov (approved):
Great job! 🔥

PR #2159, opened 2024-11-10
Solution

dmdamyan:


src/formatDate.js
```diff
@@ -8,7 +8,69 @@
...
+  const separator = fromFormat[3];
+  const dateArray = date.split(separator);
+  let day = '';
+  let month = '';
+  let year = '';
+
+  if (fromFormat[0] === 'YYYY' || fromFormat[0] === 'YY') {
+    year = dateArray[0];
+    month = dateArray[1];
+    day = dateArray[2];
```
volodymyr-soltys97: Now these values ​​are duplicated in many places, store them in a variable and use the variable everywhere
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/formatDate.js
dmdamyan: ПОдскажи как это сделать?

volodymyr-soltys97 (approved):

PR #2247, opened 2025-02-07
Solution

ochepurna:


anastasiiavorobiova (approved):
Great work!
There are a lot of magic numbers and strings, so the main recommendation is to use variables here to make the code easily maintainable and readable

PR #2258, opened 2025-02-13
Solution

SofiiaPelenska:


src/formatDate.js
```diff
@@ -8,7 +8,32 @@
...
+  }
+
+  if (fromFormat.includes('YY') && toFormat.includes('YYYY')) {
+    DATE_OBJECT['YYYY'] =
+      DATE_OBJECT['YY'] < 30
+        ? '20' + DATE_OBJECT['YY']
+        : '19' + DATE_OBJECT['YY'];
+  }
+
+  if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
```
daniil-trusov, line 28: Two 'if' blocks with almost same check in a row is not very efficient way to write code. There's 2 ways to improve this:
1) second check starts with 'else', so if first check is true, second is skipped. It's still not very efficient, but you process year only one time at least;
2) fromFormat and toFormat can be 'YY' or 'YYYY', but the main difference there is length. You can make variable like lengthDiff = fromFormat.length - toFormat.length. It's more efficient way because you have in mind all 3 possible outcomes and solve every one there:
a) lengthDiff = 0 - just return year as is;
b) lengtDiff < 0 - fromFormat is YY and toFormat is YYYY, add '20' or '19' to the year;
c) lengthDiff > 0 - fromFormat is YYYY and toFormat is YY, return last 2 digits of year.

daniil-trusov (approved):
Good job, your solution is compact and easy to read, variable names give a complete understanding of their purpose, and there also a clever use of array deconstruction. However one small improvement can be made - please see the comment. It's not critical, but I hope you'll note it)

PR #2265, opened 2025-02-18
add solution

mmaryblackk:


src/formatDate.js
```diff
@@ -8,7 +8,31 @@
...
+  }
+
+  if (fromFormat.includes('YY') && toFormat.includes('YYYY')) {
+    DATE_OBJECT['YYYY'] =
+      DATE_OBJECT['YY'] < 30
+        ? '20' + DATE_OBJECT['YY']
+        : '19' + DATE_OBJECT['YY'];
+  }
 
+  if (fromFormat.includes('YYYY') && toFormat.includes('YY')) {
```
daniil-trusov, line 28: It's better to add 'else' before second if block there, so if first check is passed, second will be skipped without entering

daniil-trusov (approved):
Awesome job! Your solution is compact and easy to read, good variable names. One small improvement can be made there, but it doesn't affect the solution - please check comment for details.

PR #2304, opened 2025-03-17
Solution

mkovshun:


mkovshun:
I don't sure if it's an optimal way to solve the task, if it is, can you tell me what I can improve in this task?

src/formatDate.js
```diff
@@ -8,7 +8,67 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const dateMass = date.split(fromFormat[3]);
+
+  let day = 0;
+  let month = 0;
+  let year = 0;
```
etojeDenys, line 15: you can try to simplify it using the object 
{
MM: ...,
DD: ....,
YY: ....,
YYYY: ....
}

and then you can easily create your response

etojeDenys (approved):
but that's okay

PR #2351, opened 2025-04-24
add format date advanced solution

DenisPitsul:


src/formatDate.js
```diff
@@ -7,8 +7,56 @@
...
+        toYear = Number('20' + fromYear);
+      } else {
+        toYear = Number('19' + fromYear);
+      }
+      toDateArr.push(toYear);
+    } else if (toFormat[i] === 'YY') {
+      if (fromYearFormat === 'YY') {
+        toYear = fromYear;
+      } else {
+        toYear = fromYear % 100;
```
lidiia-tsymborovych: Good job!!!    if  fromYear ends with 2 zeros, the toFormat will not be correct ( for ex. fromYear = 2000;  toYear = 2000 % 100 = 0)

lidiia-tsymborovych (approved):

PR #2354, opened 2025-04-26
solution for js_format_date_advanced

Dmytro-Nyzhnyk:


src/formatDate.js
```diff
@@ -8,7 +8,82 @@
...
+
+    if (toFormat[j] === 'YY') {
+      reorderDateArr[j] = year.slice(2);
+    }
+
+    if (toFormat[j] === 'MM') {
+      reorderDateArr[j] = month;
+    }
+
+    if (toFormat[j] === 'DD') {
```
VitaliyBondarenko1982, line 64: for strings as 'DD', 'MM' and other better to make constants.

VitaliyBondarenko1982 (approved):
Looks good.

PR #2358, opened 2025-04-29
Solution for formatDate

r-marianna:
I'm pretty sure there's a better and faster solution for this, but right now I'm stuck on this approach


djtyrf312 (approved):
Approved

PR #2367, opened 2025-05-06
Solution added

MaxOnRem:
Created a function that returns given date in new format.


src/formatDate.js
```diff
@@ -8,7 +8,80 @@
  * @returns {string}
  */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const fromSeparator = fromFormat[3];
+
+  const toSeparator = toFormat[3];
+
+  // Розбиваємо вхідну дату на частини
```
volodymyr-soltys97, line 15: Remove comments

src/formatDate.js
```diff
@@ -8,7 +8,80 @@
...
+        parts.DD = value;
+        break;
+    }
+  }
+
+  // Формуємо масив частин нової дати
+  const newDateParts = [];
+
+  // Додаємо до масиву відповідні частини дати згідно з новим форматом
+  for (const target of toFormat) {
```
volodymyr-soltys97, line 63: Usee `.forEach` method here

src/formatDate.js
```diff
@@ -8,7 +8,80 @@
...
+      newDateParts.push(parts.YY);
+    }
+
+    if (target === 'MM') {
+      newDateParts.push(parts.MM);
+    }
+
+    if (target === 'DD') {
+      newDateParts.push(parts.DD);
+    }
```
volodymyr-soltys97, line 78: Use switch case here instead of if conditions, it will be better

src/formatDate.js
```diff
@@ -8,7 +8,80 @@
...
+        // Якщо формат — повний рік, зберігаємо і його, і скорочену версію
+        parts.YYYY = value;
+        parts.YY = value.slice(2);
+        break;
+
+      case 'YY':
+        // Якщо скорочений рік, визначаємо століття і формуємо повний рік
+        parts.YY = value;
+
+        // const century = +value < 30 ? '20' : '19';
```
volodymyr-soltys97, line 44: Remove all comments
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

PR #2430, opened 2025-07-14
Task solution

horidor:
Added task solution.


MykytaMusaiev (approved):
Good job mate! Keep it up!

PR #2436, opened 2025-07-16
solution

isobol-js:


vladsulyma (approved):
Great work! This function is well-structured, flexible, and handles different date formats (YY, YYYY, MM, DD) with custom separators. The logic is clear.

PR #2437, opened 2025-07-17
Solution

vladsulyma:


src/formatDate.js
```diff
@@ -8,7 +8,49 @@
...
+  const currentYear = separatedDate[YEAR_INDEX];
+  const fromYearFormat = fromFormat[YEAR_INDEX];
+  const toYearFormat = toFormat[NEW_YEAR_INDEX];
+
+  if (fromYearFormat === 'YYYY' && toYearFormat === 'YY') {
+    result[NEW_YEAR_INDEX] = currentYear.slice(-2);
+  } else if (fromYearFormat === 'YY' && toYearFormat === 'YYYY') {
+    const yearNumber = parseInt(currentYear, 10);
+
+    if (yearNumber <= 25) {
```
horidor, line 41: As per task it should be yearNumber <= 30, but i won't request changes for this.

horidor (approved):
Good job! There was a small mistype in the correct year after which you should start using 20th century, but everything else was done correctly and in accordance to general code guidelines and code style requirements.

PR #2479, opened 2025-09-07
_answer test about data format_

S-Khraban:


src/formatDate.js
```diff
@@ -1,14 +1,30 @@
...
-/**
- * @param {string} date
- * @param {string[]} fromFormat
- * @param {string[]} toFormat
- *
- * @returns {string}
- */
 function formatDate(date, fromFormat, toFormat) {
-  // write code here
+  const separatorFrom = fromFormat[3];
```
DmytroLymar, line 4: Naming Convention for const in JS is UPPER_SNAKE_CASE
For example separatorFrom > SEPARATOR_FROM or FROM_SEPARATOR

And for the next lines

DmytroLymar (approved):
Good job!
Code looks good, but check naming for const.

PR #2520, opened 2025-11-10
My format date

shymkivvasyl:


src/formatDate.js
```diff
@@ -8,7 +8,26 @@
...
+  const dateObj = {
+    [fromFormat[0]]: parts[0],
+    [fromFormat[1]]: parts[1],
+    [fromFormat[2]]: parts[2],
+  };
+
+  const result = toFormat.slice(0, 3).map((part) => {
+    if (dateObj[part]) {
+      return dateObj[part];
+    } else if (part === 'YY') {
```
brespect, line 21: avoid using else

brespect (approved):
Well done!

src/formatDate.js
shymkivvasyl: const result = toFormat.slice(0, 3).map((part) => {
    if (partsObj[part]) {
      return partsObj[part];
    }

    if (part === 'YYYY') {
      const year = partsObj['YY'];

      return parseInt(year) < 30 ? '20' + year : '19' + year;
    }

    if (part === 'YY') {
      return partsObj['YYYY'].slice(-2);
    }
  });      

 I remade it like this

PR #2521, opened 2025-11-10
Solution

V4NY0K:


src/formatDate.js
```diff
@@ -8,7 +8,62 @@
...
+
+  if (parts.length !== 3) {
+    return date;
+  }
+
+  function pad2(x) {
+    const s = String(x);
+
+    return s.length === 1 ? '0' + s : s;
+  }
```
Anton-Kuchmasov, line 25:
```suggestion
 String(x).padStart(2, '0');
```
don't reinvent a wheel :)

src/formatDate.js
```diff
@@ -8,7 +8,62 @@
...
+
+  function convertFromYY(yy) {
+    const yy2 = pad2(yy);
+    const n = Number(yy2);
+
+    return (n < 30 ? '20' : '19') + yy2;
+  }
```
Anton-Kuchmasov, line 32:
```suggestion
 function convertFromYY(yy) {
  const yy2 = String(yy).padStart(2, '0');
  const n = Number(yy2);
  return (n < 30 ? '20' : '19') + yy2;
}

```

Anton-Kuchmasov: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

Anton-Kuchmasov (approved):
Good job!
Check out comments below - they might be helpful

PR #2522, opened 2025-11-12
Solution

DemonWayne:


src/formatDate.js
de-vasta, line 1: Great job on that! 💪🏼 The solution is neat, concise, and it's easy to understand enough. Got nothing to add (for real) but maybe just some small suggestions to make your code even more flawless...

src/formatDate.js
```diff
@@ -8,7 +8,38 @@
...
-  // write code here
+  let year = '';
+  let month = '';
+  let day = '';
+
+  const fromSeparator = fromFormat[fromFormat.length - 1];
+  const dateParts = date.split(fromSeparator);
+
+  fromFormat.slice(0, 3).forEach((part, index) => {
+    if (part === 'YYYY' || part === 'YY') {
```
de-vasta: I'd extract some hardcoded pieces to **constant** variables especially when they're reused more than once (e.g. `'DD'`, `'MM'` etc.)

src/formatDate.js
```diff
@@ -8,7 +8,38 @@
...
+      month = dateParts[index];
+    } else if (part === 'DD') {
+      day = dateParts[index];
+    }
+  });
+
+  const toSeparator = toFormat[toFormat.length - 1];
+
+  const formattedDateParts = toFormat.slice(0, 3).map((part) => {
+    if (part === 'YYYY') {
```
de-vasta: I'd consider to use `switch` (with its tidy structure) when there are solely _strict equality_ here. Or alternatively go with like:
```js
if (<condition>) {return <value>;}
if (<condition>) {return <value>;}
....
return <value>;
```
`else` is a little redundant here anyway, imo

de-vasta (changes requested):

src/formatDate.js
DemonWayne: Thanks! I've made the requested changes there: https://github.com/mate-academy/js_format_date_advanced/pull/2522/commits/a16912af73b547adc2131059dd9d8c2b37872160

de-vasta (approved):
Ok. Fine by me 👍🏽

ArturErmolenko (approved):

PR #2550, opened 2025-12-10
added solution

TetianaSobolieva:


src/formatDate.js
```diff
@@ -9,6 +9,57 @@
...
+    if (fromFormat[i] === 'MM') {
+      month += newDate[i];
+    }
+
+    if (fromFormat[i] === 'DD') {
+      day += newDate[i];
+    }
+  }
+
+  for (let i = 0; i < toFormat.length; i++) {
```
dmytro-halieba: Since `toFormat` includes the separator as its last element, this loop will process one extra item. This might cause issues if we add more logic to a code in the future. We could change it to `toFormat.length - 1` so we only iterate through the actual format parts

src/formatDate.js
```diff
@@ -9,6 +9,57 @@
  */
 function formatDate(date, fromFormat, toFormat) {
   // write code here
+  const result = [];
```
dmytro-halieba, line 12: In lines 12-19, for better readability, consider grouping related variables together:

```
const fromSeparator = fromFormat[fromFormat.length - 1];
const toSeparator = toFormat[toFormat.length - 1];

const dateParts = date.split(fromSeparator);

let year = '';
let month = '';
let day = '';
```

dmytro-halieba (approved):
Nice work! The implementation is clean and functional - all the core requirements are met correctly. The logic flows well and produces the expected results. Great job on this!

PR #2552, opened 2025-12-13
True version

SlidingMoon:


src/formatDate.js
```diff
@@ -8,7 +8,96 @@
...
+    if (isSep) {
+      parts[partIndex] = current;
+      partIndex += 1;
+      current = '';
+      i += fromSep.length;
+    } else {
+      current += date[i];
+      i += 1;
+    }
+  }
```
Dimononon, line 38: You created you own split() method for some reason? All this lines can be replaced by `const parts = date.split(fromSep);`

src/formatDate.js
```diff
@@ -8,7 +8,96 @@
...
+
+    if (yy < 30) {
+      year4 = `20${year2}`;
+    } else {
+      year4 = `19${year2}`;
+    }
+  }
+
+  if (year2 === undefined && year4 !== undefined) {
+    year2 = `${year4[year4.length - 2]}${year4[year4.length - 1]}`;
```
Dimononon, line 77: can be replaced by `year2 = year4.slice(2);`

src/formatDate.js
```diff
@@ -8,7 +8,96 @@
...
+  }
+
+  let result = '';
+
+  for (let idx = 0; idx < 3; idx++) {
```
Dimononon, line 82: hard coded value 3. Maybe use `toFormat.length - 1` instead? Or introduce const with descriptive name.

src/formatDate.js
```diff
@@ -8,7 +8,96 @@
...
+
+    if (token === 'DD') {
+      result += day;
+    } else if (token === 'MM') {
+      result += month;
+    } else if (token === 'YYYY') {
+      result += year4;
+    } else if (token === 'YY') {
+      result += year2;
+    }
```
Dimononon, line 97: This is not a mistake but I'd rather declare result as array, then push all days, month, year into it and then just .join(toSep) to get final result.

Dimononon (approved):
Nice job. You implemented your own split() method although this was not required. I'll approve your PR, so you can receive AI mentor review faster, but I think he'll give u similar comments. GL)

PR #2553, opened 2025-12-13
add solution

Dimononon:


src/formatDate.js
```diff
@@ -8,7 +8,60 @@
...
+  const dateArr = date.split(fromSeparator);
+  const result = [];
+
+  let year = '';
+  let month = '';
+  let day = '';
+
+  for (let i = 0; i < dateArr.length; i++) {
+    switch (fromFormat[i]) {
+      case 'YYYY':
```
illia-001: `'YYYY'`
2. [Checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values

and for the next lines

illia-001 (changes requested):
GJ👍
Please fix your code👌

src/formatDate.js
Dimononon: We don't have checklist for this task, and I'm not sure this must be changed as this are standart date tokens used in programming. But ok, I moved them to const.

illia-001 (approved):
Excellent work👌

PR #2630, opened 2026-04-02
Solution

timurradkevic:


linska (approved):
Nice solution overall 🔥

The main thing I’d improve is readability: the logic inside map is a bit overloaded, so it takes some effort to follow. Extracting a few intermediate variables could make it clearer.

PR #2632, opened 2026-04-06
feat: implement formatDate function

linska:


timurradkevic (approved):
The task is solved correctly. Small suggestion: for converting YYYY to YY, value.slice(-2) is a bit more explicit than value.slice(2)

